PR #5, opened 2017-03-23
Basic CI setup

glibsm:
Basic stuff to get CI up and running

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: go
+sudo: false
+go:
+  - 1.7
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: go
+sudo: false
+go:
```
breerly: Might consider `tip` as well.

glibsm: Yeah, the only downside of tip is it takes a lot longer to fetch and can be flaky at times

Issue #8, opened 2017-03-23
Cut 0.1


breerly:
We need to cut an `0.1` so that projects using DIG have something to pin to.

For example, UberFX should pin to `~0.1` so that patches are let in, e.g. `0.1.1`, but not new minor releases which might contain breaking changes, e.g. `0.2`.

glibsm:
https://github.com/uber-go/dig/releases/tag/v0.1

PR #9, opened 2017-03-23
Rename dig.Graph to dig.Container

### LINKED ISSUE #3: Rename Graph to Container
from @breerly 

The term "graph" collides with "application graph" and "object graph", which is a term that relates to the *users* set of shapes that must come together. 

Using "graph" in DIG's terminology is problematic for the same reason that using the term "inject" was: these terms exist to describe user-land concepts. 

Overloading these terms creates confusion. Consider the following [phrase](https://github.com/uber-go/fx/pull/312#issuecomment-281897574):

> The proposal is that all deps in the application graph, including clients and handlers, get registered into dig's graph so that the full graph can be resolved.

Here I used the term "application graph" to mean "all the shapes in user-land that must be wired using DI". This is confusing because DIG has a term "graph", which **overloads the graph with the things which must make their way into the graph**.

My suggestion is to use the classical term for this: "Container".

> All deps in the application graph, including clients and handlers, get registered into a DIG container so that the graph can be resolved.

This will prevent confusion and make it easier for users to understand and utilize the concepts in DIG:

```golang
// create a new dependency injection container
container := dig.NewContainer()

// register all shapes into the container
container.Register(...)

// build the object graph
container.Resolve()
```

breerly:
This clears up a conflation of terms by renaming `dig.Graph` to the classical term` Container`

Before: Use a Graph to build the object graph.
After: Use a Container to build the object graph.

Fixes #3


breerly:
Let's get #8 solved before this lands.

README.md
```diff
@@ -1,4 +1,4 @@
-# Dependency Injection Graph
+# dig - A dependency injection framework for Go
```
glibsm: shouldn't it be caps here?

breerly: This change sort of makes it not an acronym anymore.

It's up to you though, I thought the lowercase looks best. Similar thing happens to alot of acronym based names, same as yarpc which now shows as: https://github.com/yarpc/yarpc/blob/master/assets/yarpc.png

glibsm:
What about the `g := dig.New()` readme examples? :)

README.md
glibsm: as long as people don't call it "dee ai gee" i don't care :)
It's always going to be `package dig`, but in official literature probably referred to as DIG.

breerly: Want to go caps? DIG? dig? Dig?

Up to you @glibsm

glibsm:
granted you merge access, feel free to merge

README.md
breerly: Might even makes sense to do:

# go.uber.org/dig - A dependency injection framework for Go

breerly:
@glibsm give me today to work through all the variables and README.md

breerly:
@glibsm let's merge #12 first

README.md
glibsm: Yeah, that works too. Eventually we can get @Lupie to help with this i'm sure she would immediately know what to do :)

PR #14, opened 2017-03-24
Rename .Register to .Provide

### LINKED ISSUE #4: rename Register to Provide

glibsm:
Fixes #4


glibsm:
Doesn't look like I've done this properly.

Container now has `Provide`, but `MustRegister` is still there. Looks like I have missed all the Must* function variations.

PR #19, opened 2017-04-20
Register constructors with multiple return pointers

### LINKED ISSUE #18: Support multiple return object registration
Currently dig MustRegister takes funcs that return single object or (object, error) for registration in [Provide](https://github.com/uber-go/dig/blob/master/container.go#L61-L89). We need to add support so that users can write functions that need multiple object registration. For example, [tracing initialization](https://github.com/uber-go/fx/blob/dev/tracing/tracer.go#L35)

anuptalwalkar:
Fixes #18 
Constructors with multiple return objects are now registered and cached in the graph. 
This PR does not resolve generic objects returned on constructor initialization such as `io.Closer`

### DIFF
--- a/container.go
+++ b/container.go
@@ -74,7 +74,7 @@ func (c *Container) Provide(t interface{}) error {
 				return errReturnKind
 			}
 		}
-		return c.registerConstructor(t)
+		return c.provideConstructor(t)
 	case reflect.Ptr:
 		return c.provideObject(t, ctype)
 	default:
@@ -209,9 +209,7 @@ func (c *Container) provideObject(o interface{}, otype reflect.Type) error {
 }
 
 // constr must be a function that returns the result type and an error
-// registerConstructor takes two parameters
-// constr - constructor registered for all the return objects
-func (c *Container) registerConstructor(constr interface{}) error {
+func (c *Container) provideConstructor(constr interface{}) error {
 	ctype := reflect.TypeOf(constr)
 	// count of number of objects to be registered from the list of return parameters
 	count := ctype.NumOut()
--- a/container.go
+++ b/container.go
@@ -74,9 +74,9 @@ func (c *Container) Provide(t interface{}) error {
 				return errReturnKind
 			}
 		} else {
-			// last variable must be error
-			if ctype.Out(count-1) != _typeOfError {
-				return errReturnErrKind
+			// if last variable is an error, we will not add it to the graph
+			if ctype.Out(count-1) == _typeOfError {
+				count--
 			}
 		}
 		return c.registerConstructor(t, count)
@@ -214,6 +214,10 @@ func (c *Container) provideObject(o interface{}, otype reflect.Type) error {
 }
 
 // constr must be a function that returns the result type and an error
+// registerConstructor takes two parameters
+// constr - constructor registered for all the return objects
+// count - count of number of objects to be registered from the list of return parameters
+// If last parameter is an error, the count is reduced to igonre the error a
 func (c *Container) registerConstructor(constr interface{}, count int) error {
 	ctype := reflect.TypeOf(constr)
 	objTypes := make([]reflect.Type, count, count)
@@ -248,8 +252,10 @@ func (c *Container) registerConstructor(constr interface{}, count int) error {
 	// object needs to be part of the container to properly detect cycles
 	if cycleErr := c.detectCycles(&n); cycleErr != nil {
 		// if the cycle was detected delete from the container
-		delete(c.nodes, objTypes[0])
-		return errors.Wrapf(cycleErr, "unable to Provide %v", objTypes[0])
+		for objType := range objTypes {
+			delete(c.nodes, objType)
+		}
+		return errors.Wrapf(cycleErr, "unable to Provide %v", objTypes)
 	}
 
 	return nil
--- a/container.go
+++ b/container.go
@@ -63,24 +63,23 @@ func (c *Container) Provide(t interface{}) error {
 	defer c.Unlock()
 
 	ctype := reflect.TypeOf(t)
-
 	switch ctype.Kind() {
 	case reflect.Func:
-		switch ctype.NumOut() {
-		case 2:
-			if ctype.Out(1) != _typeOfError {
-				return errReturnErrKind
-			}
-			fallthrough
-		case 1:
+		couts := ctype.NumOut()
+		if couts == 0 {
+			return errReturnCount
+		} else if couts == 1 {
 			objType := ctype.Out(0)
 			if objType.Kind() != reflect.Ptr && objType.Kind() != reflect.Interface {
 				return errReturnKind
 			}
-		default:
-			return errReturnCount
+		} else {
+			// last variable must be error
+			if ctype.Out(couts-1) != _typeOfError {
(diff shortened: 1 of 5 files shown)


container.go
```diff
@@ -63,24 +63,23 @@ func (c *Container) Provide(t interface{}) error {
...
+		} else if count == 1 {
 			objType := ctype.Out(0)
 			if objType.Kind() != reflect.Ptr && objType.Kind() != reflect.Interface {
 				return errReturnKind
 			}
-		default:
-			return errReturnCount
+		} else {
+			// last variable must be error
+			if ctype.Out(count-1) != _typeOfError {
```
glibsm: I don't think we should ever force the error, right? Some objects never fail to construct and don't return the errors. In those cases we don't want to force people to wrap existing constructors just to return a nil here for error. 

I'd say check if the last one is an error, and if not consider them all objects for the graph.

container.go
```diff
@@ -215,34 +214,42 @@ func (c *Container) registerObject(o interface{}, otype reflect.Type) error {
 }
 
 // constr must be a function that returns the result type and an error
-func (c *Container) registerConstructor(constr interface{}) error {
+func (c *Container) registerConstructor(constr interface{}, count int) error {
```
glibsm: count param needs a comment here somewhere to indicate what's going on

container.go
```diff
@@ -215,34 +214,42 @@ func (c *Container) registerObject(o interface{}, otype reflect.Type) error {
...
+		c.nodes[objTypes[i]] = &n
+	}
 
 	// object needs to be part of the container to properly detect cycles
 	if cycleErr := c.detectCycles(&n); cycleErr != nil {
 		// if the cycle was detected delete from the container
-		delete(c.nodes, objType)
-		return errors.Wrapf(cycleErr, "unable to Provide %v", objType)
+		delete(c.nodes, objTypes[0])
+		return errors.Wrapf(cycleErr, "unable to Provide %v", objTypes[0])
```
glibsm: the error should be better here. iterate over the nodes and construct one string that joins all the types failed to provide, i.e. `failed to provide types A, B, C`

container_setup_test.go
```diff
@@ -165,3 +165,7 @@ func NewFlakyChild() (*FlakyChild, error) {
 func NewFlakyChildFailure() (*FlakyChild, error) {
 	return nil, errors.New("great sadness")
 }
+
+func threeObjects() (*Child1, *Child2, *Child3, error) {
```
glibsm: awesome!

node_test.go
```diff
@@ -20,22 +20,16 @@
 
 package dig
 
-import (
-	"testing"
+// func TestNodeStrings(t *testing.T) {
```
glibsm: wazzup here?

container_test.go
```diff
@@ -294,9 +294,8 @@ func TestCycles(t *testing.T) {
 	require.NoError(t, c.Provide(c1))
 	require.NoError(t, c.Provide(c2))
 	err := c.Provide(c3)
-
 	require.Contains(t, err.Error(), "unable to Provide dig.Type3")
-	require.Contains(t, err.Error(), "dig.Type3 -> dig.Type1 -> dig.Type3")
+	require.Contains(t, err.Error(), "func(dig.Type1) dig.Type3 -> func(dig.Type2, dig.Type3) dig.Type1 -> func(dig.Type1) dig.Type3")
```
glibsm: Hm.

I'm not sure this is a desirable change. Was the old error messaging not too descriptive?

container.go
anuptalwalkar: Oh, I thought the point of multi args return is to make the last variable strictly an `error` type. (At least that was the behavior with 2 parameter constructor). Will update.

container_test.go
anuptalwalkar: The old message was only reading the graph with strictly single object node or constructor resolving in single node. but now we have graph with constructor that creates multiple return nodes. This message is more descriptive now since constructor can have multiple args and cycle can be formed on one of the return argument in the funcNode.

anuptalwalkar:
ping?

container.go
```diff
@@ -63,24 +63,23 @@ func (c *Container) Provide(t interface{}) error {
...
 			objType := ctype.Out(0)
 			if objType.Kind() != reflect.Ptr && objType.Kind() != reflect.Interface {
 				return errReturnKind
 			}
-		default:
-			return errReturnCount
+		} else {
+			// if last variable is an error, we will not add it to the graph
+			if ctype.Out(count-1) == _typeOfError {
+				count--
```
glibsm: i'm not sure why this count var is here, since we've already done the error checking. Can't this go inside the `registerConstructor` handling so we have one less variable to pass?

container_test.go
```diff
@@ -294,9 +293,8 @@ func TestCycles(t *testing.T) {
 	require.NoError(t, c.Provide(c1))
 	require.NoError(t, c.Provide(c2))
 	err := c.Provide(c3)
-
-	require.Contains(t, err.Error(), "unable to Provide dig.Type3")
-	require.Contains(t, err.Error(), "dig.Type3 -> dig.Type1 -> dig.Type3")
+	require.Contains(t, err.Error(), "unable to Provide [dig.Type3]")
+	require.Contains(t, err.Error(), "func(dig.Type1) dig.Type3 -> func(dig.Type2, dig.Type3) dig.Type1 -> func(dig.Type1) dig.Type3")
```
glibsm: can we add a test somewhere that shows `func(type1, type2) type3, type4 -> func(type4) type1` or something like that?

container.go
anuptalwalkar: makes sense. +1

container_test.go
anuptalwalkar: Added test cycles withConstructor

node.go
```diff
@@ -125,26 +128,39 @@ func (n *funcNode) value(g *Container) (reflect.Value, error) {
...
+				n.cachedValue = append(n.cachedValue, v)
+			}
+		}
 	}
 
-	n.cached = true
-	n.cachedValue = v
-	return v, err
+	// last value must be an error
+	err, _ := values[len(values)-1].Interface().(error)
```
glibsm: either the comment here is not true, or it's not doing what we expect it to do.

`func(type1, type2) (type3, type4)` should be legit without errors

node.go
```diff
@@ -125,26 +128,39 @@ func (n *funcNode) value(g *Container) (reflect.Value, error) {
...
-	v := values[0]
 
-	// If the function has two return values, the second one is an error.
-	var err error
-	if len(values) > 1 {
-		err, _ = values[1].Interface().(error)
+	// cache constructed values in the node
+	for _, node := range n.nodes {
+		for _, v := range values {
+			if node.objType == v.Type() {
```
glibsm: might be worthwhile to consider nodes to be `map[reflect.Type]node`

node.go
anuptalwalkar: Fixing the comment. Its not "must be an error". Its only checking if last value is an error and returning it.

node.go
anuptalwalkar: I'll follow up PR on this since its not impacting performance too much.

Issue #21, opened 2017-04-21
Add ability to directly call a constructor through dig


glibsm:
Let's say that I have a constructor `func (A, B) (C, error) { // codez }`. 

In order to partake in dig, I'd have to first Provide said constructor it to the graph, then declare a new variable of type C, and ask the graph to resolve:

```go
container := dig.New()
container.Provide(func (A, B) (C, error) { // codez })
var c C
conatiner.Resolve(&c)
// c is now usable
```

I propose to add the `Call` function to `Container` interface, to make this process a lot smoother. `Call` would execute the function, inject the dependencies (without permanently adding it to the graph) and return the interface{} result (ready for casting).

The new experience would be:

```go
container := dig.New()
c := container.Call(func (A, B) (C, error) { // codez }).(C)
// c is bueno to use
```

Not only this improves the API, but also allows users to create multiple objects of the same type that should they choose to.

anuptalwalkar:
Interesting. A few questions - 
1. How will this work when A and B are not provided? In order to `Call` the func, you need to make sure that A and B are provided and not creating a cycle. 
2. What happens for multiple return args? The way dig is structured, resolve `c` will get you the constructor execution without needing to initialize other variables, but then what if single return object is creating dependency cycle? 
e.g.  `c, _ := container.Call(func (A, B) (C, B, error) { // codez })`

glibsm:
1. This case is easy - error return or a panic, since dig will fail to resolve A and B obviously.
2. This is slightly more sophisticated, perhaps we only consider `type, error` returns for Call, or the api needs to be re-thunk. I.e. `Call(func(), ...interface{})` the vararg of interfaces will be populated in the right order, or an error returned.

anuptalwalkar:
Ok, We can do this in 3 ways -

Option 1: return `[]interface{}` to users which is equivalent to the return args in the function provided. Users can then go on and convert the objects to their given types. 

`	result := c.Call(func(o1 Object1, o2 Object2) (Ret1, Ret2, error) {
		return Ret1{}, Ret2{}, nil
	})
	c1 := result[0].(*Ret1)
	c2 := result[1].(*Ret2)
`
Option 2: variadic arguments for Call function allows objects to be resolved in the call itself.
`	var c1 Ret1
	var c2 Ret2
	result := c.Call(func(o1 Object1, o2 Object2) (Ret1, Ret2, error) {
		return Ret1{}, Ret2{}, nil
	}, c1, c2)
`
[...]

PR #23, opened 2017-04-26
Add Invoke function to resolve dependency

anuptalwalkar:
#21 
Invoke function resolves the dependencies provided as arguments to the constructor, and invokes the function. The Invoke(func) can be used as a callback function with resolved dependencies from dig.

### DIFF
--- a/container.go
+++ b/container.go
@@ -56,6 +56,8 @@ type Container struct {
 // Invoke the function and resolve the dependencies immidiately without providing the
 // constructor to the graph. The Invoke function returns error object which can be
 // occurred during the execution
+// The return arguments from Invoked function are registered in the graph for later use
+// The last parameter, if it is an error, is returned to the Invoke caller
 func (c *Container) Invoke(t interface{}) error {
 	ctype := reflect.TypeOf(t)
 	switch ctype.Kind() {
@@ -82,7 +84,23 @@ func (c *Container) Invoke(t interface{}) error {
 		cv := reflect.ValueOf(t)
 
 		// execute the provided func
-		cv.Call(args)
+		values := cv.Call(args)
+		if len(values) > 0 {
+			err, _ := values[len(values)-1].Interface().(error)
+			c.Lock()
+			for _, v := range values {
+				n := objNode{
+					node: node{
+						objType:     v.Type(),
+						cached:      true,
+						cachedValue: v,
+					},
+				}
+				c.nodes[v.Type()] = &n
+			}
+			c.Unlock()
+			return err
+		}
 	default:
 		return errParamType
 	}
--- a/container.go
+++ b/container.go
@@ -68,7 +68,7 @@ func (c *Container) Invoke(t interface{}) error {
 		for idx := range args {
 			arg := ctype.In(idx)
 			if node, ok := c.nodes[arg]; ok {
-				v, err := node.value(c)
+				v, err := node.value(c, arg)
 				if err != nil {
 					return errors.Wrapf(err, "unable to resolve %v", arg)
 				}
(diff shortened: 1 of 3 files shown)


container.go
```diff
@@ -53,6 +53,40 @@ type Container struct {
 	nodes map[interface{}]graphNode
 }
 
+// Invoke the function and resolve the dependencies immidiately without providing the
+// constructor to the graph. The Invoke function returns error object which can be
+// occurred during the execution
+func (c *Container) Invoke(t interface{}) error {
+	c.Lock()
+	defer c.Unlock()
```
anuptalwalkar: not sure if locking is needed here since we are not updating the container.

container_test.go
```diff
@@ -241,6 +242,43 @@ func TestConstructorErrors(t *testing.T) {
...
+
+	err := c.ProvideAll(
+		&Parent1{},
+		&Parent12{},
+	)
+	assert.NoError(t, err)
+	var c1 Child1
+	var c2 Child2
+
+	err = c.Invoke(func(p1 *Parent1, p2 *Parent12) {
```
glibsm: validate p1 and p2

container.go
glibsm: agreed. lets move to right before Call

anuptalwalkar: updated to RLock()

container.go
```diff
@@ -112,6 +145,8 @@ func (c *Container) Resolve(obj interface{}) (err error) {
 	objElemType := reflect.TypeOf(obj).Elem()
 	objVal := reflect.ValueOf(obj)
 
+	c.Lock()
+	defer c.Unlock()
```
glibsm: might also be good to just put the RLock around the map access, but not a big deal.

glibsm:
I'm not sure that this fully satisfies the requirement we have for `Invoke`. Not only we want to be able to call, but also sometimes store the returned values into the graph (if we wanted to directly call a constructor).

That can probably be an option passed into `Invoke`, so this diff can be merged anyway, but lets talk through the scenario that I have in mind to make sure this is setting us on the correct path.

anuptalwalkar:
I added the functionality to register returned objects back to the graph. This satisfies the requirement of executing Invoke function and having the returned objects available for later Resolve. The Invoke will also support anonymous functions, if users don't wish to register any objects back to dig.

container.go
```diff
@@ -48,11 +48,63 @@ func New() *Container {
...
+			if ok {
+				c.RLock()
+				v, err := node.value(c, arg)
+				c.RUnlock()
+				if err != nil {
+					return errors.Wrapf(err, "unable to resolve %v", arg)
+				}
+				args[idx] = v
+			} else {
+				return fmt.Errorf("%v dependency of type %v is not registered", ctype, arg)
```
glibsm: i think we should move this to `c.get(type)` so this is not duplicated

glibsm: this should also do errors.Wrapf like the one above

container.go
```diff
@@ -196,16 +247,19 @@ func (c *Container) provideObject(o interface{}, otype reflect.Type) error {
 		otype = otype.Elem()
 		v = v.Elem()
 	}
+	c.insertObjectToGraph(v, otype)
+	return nil
+}
+
+func (c *Container) insertObjectToGraph(v reflect.Value, vtype reflect.Type) {
```
glibsm: no lock in here?

anuptalwalkar: The lock is applied in `Provide` and `Invoke` around insertObject func. I am refactoring this to move locking out and away from the container to simplify the API.

glibsm:
forgot to mention, could you update changelog?

PR #29, opened 2017-04-28
Refactor dig to separate graph handling and calling API

anuptalwalkar:
access to the nodes map, duplicate code, and all the locking logic is getting difficult to work around. This refactor should help.
 
1. container.go -> user APIs to deal with `package dig` (`Provide`, `Resolve`, `Invoke`, etc)

pros - container doesn't know about nodes map. 

2. `package graph` contains all the logic to insert object/constructors to the graph, read values from the graph and maintain locking throughout the usage.

pros- graph can manage its own lock


graph/graph.go
```diff
@@ -0,0 +1,200 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package graph
```
breerly: Let's move this package `internal/` - I'd like to keep the API surface as small as possible.

graph/graph.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+var (
+	errArgKind = errors.New("constructor arguments must be pointers")
+
+	_typeOfError = reflect.TypeOf((*error)(nil)).Elem()
+)
+
+// Graph contains all Graph for current graph
+type Graph struct {
```
glibsm: Im still thinking through this PR, but one thing is for certain: we don't want this to be a public type that users can interact with. It should either be private in the same package, or at the very least an internal package.

We have known for a while that options are coming: things like named instances, or private non-shared instances. I'd like for us to start thinking and working through those APIs, because only then we will be able to tell if this refactor is a good or bad thing.

anuptalwalkar:
@breerly Locking is needed right now because we have APIs reading and iterating over the `nodes` map, Resolving of `ProvideConstructor`, and `Invoke` are iterating over the map and updating the map in the same call. `nodes.value()` call does the same recursively.

graph/graph.go
anuptalwalkar: +1 agreed. moving to internal.

graph/graph.go
anuptalwalkar: Moving to internal package to avoid exposing graph to public. 
With options coming, this will save us lot of headache of dealing with locks, maps, and reflect :)

container.go
```diff
@@ -62,24 +56,21 @@ func (c *Container) Invoke(t interface{}) error {
...
+
 		if len(values) > 0 {
-			c.Lock()
-			defer c.Unlock()
 			err, _ := values[len(values)-1].Interface().(error)
 			for _, v := range values {
 				switch v.Type().Kind() {
 				case reflect.Ptr:
-					c.insertObjectToGraph(v, v.Type())
+					c.Graph.InsertObject(v, v.Type())
```
alsamylkin: looks like we should strip type from the InsertObject, container should get it form the object itself.

internal/graph/graph.go
```diff
@@ -0,0 +1,215 @@
...
+	errArgKind = errors.New("constructor arguments must be pointers")
+
+	_typeOfError = reflect.TypeOf((*error)(nil)).Elem()
+)
+
+// Graph contains all Graph for current graph
+type Graph struct {
+	sync.RWMutex
+
+	nodes map[interface{}]graphNode
```
alsamylkin: why it is a `map[interface{}]graphNode` not a `map[reflect.Type]graphNode`?

container.go
anuptalwalkar: good point. I'll update.

internal/graph/graph.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	argc := ctype.NumIn()
+	n := funcNode{
+		deps:        make([]interface{}, argc),
+		constructor: constr,
+		nodes:       nodes,
+	}
+	for i := 0; i < argc; i++ {
+		arg := ctype.In(i)
+		if arg.Kind() != reflect.Ptr && arg.Kind() != reflect.Interface {
```
alsamylkin: hmm, should the container care about it? Extending for slices and map will be hard then.

internal/graph/graph.go
anuptalwalkar: I don't want to change the map type in the same PR as it will get hard to read and change. Will follow up.

internal/graph/graph.go
anuptalwalkar: We are constructing values from the node, which container isn't aware of, so as of this PR graph is ideal place for building constructor arguments. I'll add/change it once slices and maps are introduced.

internal/graph/graph.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// Read reads value from the Graph
+func (g *Graph) Read(objType reflect.Type) (reflect.Value, error) {
+	// check if the type is a registered objNode
+	g.RLock()
+	n, ok := g.nodes[objType]
+	g.RUnlock()
+	if !ok {
+		return reflect.Zero(objType), fmt.Errorf("type %v is not registered", objType)
```
alsamylkin: nit: just reflect.Value{}?

alsamylkin: On the other hand, why return values, not pointers?

internal/graph/graph.go
```diff
@@ -0,0 +1,215 @@
...
+			cached:      true,
+			cachedValue: v,
+		},
+	}
+	g.nodes[v.Type()] = &onode
+	return nil
+}
+
+// InsertConstructor adds the constructor to the Graph
+func (g *Graph) InsertConstructor(constr interface{}) error {
```
alsamylkin: nit: usually constructor is abbreviated with ctor, constr is to close to const :)

internal/graph/graph.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+		return errors.Wrapf(cycleErr, "unable to Provide %v", objTypes)
+	}
+
+	return nil
+}
+
+// When a new constructor is being inserted, detect any present cycles
+func (g *Graph) detectCycles(n *funcNode) error {
+	l := []string{}
```
alsamylkin: that's not needed, recursiveDetectCycles will work with nil too.

anuptalwalkar: interesting..

internal/graph/graph.go
anuptalwalkar: We set the pointer value of the provided object to the instance pointer in container. It makes sense since the function is defined to read value from the graph.

anuptalwalkar: Also,"reflect.Zero returns a Value representing the zero value for the specified type." seems more appropriate to return as a default than reflect.Value{}?

alsamylkin: There is no point of having a `reflect.Value` of specific type if there is an error :)
My comment was about unnecessary copying of reflect.Value that is going all around the graph. After digging through the code it looks like you'll need to change the graphNode interface to avoid it. May be another time then.

PR #31, opened 2017-04-29
Add readme badges

glibsm:
Add godoc, covrage, travis and report card badges to the readme. I think the coveralls one might take a build or two to fully kick in (I just enabled in)


glibsm:
image

breerly:
(suggest codecov.io over coveralls)

glibsm:
i was just being consistent with the rest of the org. if codecov is better i can investigate

Issue #34, opened 2017-05-03
 Combine Resolve, ResolveAll, and Invoke into a Locate func


breerly:
When you ask the container to resolve dependencies, that is called the "service locator" pattern.

Combining `Resolve`, `ResolveAll`, and `Invoke` into a `Locate` func will streamline the experience:

```go
// locate 1 or more types
var logger *zap.Logger
var dispatcher *yarpc.Dispatcher
container.Locate(&logger, &dispatcher)

// same as above, using a func
container.Locate(func(logger *zap.Logger, dispatcher *yarpc.Dispatcher) {
 
})
```

This has the benefit of:

* Being more specific, giving a nod to the pattern actually being used here
* Reduces API surface and simplifies using DIG

breerly:
(p.s. this issue is to *discuss* - had the idea this A.M. and wanted to jam on it)

breerly:
In fact, I'm even in favor of *only* having this form:

```go
container.Locate(func(logger *zap.Logger, dispatcher *yarpc.Dispatcher) {
 
})
```
I don't really see the benefit to the original `Resolve` form with the above.

glibsm:
This is definitely something I've thought about once we solidified the api for `Invoke`. @anuptalwalkar and I chatted offline and realized we can completely remove all forms of `Provide` and replace with this one thing.

I will let him chime in, since he had some thoughts and why we shouldn't do it.

anuptalwalkar:
I like having three APIs - `Provide`, `Resolve` and `Invoke` in the form of variadic arguments since each have its own use. We should remove all Must functions, and All functions.

I like the naming of `Invoke` rather than locate, simply because `Invoke` comes with a sense of providing a function (be it anonymous func), `Locate` sounds more like an API on an object. 

My main issue with only relying on Invoke was all the locking around the functionality, which is not the case anymore with the refactor. That being said, we should keep both `Resolve` and `Invoke`. The difference between `Invoke` and `Resolve` is the need to have everything wrapped in a function. What if I just need logger from the graph? Following is not very intuitive- 
```go
var l *zap.logger 
container.Invoke(func(logger *zap.Logger) {
   l = logger
})
```

`Provide` is allowing us to add multiple objects, constructors to the graph. Again, wrapping into anonymous function or invoking a function is not very intuitive.

breerly:
@anuptalwalkar, ok I buy that the first form is also useful, this is pretty ugly:

```go
var l *zap.logger 
[...]

PR #36, opened 2017-05-03
Remove Must* funcs to greatly reduce API surface area.

### LINKED ISSUE #35: Remove Must* methods
The Must* methods provide a nice convenience for end users, but aren't necessary because:

DIG is not for end users. DIG should be used to build init/application frameworks.

Forcing init/app frameworks to explicitly manage errors means a reduced API surface, and overall simplified experience.

breerly:
Resolves #35

### DIFF
--- a/container.go
+++ b/container.go
@@ -114,13 +114,6 @@ func (c *Container) Provide(t interface{}) error {
 	}
 }
 
-// MustRegister will attempt to Provide the object and panic if error is encountered
-func (c *Container) MustRegister(t interface{}) {
-	if err := c.Provide(t); err != nil {
-		panic(err)
-	}
-}
-
 // Resolve all of the dependencies of the provided class
 //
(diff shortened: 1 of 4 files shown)


breerly:
Note that from here, I'd like to look at #34, which would leave us with this public API:

* `Container.Provide`
* `Container.Locate`

I think from there, we'd be in a good place to discuss a 1.0.

container.go
```diff
@@ -114,13 +114,6 @@ func (c *Container) Provide(t interface{}) error {
 	}
 }
 
-// MustRegister will attempt to Provide the object and panic if error is encountered
-func (c *Container) MustRegister(t interface{}) {
```
breerly: Looks like we didn't rename this to `MustProvide` even.

akshayjshah:
Works for me.

Issue #26, opened 2017-04-27
Slices and maps semantics in dependencies and return types


glibsm:
Currently, dig only works with either a pointer to an object, or an interface. Slices and maps can sort be used by using a pointer to slice/map, or wrapping them in another struct.

Ideally, dig would support slices and maps as first class citizens. 

NOTE: there may be api implications for constructor parameters that are slices/maps. For example, if dig was handling the following constructor

```
func ([]typeA, typeB) typeC { }
```

What does it mean?
1. Does it mean that there is exactly one slice of type `[]typeA` registered in the graph and we should return that?
2. Does it mean there may be multiple constructors that return `func() typeA` or `func() []typeA` and we should fetch them all and put them into an slice?

Not handling case2 should probably be a starting point: there is only one slice of a given type that can live in the graph, we can deal with merging/extending existing slices through dig options in the future.

alsamylkin:
Well, A and []A are different types, so we shouldn't do anything special to them.

As for option 1, we do have a different behavior in `fx.config`: for slices we replace and for maps we merge contents. I think it was inherited to support yaml merging though.

akshayjshah:
A portion of this is addressed in #72.

glibsm:
Value types are now supported, which makes all sorts of slices and maps work nicely with dig.

There are no special semantics for when two arrays are provided into the container - that would result in an error. Named instances, however, allow for a third constructor to come in and merge them together.

PR #38, opened 2017-05-03
Introduce maps, slice and arrays to dig

anuptalwalkar:
#26 
Dig to support non named typed maps slices and arrays. 

We have constructors and providers to put `n` objects of type `x` into dig. This PR will allow users to add slices and maps of special types into dig and make objects ready for use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ err := c.Provide(&Type1{Name: "I am an thing"})
 
 ### Provide an Maps, slices or arrays
 
-Dig is now able to support maps, slices and arrays as objects to
+Dig also support maps, slices and arrays as objects to
 resolve, or provided as a dependency to the constructor.
 
 ```go
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ BETA. Expect potential API changes.
 There are two ways to Provide an object:
 
 1. Provide a pointer to an existing object
+1. Provide a slice, map, array
 1. Provide a "constructor function" that returns one pointer (or interface)
 
 ### Provide a Constructor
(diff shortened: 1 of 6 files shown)


glibsm:
image

glibsm:
some comments should also change on the functions, as well as potentially readme

README.md
```diff
@@ -63,6 +64,28 @@ err := c.Provide(&Type1{Name: "I am an thing"})
 ```
 
 
+### Provide an Maps, slices or arrays
+
+Dig is now able to support maps, slices and arrays as objects to
```
breerly: Suggest:

"Dig also supports maps, slices, and arrays:"

No reason to say "is now able to" in the readme.

breerly:
@anuptalwalkar can you update the PR description with background and motivation for the PR? 

I want to resist adding additional features if we dont have a pressing need for them.

anuptalwalkar:
Will do. We have to write hacky way of wrapping slices into struct so that we can use them in dig (eg, `[]yarpc.Transports` slice) I would really like to avoid doing that.

breerly:
@anuptalwalkar oh yeah, i remember that. 

Can you point me to the test case that shows that scenario under this paradigm?

README.md
```diff
@@ -63,6 +64,28 @@ err := c.Provide(&Type1{Name: "I am an thing"})
...
+	typemap = map[string]int{}
+	typeslice = []int{1, 2, 3}
+	typearray = [2]string{"one", "two"}
+)
+err := c.ProvideAll(typemap, typeslice, typearray)
+
+var resolveslice []int
+err := c.Resolve(&resolveslice)
+
+c.Invoke(func(map map[string]int, slice []int, array [2]string) {
```
breerly: hmmm... this is actually kind of scary...

Am I reading this right - there can only be 1 dep with a `[]int` here?

anuptalwalkar:
not sure what you mean. I have added test cases in this PR.

PR #39, opened 2017-05-03
Added constructor details in error message

### LINKED ISSUE #25: Error message details from the container
[Errors](https://github.com/uber-go/dig/blob/master/container.go#L33-L37) defined in container right now are not useful in mentioning which constructor failed the execution. 

example, A useful error would be - 
"constructor return type must be a pointer, constructor: `func foo() *ReturnType`"

anuptalwalkar:
Fixes #25 
error message now has details of the constructor and function signature.

### DIFF
--- a/container.go
+++ b/container.go
@@ -117,7 +117,7 @@ func (c *Container) Provide(types ...interface{}) error {
 				return err
 			}
 		default:
-			return errors.Wrapf(errParamType, " for constructor %v", ctype)
+			return errors.Wrapf(errParamType, "for constructor %v", ctype)
 		}
 	}
 	return nil
--- a/container.go
+++ b/container.go
@@ -35,6 +35,9 @@ var (
 	errArgKind     = errors.New("constructor arguments must be pointers")
 
 	_typeOfError = reflect.TypeOf((*error)(nil)).Elem()
+
+	_forCtor   = "for constructor %v"
+	_invokeErr = "error invoking the function %v"
 )
 
 // New returns a new DI Container
@@ -58,7 +61,7 @@ func (c *Container) Invoke(t interface{}) error {
 	case reflect.Func:
 		args, err := c.Graph.ConstructorArguments(ctype)
 		if err != nil {
-			return errors.Wrapf(err, "error invoking function %v", t)
+			return errors.Wrapf(err, _invokeErr, t)
 		}
 		cv := reflect.ValueOf(t)
 
@@ -67,7 +70,7 @@ func (c *Container) Invoke(t interface{}) error {
 		count := len(values)
 		if count > 0 && values[count-1].Type() == _typeOfError {
 			if err, _ := values[count-1].Interface().(error); err != nil {
-				return errors.Wrapf(err, "error executing the function %v", ctype)
+				return errors.Wrapf(err, _invokeErr, ctype)
 			}
 			count--
 		}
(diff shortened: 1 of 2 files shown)


container.go
```diff
@@ -58,7 +58,7 @@ func (c *Container) Invoke(t interface{}) error {
 	case reflect.Func:
 		args, err := c.Graph.ConstructorArguments(ctype)
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "error invoking function %v", t)
```
alsamylkin: Is it going to print actual function, not an address?

container_test.go
```diff
@@ -72,7 +72,7 @@ func TestRegister(t *testing.T) {
 			err := c.Provide(tc.param)
 
 			if tc.err != nil {
-				require.EqualError(t, err, tc.err.Error())
+				require.Contains(t, err.Error(), tc.err.Error())
```
alsamylkin: Can we add a test that check, that function name is present in the error message?

container.go
anuptalwalkar: yes. its printing `ctype`, which is actual func signature

container_test.go
anuptalwalkar: Will do.

container_test.go
```diff
@@ -57,22 +57,22 @@ func TestRegister(t *testing.T) {
...
+		err   string
 	}{
-		{"non function", struct{}{}, errParamType},
-		{"wrong return count", noReturn, errReturnCount},
-		{"non pointer return", returnNonPointer, errReturnKind},
-		{"wrong parameters type", nonPointerParams, errArgKind},
-		{"pointer Provide", &struct{}{}, nil},
+		{"non function", struct{}{}, "for constructor struct {}: " + errParamType.Error()},
+		{"wrong return count", noReturn, "for constructor func(): " + errReturnCount.Error()},
+		{"non pointer return", returnNonPointer, "for constructor func() dig.S: " + errReturnKind.Error()},
```
alsamylkin: where is `returnNonPointer` in the error message?

anuptalwalkar: no, that is not the intention. The intention was to identify what func call is the cause. e.g., knowing "func() dig.S" returns no pointer is more valuable.

alsamylkin: We can do better:
```golang
runtime.FuncForPC(reflect.ValueOf(functionOnj).Pointer()).Name()
```

container.go
```diff
@@ -117,7 +117,7 @@ func (c *Container) Provide(types ...interface{}) error {
 				return err
 			}
 		default:
-			return errParamType
+			return errors.Wrapf(errParamType, "for constructor %v", ctype)
```
glibsm: "for constructor %v" is used a lot, can we const-ify that?

container.go
```diff
@@ -76,11 +76,11 @@ func (c *Container) Invoke(t interface{}) error {
 			case reflect.Slice, reflect.Array, reflect.Map, reflect.Ptr, reflect.Interface:
 				c.Graph.InsertObject(values[i])
 			default:
-				return errors.Wrapf(errReturnKind, "%v", ctype)
+				return errors.Wrapf(errReturnKind, "error invoking function %v", ctype)
 			}
 		}
 	default:
-		return errParamType
+		return errors.Wrapf(errParamType, "error invoking function %v", ctype)
```
glibsm: same question here

PR #41, opened 2017-05-04
Remove *All APIs from dig

### LINKED ISSUE #40: Remove *All funcs
Background #34 

In order to provide a tight API surface, let's remove the `*All` functions, leaving just:

* `Provide`
* `Resolve`
* `Invoke`

This will mean combining the `ProvideAll` and `ResolveAll` into `Provide` and `Resolve`, so that those functions deal with variadic arguments so they can operate on one or many deps.

anuptalwalkar:
resolves #40 
`Provide`, `Resolve` APIs support variadic arguments to reduce API surface area to `Provide`, `Resolve` and `Invoke`

### DIFF
--- a/container.go
+++ b/container.go
@@ -84,35 +84,42 @@ func (c *Container) Invoke(t interface{}) error {
 	return nil
 }
 
-// Provide an object in the Container
+// Provide registers all the provided args in the Container
 //
 // The provided argument must be a function that accepts its dependencies as
 // arguments and returns one or more results, which must be a pointer type, map, slice or an array.
 // The function may optionally return an error as the last argument.
-func (c *Container) Provide(t interface{}) error {
-	ctype := reflect.TypeOf(t)
-	switch ctype.Kind() {
-	case reflect.Func:
-		switch ctype.NumOut() {
-		case 0:
-			return errReturnCount
-		case 1:
-			objType := ctype.Out(0)
-			if objType.Kind() != reflect.Ptr && objType.Kind() != reflect.Interface {
-				return errReturnKind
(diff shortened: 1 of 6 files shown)


container.go
```diff
@@ -83,85 +83,72 @@ func (c *Container) Invoke(t interface{}) error {
...
-	switch ctype.Kind() {
-	case reflect.Func:
-		switch ctype.NumOut() {
-		case 0:
-			return errReturnCount
-		case 1:
-			objType := ctype.Out(0)
-			if objType.Kind() != reflect.Ptr && objType.Kind() != reflect.Interface {
-				return errReturnKind
+func (c *Container) Provide(types ...interface{}) error {
```
alsamylkin: This means DIG will not support any tagging with names in Provide in the future, right?

breerly: That's right - this API will become single purpose for the time being.

container.go
```diff
@@ -83,85 +83,72 @@ func (c *Container) Invoke(t interface{}) error {
...
+			case 1:
+				objType := ctype.Out(0)
+				if objType.Kind() != reflect.Ptr && objType.Kind() != reflect.Interface {
+					return errReturnKind
+				}
 			}
+			if err := c.Graph.InsertConstructor(t); err != nil {
+				return err
+			}
+		case reflect.Slice, reflect.Array, reflect.Map, reflect.Ptr:
```
alsamylkin: Array is more a value type not a pointer like Slice/Map/Interface, not sure if it should be here.

alsamylkin: surprised why channels are not there.

PR #42, opened 2017-05-04
Invoke to propagate error to caller

anuptalwalkar:
If called function returns an error, we are trying to insert it into the graph (which fails because error is not a pointer type). We should rather just return the error back to the caller.

### DIFF
--- a/container.go
+++ b/container.go
@@ -67,6 +67,9 @@ func (c *Container) Invoke(t interface{}) error {
 
 		if len(values) > 0 {
 			err, _ := values[len(values)-1].Interface().(error)
+			if err != nil {
+				return errors.Wrapf(err, "Error executing the function %v", ctype)
+			}
 			for _, v := range values {
 				switch v.Type().Kind() {
 				case reflect.Ptr:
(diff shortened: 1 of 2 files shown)


container.go
```diff
@@ -67,6 +67,9 @@ func (c *Container) Invoke(t interface{}) error {
 
 		if len(values) > 0 {
 			err, _ := values[len(values)-1].Interface().(error)
```
alsamylkin: coalesce in one line if?

container_test.go
```diff
@@ -384,6 +385,15 @@ func TestInvokeAndRegisterFailure(t *testing.T) {
 	require.Contains(t, err.Error(), "constructor return type must be a pointer")
 }
 
+func TestInvokeReturnedError(t *testing.T) {
+	t.Parallel()
```
alsamylkin: one more test, when error is nil?

anuptalwalkar: Added

container.go
```diff
@@ -66,16 +66,17 @@ func (c *Container) Invoke(t interface{}) error {
 		values := cv.Call(args)
 
 		if len(values) > 0 {
-			err, _ := values[len(values)-1].Interface().(error)
+			if err, _ := values[len(values)-1].Interface().(error); err != nil {
+				return errors.Wrapf(err, "Error executing the function %v", ctype)
+			}
 			for _, v := range values {
```
alsamylkin: we shouldn't register an error here: just
```golang
for _, v := range values[:len(values]
```

Issue #43, opened 2017-05-04
DIG shouldn't register the last element of a constructor if it is an error


alsamylkin:
https://github.com/uber-go/dig/blob/master/internal/graph/node.go#L126

anuptalwalkar:
We are currently registering error type with nil value (which succeeds since nil error type is treated as `reflect.Interface`. If an actual error occurs, we throw it back to the caller. Regardless we need to add fix to dig to not register `error` object.

PR #44, opened 2017-05-05
Fix error handling for constructors and Invoke

anuptalwalkar:
#43 
Provide, Invoke don't register type `error` which is recommended to be last parameter in the function return.

### DIFF
--- a/container.go
+++ b/container.go
@@ -71,14 +71,12 @@ func (c *Container) Invoke(t interface{}) error {
 			}
 			count--
 		}
-		if count > 0 {
-			for i := 0; i < count; i++ {
-				switch values[i].Type().Kind() {
-				case reflect.Slice, reflect.Array, reflect.Map, reflect.Ptr, reflect.Interface:
-					c.Graph.InsertObject(values[i])
-				default:
-					return errors.Wrapf(errReturnKind, "%v", ctype)
-				}
+		for i := 0; i < count; i++ {
+			switch values[i].Type().Kind() {
+			case reflect.Slice, reflect.Array, reflect.Map, reflect.Ptr, reflect.Interface:
+				c.Graph.InsertObject(values[i])
+			default:
+				return errors.Wrapf(errReturnKind, "%v", ctype)
 			}
 		}
 	default:
--- a/container.go
+++ b/container.go
@@ -64,15 +64,20 @@ func (c *Container) Invoke(t interface{}) error {
 
 		// execute the provided func
 		values := cv.Call(args)
-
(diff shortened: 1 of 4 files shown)


container.go
```diff
@@ -64,15 +64,20 @@ func (c *Container) Invoke(t interface{}) error {
...
+			if values[count-1].Type() == _typeOfError {
+				if err, _ := values[count-1].Interface().(error); err != nil {
+					return errors.Wrapf(err, "Error executing the function %v", ctype)
+				}
+				count--
 			}
-			for _, v := range values {
-				switch v.Type().Kind() {
+		}
+		if count > 0 {
```
alsamylkin: combine two ifs?

container.go
```diff
@@ -64,15 +64,20 @@ func (c *Container) Invoke(t interface{}) error {
...
 		values := cv.Call(args)
-
-		if len(values) > 0 {
-			if err, _ := values[len(values)-1].Interface().(error); err != nil {
-				return errors.Wrapf(err, "Error executing the function %v", ctype)
+		count := len(values)
+		if count > 0 {
+			if values[count-1].Type() == _typeOfError {
+				if err, _ := values[count-1].Interface().(error); err != nil {
+					return errors.Wrapf(err, "Error executing the function %v", ctype)
```
alsamylkin: lowercase errors, please..

internal/graph/node.go
```diff
@@ -114,18 +114,22 @@ func (n *funcNode) value(g *Graph, objType reflect.Type) (reflect.Value, error)
...
-	// if last value is an error, it is returned as a separate argument, otherwise nil
-	err, _ = values[len(values)-1].Interface().(error)
-
-	for _, v := range values {
-		if objType == v.Type() {
-			return v, err
+	count := len(values)
+	if count > 0 {
+		// check if last argument is an error
+		if values[count-1].Type() == _typeOfError {
```
alsamylkin: isn't it already checked in Register()?

container_test.go
```diff
@@ -319,6 +319,8 @@ func TestConstructorErrors(t *testing.T) {
 			var p1 *FlakyParent
 			err := c.Resolve(&p1)
 			if tt.wantErr != "" {
+				var registeredError *error
```
alsamylkin: What about invoke tests?

internal/graph/node.go
anuptalwalkar: these are returned values from constructor execution, so need to make sure we are not adding them to dig.

Issue #46, opened 2017-05-15
Drop all the locks?


glibsm:
Now that dig is completely hidden from the user's eye through fx, I think one of the suggestions was to drop all the locks, I'm guessing with the assumption that it makes the dig implementation code simpler.

I'm all for making dig internals as clear as possible, but please let me know your thoughts around this @breerly @anuptalwalkar

breerly:
Seeing how we intend dig to be used to power serial object-graph resolution at process-startup time, I'm in favor of simplifying.

If we do need it, great. Until then, I'm ok w leaning on YAGNI. I'd like Anup to be onboard with this in spirit.

breerly:
What I meant was, I want to make sure Anup is onboard in spirit before we go one way or another :)

anuptalwalkar:
Well, locks are not that complex right now and hidden from front APIs. I am ok with removing this as long as we make sure not to expose dig outside, or to multiple go routines. Of course we can add it back if needs be:)

PR #49, opened 2017-05-16
Updated Provide ctors with common returns types to return error

### LINKED ISSUE #47: conflicting constructor registration with matching return types
Today, constructors are registered (provided) in sequential order in which they are received and can cause inconsistent behavior. eg, registering following constructors will cause different behavior depending on the order in which they are registered - 
```
ctor1(a *A, b *B) (x *x, y *Y)
ctor2() (x *x)
```
result of registration - 
```go 
//ctor1 registered first
y -> ctor1 
x -> ctor2 (registering ctor2 will override x -> ctor1)
 OR 
//ctor2 registered first -
y -> ctor1 
x -> ctor1 (registering ctor1 will override x -> ctor2)
```
Same issue occurs during call to ```Invoke``` as today dig just overwrites the values. 

Thoughts?
1. continue overwrite and explicitly state as in the document
2. Avoid overwriting constructors and invoke function. If object exist in dig cache, ignore. This will allow partial execution in any remaining objects that are not resolved. 
3. Panic on providing constructor or invoking function if existing object type already has a funcNode or objNode associated with it.

anuptalwalkar:
resolves #47 
Providing constructors with common return types will now result in an error. For example- 
```
ctor1(a *A, b *B) (x *x)
ctor2() (a *A)
```
Provide(ctor1) - success
Provide(ctor2) - returns an error -> `"ctor: func() (*X, error), object type: *X: node already exist for the constructor"`

### DIFF
--- a/container_test.go
+++ b/container_test.go
@@ -438,20 +438,6 @@ func TestProvide(t *testing.T) {
 	require.NotNil(t, p1.c1.gc1, "Grandchild1 must have been registered")
 }
 
-func TestConcurrentAccess(t *testing.T) {
-	t.Parallel()
-	c := New()
-
-	for i := 0; i < 10; i++ {
-		go func() {
-			require.NoError(t, c.Provide(NewGrandchild1))
-
-			var gc1 *Grandchild1
-			require.NoError(t, c.Resolve(&gc1))
-		}()
-	}
-}
-
 func TestCycles(t *testing.T) {
 	t.Parallel()
 	c := New()
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -30,8 +30,8 @@ import (
 )
 
 var (
-	errArgKind = errors.New("constructor arguments must be pointers")
-
+	errArgKind   = errors.New("constructor arguments must be pointers")
+	errRetNode   = errors.New("node already exist for the constructor")
 	_typeOfError = reflect.TypeOf((*error)(nil)).Elem()
 )
 
@@ -102,6 +102,10 @@ func (g *Graph) InsertConstructor(ctor interface{}) error {
 		objTypes[i] = ctype.Out(i)
 	}
 
+	if err := g.ValidateReturnTypes(ctype); err != nil {
+		return err
+	}
+
 	nodes := make([]node, count, count)
 	for i := 0; i < count; i++ {
 		nodes[i] = node{
@@ -140,6 +144,18 @@ func (g *Graph) InsertConstructor(ctor interface{}) error {
 		return errors.Wrapf(cycleErr, "unable to Provide %v", objTypes)
 	}
 
+	return nil
+}
+
+// ValidateReturnTypes validates if ctor's return type is already insterted in the graph
+func (g *Graph) ValidateReturnTypes(ctype reflect.Type) error {
+	g.RLock()
+	defer g.RUnlock()
+	for i := 0; i < ctype.NumOut(); i++ {
+		if _, ok := g.nodes[ctype.Out(i)]; ok {
+			return errors.Wrapf(errRetNode, "ctor: %v, object type: %v", ctype, ctype.Out(i))
+		}
+	}
 	return nil
 }
 
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -30,9 +30,10 @@ import (
 )
 
 var (
-	errArgKind   = errors.New("constructor arguments must be pointers")
-	errRetNode   = errors.New("node already exist for the constructor")
-	_typeOfError = reflect.TypeOf((*error)(nil)).Elem()
+	errArgKind     = errors.New("constructor arguments must be pointers")
+	errRetNode     = errors.New("node already exist for the constructor")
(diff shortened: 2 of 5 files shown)


internal/graph/graph_test.go
```diff
@@ -111,6 +111,17 @@ func TestGraphString(t *testing.T) {
...
+func TestCtorConflicts(t *testing.T) {
+	t.Parallel()
+	g := NewGraph()
+
+	err := g.InsertConstructor(threeObjects)
+	require.NoError(t, err)
+
+	err = g.InsertConstructor(oneObject)
+	require.Contains(t, err.Error(), "ctor: func() (*graph.Child1, error), object type: *graph.Child1: node already exist for the constructor")
+}
```
glibsm: What happens if you write a test to provide `oneObject`, then `oneObject` again?
The current functionality is that it will override, but I think with this code it will error out?

container_test.go
```diff
@@ -438,20 +438,6 @@ func TestProvide(t *testing.T) {
 	require.NotNil(t, p1.c1.gc1, "Grandchild1 must have been registered")
 }
 
-func TestConcurrentAccess(t *testing.T) {
```
alsamylkin: Why is it removed?

internal/graph/graph_test.go
```diff
@@ -111,6 +111,17 @@ func TestGraphString(t *testing.T) {
 	require.Contains(t, g.String(), "*graph.Parent12 -> (object) obj: *graph.Parent12")
 }
 
+func TestCtorConflicts(t *testing.T) {
+	t.Parallel()
+	g := NewGraph()
+
+	err := g.InsertConstructor(threeObjects)
```
alsamylkin: Can you try to insert this constructor ?:)
```golang
func () (*Child1, *Child1, *Child3, error) {
	return &Child1{}, &Child1{}, &Child3{}, nil
}
```

internal/graph/graph_test.go
anuptalwalkar: Yes. Providing same constructor will cause an error.

container_test.go
anuptalwalkar: The test is not providing any values. We are removing the locks in #46 anyway.

internal/graph/graph_test.go
anuptalwalkar: sure.

internal/graph/graph_test.go
```diff
@@ -111,6 +111,22 @@ func TestGraphString(t *testing.T) {
...
 }
 
+func TestCtorConflicts(t *testing.T) {
+	t.Parallel()
+	g := NewGraph()
+
+	err := g.InsertConstructor(func() (*Child1, *Child1, *Child3, error) {
+		return &Child1{}, &Child1{}, &Child3{}, nil
+	})
+	require.NoError(t, err)
```
alsamylkin: it is not the issue, that is supposed to be solved by this PR?

anuptalwalkar: what do you mean? First InsertConstructor will always succeed since return objects are not present in constructor, followed by the error (line 123 t0 127)

container_test.go
alsamylkin: We should do this in the PR where we remove locks, otherwise it hard to track commits..
e.g. you are still using locks in `ValidateReturnTypes`

internal/graph/graph.go
```diff
@@ -30,9 +30,10 @@ import (
 )
 
 var (
-	errArgKind = errors.New("constructor arguments must be pointers")
-
-	_typeOfError = reflect.TypeOf((*error)(nil)).Elem()
+	errArgKind     = errors.New("constructor arguments must be pointers")
+	errRetNode     = errors.New("node already exist for the constructor")
+	errCtorRetNode = errors.New("constructor return types must be distinct")
```
alsamylkin: Can we reuse errRetNode instead?

internal/graph/graph.go
```diff
@@ -143,6 +148,24 @@ func (g *Graph) InsertConstructor(ctor interface{}) error {
 	return nil
 }
 
+// ValidateReturnTypes validates if ctor's return type is already insterted in the graph
+func (g *Graph) ValidateReturnTypes(ctype reflect.Type) error {
+	g.RLock()
+	defer g.RUnlock()
+	objMap := make(map[reflect.Type]bool)
```
alsamylkin: map[reflect.Type]struct{} uses 2 times less memory!:)

container_test.go
anuptalwalkar: What I meant was this test is repeatedly Providing single constructor which is a failure with current PR since it now results in an error. Rather than adding efforts on fixing this test just to remove in the follow up PR, Im simply removing it here.

internal/graph/graph.go
anuptalwalkar: Its a throwaway map for start up. I am pretty sure we can live with a boolean rather than handling empty structs.:)

alsamylkin: Well, at least, you can use the ctype.NumOut capacity in advance.

PR #50, opened 2017-05-17
Remove locks from graph for simplicity

### LINKED ISSUE #46: Drop all the locks?

anuptalwalkar:
resolves #46

### DIFF
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -24,7 +24,6 @@ import (
 	"bytes"
 	"fmt"
 	"reflect"
-	"sync"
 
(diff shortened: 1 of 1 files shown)


glibsm:
`go test -race` still works right? :)

anuptalwalkar:
yep, make also runs test with race.

PR #52, opened 2017-05-17
A pass to clean up the readme somewhat

glibsm:
- Add a Container section to introduce the concept a little bit
- Clean up some small spelling errors
- Remove a duplicate section (bad merge earlier)

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,13 +58,13 @@ type Type1 struct {
 }
 
 c := dig.New()
-err := c.Provide(&Type1{Name: "I am an thing"})
+err := c.Provide(&Type1{Name: "I am a thing"})
 // dig container is now able to provide *Type1 as a dependency
 // to other constructors that require it.
 ```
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,9 +13,32 @@ For runnable examples, see the [examples directory](examples/).
...
+	c := dig.New()
+	// dig container `c` is ready to use!
+}
+```
+
+Objects in the container are identified by their `reflect.Type` and **everything
+is treated as a singleton**, meaning there can be only one object in the graph
+of a given type.
+
+For more advanced usecases, consider using a factory pattern. That is to say,
```
anuptalwalkar: "use cases"

anuptalwalkar: Do we want to lead users to factory pattern?

glibsm: Yes, we want to hint that you can solve multi-types through OO, i.e. a factory. More fidelity can be added into it later

Issue #48, opened 2017-05-16
Invoke on provided/unprovided ctors


anuptalwalkar:
Calling `Invoke` on provided and unprovided funcs can result in executing constructors more than once and overwriting the graph with the resulted return objects.
for provided constructors - 
 ```
ctor1(a *A, b *B) (x *x)
ctor2() (a *A)
```
Invoking constructors in order will result in ctor1 resolving *A and *B. 
Invoking ctor2 will result in resolving *A again.

**Comment by anuptalwalkar** [comment]
Here, execution is not the problem since the call can contain critical code, or functions that we Invoke multiple times. We have few choices - 
1. Don't Invoke function if return type is registered in the graph. `Invoke` will ignore the function and continue.

2. Error out on `Invoke` call if return types are already registered. (However, this will enforce required order in which functions must be Invoke'd.

3. `Invoke` the function. Register only values that are absent in the graph. Ignore already registered values and avoid overwriting.

4. Execute Invoke function every time and only fail on provided option. We can introduce `InvokeOption` to perform following - 
- Provide return types on function execution, overwrite is allowed (`Provide=true`)
- Ignore return types on function execution (`Provide=false`)
- Force an `error` if return type is already provided to the graph (Force = true)

As a default, 
`Provide = true` - always overwrite if not specified
`Force = false` - overwrite and don't return error

**Comment by anuptalwalkar** [comment]
@breerly @glibsm thoughts? 
I am split between Option 1 and 4, but leaning towards 4.
Option 1 will keep a clean API, except it will add reentrant characteristic to Invoke. This means your function might not execute if return parameter is registered in dig, and will lead to strange behavior.
Option 4 is more explicit with InvokeOptions (similar to zap's options) and can extend Invoke's use case in the future.

**Comment by alsamylkin** [comment]
I'd rather keep the Invoke as is, but provide a wrapper, dig.Once:
```golang
func Once(ctor dig.Fn) dig.Fn {
  tp := reflect.TypeOf(ctor)
  if tp.Kind() != reflect.Func {
    panic("expect function to listen")
  }

  ctorVal := reflect.ValueOf(ctor)
  listenerVal := reflect.ValueOf(listener)
  execute := true
  outs := []reflect.Value
  res := reflect.MakeFunc(tp, func(args []reflect.Value) []reflect.Value {
    if execute {
       outs = ctorVal.Call(args)
       execute = false
    }
    return outs
  })

  return res.Interface()
}
```

breerly:
[...]

PR #55, opened 2017-05-17
[Prototype] Add InvokeOnce functionality to container

anuptalwalkar:
#48 
Cooked up some code for `InvokeOnce` that will be helpful for discussion -

When `InvokeOnce` is called, it performs additional validations to check if return object types are already registered and cached in the graph. If object is cached, `InvokeOnce` skips `Invoke` and returns `ErrInvokeOnce` to the caller. 

It is up to the caller to choose to panic or ignore the error message knowing that method has not been invoked since object is already resolved in the graph.

### DIFF
--- a/container.go
+++ b/container.go
@@ -61,7 +61,7 @@ func (c *Container) InvokeOnce(t interface{}) error {
 	if ctype.Kind() != reflect.Func {
 		return errors.Wrapf(errParamType, _invokeErr, ctype)
 	}
-	if err := c.ValidateReturnTypes(ctype, true); err != nil {
+	if err := c.ValidateInvokeReturnTypes(ctype); err != nil {
 		return ErrInvokeOnce
 	}
 	return c.Invoke(t)
--- a/container.go
+++ b/container.go
@@ -53,7 +53,7 @@ type Container struct {
 	graph.Graph
 }
 
-// InvokeOnce only allows function invokation once to register the
+// InvokeOnce only allows function invocation once to register the
 // return types. If return types are already registered, specific error
 // is returned to the caller
 func (c *Container) InvokeOnce(t interface{}) error {
--- a/container.go
+++ b/container.go
@@ -34,6 +34,9 @@ var (
 	errReturnKind  = errors.New("constructor return type must be a pointer")
 	errArgKind     = errors.New("constructor arguments must be pointers")
 
+	// ErrInvokeOnce is returned if function is already invoked
+	ErrInvokeOnce = errors.New("function is already invoked")
+
 	_typeOfError = reflect.TypeOf((*error)(nil)).Elem()
 
 	_forCtor   = "for constructor %v"
@@ -48,6 +51,20 @@ func New() *Container {
 // Container facilitates automated dependency resolution
 type Container struct {
 	graph.Graph
+}
+
+// InvokeOnce only allows function invokation once to register the
+// return types. If return types are already registered, specific error
+// is returned to the caller
+func (c *Container) InvokeOnce(t interface{}) error {
+	ctype := reflect.TypeOf(t)
+	if ctype.Kind() != reflect.Func {
+		return errors.Wrapf(errParamType, _invokeErr, ctype)
+	}
+	if err := c.ValidateReturnTypes(ctype, true); err != nil {
+		return ErrInvokeOnce
+	}
+	return c.Invoke(t)
 }
 
 // Invoke the function and resolve the dependencies immidiately without providing the
--- a/container_test.go
+++ b/container_test.go
@@ -384,6 +384,11 @@ func TestInvokeOnceFailOnResolvedTypes(t *testing.T) {
 	c.Provide(newT1, newT2, newT3)
 	require.NoError(t, c.InvokeOnce(newT1))
 	assert.Equal(t, c.InvokeOnce(newT2), ErrInvokeOnce)
+
+	c.Reset()
+	c.Provide(newT1, newT2, newT3)
+	require.NoError(t, c.InvokeOnce(newT2))
+	require.NoError(t, c.InvokeOnce(newT1))
 }
 
 func TestInvokeAndRegisterSuccess(t *testing.T) {
--- a/container_test.go
+++ b/container_test.go
@@ -373,6 +373,10 @@ func TestInvokeOnce(t *testing.T) {
 	})
 	assert.NoError(t, err)
 	assert.NotNil(t, c1)
+
+	type empty struct{}
+	require.NoError(t, c.InvokeOnce(func() *empty { return &empty{} }))
+	assert.Equal(t, c.InvokeOnce(func() *empty { return &empty{} }), ErrInvokeOnce)
 }
 
 func TestInvokeAndRegisterSuccess(t *testing.T) {
--- a/container_test.go
+++ b/container_test.go
@@ -349,6 +349,32 @@ func TestInvokeSuccess(t *testing.T) {
 	assert.NotNil(t, c1)
 }
 
+func TestInvokeOnce(t *testing.T) {
+	t.Parallel()
+	c := New()
+
+	err := c.Provide(
+		NewParent1,
+		NewChild1,
+		NewGrandchild1,
+	)
+	assert.NoError(t, err)
+	var c1 *Child1
+
+	err = c.InvokeOnce(NewParent1)
+	assert.NoError(t, err)
+
+	err = c.InvokeOnce(NewParent1)
+	assert.Equal(t, ErrInvokeOnce, err)
+
+	err = c.InvokeOnce(func(p1 *Parent1) {
+		require.NotNil(t, p1)
+		c1 = p1.c1
+	})
+	assert.NoError(t, err)
+	assert.NotNil(t, c1)
+}
+
 func TestInvokeAndRegisterSuccess(t *testing.T) {
 	t.Parallel()
 	c := New()
--- a/container_test.go
+++ b/container_test.go
@@ -379,6 +379,13 @@ func TestInvokeOnce(t *testing.T) {
 	assert.Equal(t, c.InvokeOnce(func() *empty { return &empty{} }), ErrInvokeOnce)
 }
 
+func TestInvokeOnceFailOnResolvedTypes(t *testing.T) {
+	c := New()
+	c.Provide(newT1, newT2, newT3)
+	require.NoError(t, c.InvokeOnce(newT1))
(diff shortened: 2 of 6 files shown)


container.go
```diff
@@ -50,6 +53,20 @@ type Container struct {
 	graph.Graph
 }
 
+// InvokeOnce only allows function invokation once to register the
+// return types. If return types are already registered, specific error
+// is returned to the caller
+func (c *Container) InvokeOnce(t interface{}) error {
```
alsamylkin: Not going to work:
```golang
func TestInvokeOnce(t *testing.T) {
	t.Parallel()
	c := New()

	type empty struct{}

	require.NoError(t, c.InvokeOnce(func() *empty{return &empty{}}))
	require.NoError(t, c.InvokeOnce(func() *empty{return &empty{}}))
}
```

anuptalwalkar: What you mean its not going to work? The test you mentioned will fail as It expects `ErrInvokeOnce` on second invoke. I'll add your test case.

container.go
```diff
@@ -50,6 +53,20 @@ type Container struct {
 	graph.Graph
 }
 
+// InvokeOnce only allows function invokation once to register the
```
alsamylkin: typo: invocation

container.go
```diff
@@ -34,6 +34,9 @@ var (
 	errReturnKind  = errors.New("constructor return type must be a pointer")
 	errArgKind     = errors.New("constructor arguments must be pointers")
 
+	// ErrInvokeOnce is returned if function is already invoked
+	ErrInvokeOnce = errors.New("function is already invoked")
```
alsamylkin: why do you want to export only this error?

internal/graph/graph.go
```diff
@@ -94,7 +94,7 @@ func (g *Graph) InsertConstructor(ctor interface{}) error {
 		objTypes[i] = ctype.Out(i)
 	}
 
-	if err := g.ValidateReturnTypes(ctype); err != nil {
+	if err := g.ValidateReturnTypes(ctype, false); err != nil {
```
alsamylkin: why does it need to be false?

container.go
anuptalwalkar: me and @glibsm were talking about whats the expected behavior on InvokeOnce. If InvokeOnce fail, should we silently continue or fail? There are some cases when we just want to continue. strongly typed error `dig.ErrInvokeOnce` can allow callers to decide whether they just want to log and continue or panic with an equals check.

internal/graph/graph.go
anuptalwalkar: Just trying to reuse `ValidateReturnTypes` method. checkCachedObjects=false ignores cached objects for ctor provide and builds the graph to resolve later.

internal/graph/graph_test.go
```diff
@@ -128,6 +128,20 @@ func TestCtorConflicts(t *testing.T) {
 	require.Contains(t, err.Error(), "ctor: func() (*graph.Child1, *graph.Child1, error), object type: *graph.Child1: node already exist for the constructor")
 }
 
+func TestValidateReturnTypes(t *testing.T) {
```
glibsm: test should describe the behavior at hand, i.e. `TestConstructorOverrideReturnsError`

internal/graph/graph.go
anuptalwalkar: for Invoke we only want to check if object is cached or not, and ignore provided unresolved constructor.

internal/graph/graph.go
```diff
@@ -137,16 +137,23 @@ func (g *Graph) InsertConstructor(ctor interface{}) error {
...
-func (g *Graph) ValidateReturnTypes(ctype reflect.Type) error {
+// ValidateReturnTypes validates if Invoke func's return type is Provided to the graph
+// checkCachedObjects=true additionally checks if objects are resolved and cached in the graph
+func (g *Graph) ValidateReturnTypes(ctype reflect.Type, checkCachedObjects bool) error {
 	objMap := make(map[reflect.Type]bool, ctype.NumOut())
 	for i := 0; i < ctype.NumOut(); i++ {
 		objType := ctype.Out(i)
 		if _, ok := g.nodes[objType]; ok {
-			return errors.Wrapf(errRetNode, "ctor: %v, object type: %v", ctype, ctype.Out(i))
+			if checkCachedObjects {
```
glibsm: when would you need this vs not need this, since this is internal?

anuptalwalkar: Added comment. This is needed when we Invoke already provided constructor. At that point, we just want to check if object is resolved and cached or not.

internal/graph/graph.go
```diff
@@ -137,16 +137,27 @@ func (g *Graph) InsertConstructor(ctor interface{}) error {
 	return nil
 }
 
-// ValidateReturnTypes validates if ctor's return type is already insterted in the graph
-func (g *Graph) ValidateReturnTypes(ctype reflect.Type) error {
+// ValidateReturnTypes validates if Invoke func's return type is Provided to the graph
+// checkCachedObjects=true ensures to throw an error only when the node is already
+// resolved and cached.
+//
+func (g *Graph) ValidateReturnTypes(ctype reflect.Type, checkCachedObjects bool) error {
```
glibsm: This function really confuses me now. I know that there is some code reuse going on, but we can we maybe split it into two separate ones?

Also, I feel like we should spell out, *exactly*, step-by-step what Invoke actually does because the code for everything involved in it no longer fits cleanly into my head.

anuptalwalkar: alright. I'll add some clarity and refactor.

container.go
```diff
@@ -50,6 +53,20 @@ type Container struct {
...
+// is returned to the caller
+func (c *Container) InvokeOnce(t interface{}) error {
+	ctype := reflect.TypeOf(t)
+	if ctype.Kind() != reflect.Func {
+		return errors.Wrapf(errParamType, _invokeErr, ctype)
+	}
+	if err := c.ValidateReturnTypes(ctype, true); err != nil {
+		return ErrInvokeOnce
+	}
+	return c.Invoke(t)
```
glibsm: `InvokeOnce` should be recursive, i.e. it should InvokeOnce other things as well, otherwise this doesn't solve the problem we need addressing.

What we want to achieve is that the order of operations in fx.Start doesn't matter. In this case, we will still double-invoke things given a particular order.

anuptalwalkar: What do you mean by being recursive? There are no recursive calls with Invoke.
InvokeOnce is Invoking the function once, which is using `node.value()` to resolve dependencies recursively. InvokeOnce is just a wrapper around `Invoke` to make sure that object is not resolved multiple times. 

InvokeOnce returns `ErrInvokeOnce` which is used as a decider to skip double invokes.

fx.Start will be fixed with simple check - 
```go
if err := s.container.InvokeOnce(c); err != nil && err != dig.ErrInvokeOnce {
	return err
}
```

glibsm: This is what I want to know:

```
type type1 struct {}
type type2 struct{}
type type3 struct{}
func newT1(type2, type3) *type1 {}
func newT2() *type2 {}
func newT3() *type3{}

c := dig.New()
c.InvokeOnce(newT1) // calls newT2 and newT3 first to satisfy deps, then newT1
c.InvokeOnce(newT2) // <--- here, will actually execute T2?
```

And I'd actually really like a UT for that. To make sure all constructors are called only once.

anuptalwalkar: gotchu. `c.InvokeOnce(newT2)` will return `dig.ErrInvokeOnce` and not Invoke `newT2`. Will add test.

glibsm: Nice. What about the following scenario:

```
c := dig.New()
c.InvokeOnce(newT2) // once the T2
c.InvokeOnce(newT1) // <--- should use cached T2, but actually invoke T3
```

We still good in this case?

anuptalwalkar: yep. for newT1, T2 is cached and T3 is recursively called to resolve ctor via `node.value`. Will add this scenario.

container.go
```diff
@@ -50,6 +53,20 @@ type Container struct {
...
+// InvokeOnce only allows function invocation once to register the
+// return types. If return types are already registered, specific error
+// is returned to the caller
+func (c *Container) InvokeOnce(t interface{}) error {
+	ctype := reflect.TypeOf(t)
+	if ctype.Kind() != reflect.Func {
+		return errors.Wrapf(errParamType, _invokeErr, ctype)
+	}
+	if err := c.ValidateInvokeReturnTypes(ctype); err != nil {
+		return ErrInvokeOnce
```
alsamylkin: This completely hides the original error, which may be because of duplicate returns, not `function is already invoked`..

container_test.go
```diff
@@ -349,6 +349,48 @@ func TestInvokeSuccess(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.NotNil(t, c1)
+
+	type empty struct{}
+	require.NoError(t, c.InvokeOnce(func() *empty { return &empty{} }))
+	assert.Equal(t, c.InvokeOnce(func() *empty { return &empty{} }), ErrInvokeOnce)
+}
+
+func TestInvokeOnceFailOnResolvedTypes(t *testing.T) {
+	c := New()
```
alsamylkin: t.Parallel?

internal/graph/graph.go
```diff
@@ -137,16 +137,44 @@ func (g *Graph) InsertConstructor(ctor interface{}) error {
...
+// if the grah node of return type is resolved and cached.
+func (g *Graph) ValidateInvokeReturnTypes(ctype reflect.Type) error {
+	if err := g.checkDuplicateReturns(ctype); err != nil {
+		return err
+	}
 	for i := 0; i < ctype.NumOut(); i++ {
 		objType := ctype.Out(i)
 		if _, ok := g.nodes[objType]; ok {
-			return errors.Wrapf(errRetNode, "ctor: %v, object type: %v", ctype, ctype.Out(i))
+			if obj, ok := g.nodes[objType].(*objNode); ok && obj.cached {
```
alsamylkin: This does the same lookup in g.nodes, as in previous if above, just use a variable there.

container.go
anuptalwalkar: True. It does simplify the check and decision making for caller. Although the caller has access to the function signature to log it, or error out.

breerly:
Let's hold off on merging this - we're going to play with the an all-explicit API where this approach might not be necessary.

glibsm:
Fx will go with explicit ordering on start, so Once functionality is no longer desirable. We can bring it back in the future should there be a need for it.

PR #60, opened 2017-05-19
Added readme for Invoke

anuptalwalkar:


README.md
```diff
@@ -134,6 +134,50 @@ until the constructor for `config.Provider` can be fully satisfied.
 If resolution is not possible, for instance one of the required dependencies has
 does not have a constructor and doesn't appear in the graph, an error will be returned.
 
+## Invoke
+
+`Invoke` API executes the provided function and stores return results in the dig container.
```
Lupie: "return" or "returned"?

Lupie: The `Invoke` API

README.md
```diff
@@ -134,6 +134,50 @@ until the constructor for `config.Provider` can be fully satisfied.
...
+
+`Invoke` looks through the dig graph and resolves all the constructor parameters for execution.
+The return results, except the `error` object is inserted into the graph for further use.
```
Lupie: - comma after "object"
- is --> are
- it would be better to say _who_ or _what_ injects the results into the graph

README.md
```diff
@@ -134,6 +134,50 @@ until the constructor for `config.Provider` can be fully satisfied.
...
+To utilize `Invoke`, function needs following form -
```
Lupie: - utilize --> use
- function needs following form - --> make sure your function meets the following criteria:

README.md
```diff
@@ -134,6 +134,50 @@ until the constructor for `config.Provider` can be fully satisfied.
...
+
+1. function arguments are pointers, maps, slice or arrays
```
Lupie: - Delete "function" from all of these and say either "Arguments…" or "Its arguments" (we already know you're talking about a function)
- Capitalize the first letter in this numbered list

Lupie: - slice --> slices
- comma after "slices"

README.md
```diff
@@ -134,6 +134,50 @@ until the constructor for `config.Provider` can be fully satisfied.
...
+2. function arguments must be provided to dig
```
Lupie: Saying "must" here but not in #1, for example, makes it seem that the function arguments can be something other than pointers, maps, slices, or arrays. Every list item _must_ be followed, right? If so, change "must be" to "are".

README.md
```diff
@@ -134,6 +134,50 @@ until the constructor for `config.Provider` can be fully satisfied.
...
+3. function return parameters are optional
```
Lupie: 3. (Optional) Function has return parameters

README.md
```diff
@@ -134,6 +134,50 @@ until the constructor for `config.Provider` can be fully satisfied.
...
+4. provided function can be anonymous
```
Lupie: It might make sense to remove this from the list. Put it in the following paragraph instead.
Another option would be to turn this list into a table, where one column is either "Optional" or "Mandatory" and the other is the feature, like "Has return parameters."

README.md
```diff
@@ -134,6 +134,50 @@ until the constructor for `config.Provider` can be fully satisfied.
...
+
+For example, `Invoke` scenarios:
+
+Invoke anonymous function:
```
Lupie: Would it make sense to add "an" or "the" before "anonymous function"?

README.md
```diff
@@ -134,6 +134,50 @@ until the constructor for `config.Provider` can be fully satisfied.
...
+```go
+ // c := dig.New()
+// c.Provide... config.Provider, zap.Logger, etc
+err := c.Invoke(func(cfg *config.Provider, l *zap.Logger) error {
+  // function body
+	return nil
+})
+```
+
+Invoke recursively resolve dependencies:
```
Lupie: This sentence doesn't seem quite right. Do you mean "recursively resolved dependencies"?

anuptalwalkar: changing to - "`Invoke` recursively resolves all the dependencies"

glibsm:
@Lupie could you yay/nay the PR after the changes Anup made?

README.md
```diff
@@ -132,6 +132,50 @@ until the constructor for `config.Provider` can be fully satisfied.
 If resolution is not possible (for instance, if one of the required dependencies
 lacks a constructor and doesn't appear in the graph), `dig` returns an error.
 
+## Invoke
+
+The `Invoke` API executes the provided function and stores returned results in the dig container.
```
Lupie: I think we should monospace `dig` right?

README.md
```diff
@@ -132,6 +132,50 @@ until the constructor for `config.Provider` can be fully satisfied.
...
+
+`Invoke` looks through the dig graph and resolves all the constructor parameters for execution.
+At the end of method invokation, `Invoke` inserts return results except the `error` object, into the
```
Lupie: - invokation --> invocation
- return results or returned results?
- results except --> results, except

Also what does "end of method invocation" mean? Like when the method is fully invoked?

README.md
```diff
@@ -132,6 +132,50 @@ until the constructor for `config.Provider` can be fully satisfied.
...
+graph for further use.
+To use `Invoke`, make sure your function meets the following criteria:
+
+1. Arguments must be pointers, maps, slices, or arrays
+2. Arguments must be provided to dig
+3. (Optional) Function has return parameters
+
+For example, `Invoke` scenarios:
```
Lupie: These are example `Invoke` scenarios? Can you introduce what's going on a bit more? Like, "For example, see how `Invoke` works in…".

README.md
anuptalwalkar: Yes, it means when method is fully invoked. Any better way to write it?

Lupie: Yeah, how about "Once the method is invoked, `Invoke` inserts all return(ed?) results except the `error` object into the…"? It's a little weird because we're saying the `Invoke` method does something after the method is invoked. But I think it's OK. What do you think?

anuptalwalkar: Thats actually sounds better and clear. Will update.

PR #61, opened 2017-05-22
Added benchmarks for container API

anuptalwalkar:
```
Benchmark_CtorInvoke-8                          	 1000000	      2137 ns/op	     304 B/op	      10 allocs/op
Benchmark_CtorInvokeWithObjects-8               	 1000000	      1497 ns/op	     200 B/op	       6 allocs/op
Benchmark_InvokeCtorWithMapsSlicesAndArrays-8   	  500000	      2571 ns/op	     440 B/op	       9 allocs/op
Benchmark_CtorProvideAndResolve-8               	 1000000	      2183 ns/op	     320 B/op	      11 allocs/op
Benchmark_CtorResolve-8                         	 1000000	      1903 ns/op	     216 B/op	       7 allocs/op
Benchmark_ResolveCtors-8                        	  500000	      2252 ns/op	     344 B/op	      14 allocs/op
```

### DIFF
--- a/container_benchmark_test.go
+++ b/container_benchmark_test.go
@@ -1,0 +1,115 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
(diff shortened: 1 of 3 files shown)


container_benchmark_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+func benchInvokeFunc(g1 *Grandchild1) (*Parent1, error) {
+	return &Parent1{
+		c1: &Child1{
+			gc1: g1,
+		},
+	}, nil
+}
+
+func Benchmark_CtorInvoke(b *testing.B) {
```
alsamylkin: We need to save the results of benchmarks somewhere..

container_benchmark_test.go
```diff
@@ -0,0 +1,115 @@
...
+		&Grandchild1{},
+	)
+
+	b.ResetTimer()
+	for n := 0; n < b.N; n++ {
+		c.Invoke(benchInvokeFunc)
+	}
+}
+
+func Benchmark_InvokeCtorWithMapsAndSlices(b *testing.B) {
```
alsamylkin: Is it worth it to split maps and slices?
What about arrays?

glibsm:
I'd love to be able to see the cost of a fresh Invoke (without caching), i.e.:

```
b.PasueTimer()
c := dig.New()
c.Provide(a,b,c,d,e)
b.Unpause()
c.Invoke(func(a,b,c,d,e))
```

I want to see uncached state for each iteration loop, basically. The current benchmarks are not telling the full story, as the cost of the first run gets split across all the other runs as well

anuptalwalkar:
> I want to see uncached state for each iteration loop

Its going to be painfully slow :)

container_benchmark_test.go
anuptalwalkar: I will add final results to the readme.

anuptalwalkar:
updated description with benchmarks on uncached objects and ctors.

container_benchmark_test.go
```diff
@@ -0,0 +1,123 @@
...
+func Benchmark_CtorInvoke(b *testing.B) {
+	b.ResetTimer()
+	for n := 0; n < b.N; n++ {
+		b.StopTimer()
+		c := New()
+		c.Provide(
+			NewGrandchild1,
+		)
+		b.StartTimer()
+		c.Invoke(benchInvokeFunc)
```
alsamylkin: I think you should stop timer here, otherwise you are counting precious tics on ++ op :)

anuptalwalkar: we need timer for a timer :)

glibsm:
> Its going to be painfully slow :)

Yeah, but that's what we actually needed to know, heh. Just *how* slow

PR #63, opened 2017-05-23
[Prototype] Zero-value validation approach

anuptalwalkar:
Currently dig returns an error `dependency of type is not registered` if constructor argument fails to resolve during `Invoke` or `Resolve`. This change is introduced to resolve any absent arguments with zero value objects. Zero value objects are resolved during resolving constructor or invoking a function with arguments not yet present in the DIG container. 

When can it happen?
1. Constructors are provided out of order
2. Optional objects are not provided to the constructor

What will be the expected behavior?
1. No change in existing API on the container 
2. Invoke, resolving constructor never fails on missing constructor arguments in DIG container.
3. Constructor is provided with zero value argument during call to continue execution
4. Invoke/Resolve logs error message when zero value object is used
5. Zero value object is never cached.


anuptalwalkar:
@breerly @glibsm Thought I would just write some code and play with it. Added some description about the zero value behavior validation around it.

internal/graph/graph.go
```diff
@@ -204,7 +205,8 @@ func (g *Graph) ConstructorArguments(ctype reflect.Type) ([]reflect.Value, error
 			}
 			args[idx] = v
 		} else {
-			return nil, fmt.Errorf("%v dependency of type %v is not registered", ctype, arg)
+			log.Printf("Assigning zero value for arguments in %v. Dependency of type %v is not yet registered", ctype, arg)
```
anuptalwalkar: Not sure we need logging here, but without the log message the responsibility of knowing if an object is of zero value shifts to users and forces them to call and log `dig.Validate()`

glibsm:
Lets assume that we extract a validation library out that will solve common concerns in other go code as well (having to check nil for all the interfaces and pointers that are taken in).

What sort of a machinery can we then come up with to preserve error messaging. For instance, I would like to preserve `failed to resolve func(foo, bar) -> baz because foo dependency is missing`.

Would we use some sort of a special error format, would we catch panics from the constructor execution?

I'm trying to envision how a constructor such as this:

```go
func newBar(f *foo, b *bar) baz {
  vallie.Required(f, b)
}
```

Would be able to report the error through dig

akshayjshah:
Think there's still one missing change - we also need to remove the graph completeness check (`Graph.validateGraph`). I'll push a commit that fixes that and adds some end-to-end tests for the way `Provide`, `Invoke`, and `Resolve` work with missing deps.

container_test.go
```diff
@@ -551,10 +536,50 @@ func TestMultiObjectRegisterResolve(t *testing.T) {
...
 	require.NotNil(t, first, "Child1 must have been registered")
 	require.NotNil(t, second, "Child2 must have been registered")
 	require.NotNil(t, third, "Child3 must have been registered")
 }
+
+func TestZeroValueOnInvoke(t *testing.T) {
+	type missing struct{}
+	type present struct{}
+
+	t.Parallel()
```
breerly: Move this to top of this test

container_test.go
```diff
@@ -551,10 +536,50 @@ func TestMultiObjectRegisterResolve(t *testing.T) {
...
+	assert.NoError(t, c.Invoke(func(p *present, m *missing) {
+		assert.Nil(t, m, "expected zero value for missing dep")
+		assert.NotNil(t, p, "expected non-zero value for present dep")
+		called++
+	}), "unexpected failure invoking with a mix of present and missing deps")
+
+	assert.Equal(t, 2, called, "didn't run invokes")
+}
+
+func TestZeroValueOnProvide(t *testing.T) {
```
breerly: Add t.Parallel()

vallie/vallie.go
```diff
@@ -0,0 +1,30 @@
+// Package vallie stands in for a stand-alone validation library
+// purely for discussion purposes.
+//
+// Assuming this is a desired approach, more fidelity will be added
+// and vallie would be extracted into it's own repo
+package vallie
```
breerly: ''Tis begun :P

vallie/vallie.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"reflect"
+
```
anuptalwalkar: What is "vallie"? Its not intuitive and something that users will need to be reminded of. It sounds like a toy name. Why not just keep it on 'dig.Validate'?

breerly: I believe the intention here is to incubate a lib that would be moved out of this repo entirely, should we go down this road.

These are orthogonal concerns, so we shouldn't ship them together if we don't have to.

The actual name "vallie" is referring to a joke Glib, Akshay, and I made on a call - I don't think any of us feel strongly as to what the actual name is called, should we determine this is the right path to head down.

breerly:
Just verified this works as expected from the playgrounds POV - good stuff.

vallie/vallie.go
breerly: Ah, just noticed - read the docblock :P

glibsm: Yeah the name is a total placeholder. The reason we probably wouldn't want to keep it as part of dig is because this validation library could be very useful without any connection to dig. It's very common to check non-nils for pointers and interfaces when you write an exported function. It's similar to checking for nulls in java and the thought was to provoke a library capable of solving both problems, even if it's primary goal is to serve dig.

glibsm:
Still have to sleep on this, but the approach is growing on me.

vallie/vallie.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/pkg/errors"
+)
+
+var errZeroVal = errors.New("Object is a zero value")
+
+// Validate accepts objects to be validated from constructor and returns an error
+// if any of the provided object is a zero value object
+//
+// TODO: if Validate returns an error, how can we hook into the dig resolution
```
alsamylkin: What about actionable results? E.g. return 2 things: a constructor for unrecognized types and an error, so a user can simply register it if they want. Another option would be to return a slice of types for a user to set(may be even other than a zero value) and register manually.

internal/graph/graph.go
```diff
@@ -178,18 +178,6 @@ func (g *Graph) recursiveDetectCycles(n graphNode, l []string) error {
 	return nil
 }
 
-func (g *Graph) validateGraph(ct reflect.Type) (reflect.Value, error) {
```
anuptalwalkar: We can talk about this behavior in person, but I don't think we should make everything zero value - if absent. I was going for this behavior - 

1. `Provide, Resolve` happens  on objects or Provided constructor that always return objects in a restrictive environment. The graph needs to be complete in order to resolve anything and zero value means we loose any context on why the graph is incomplete. I strongly feel we should error out on any incompleteness of the graph during resolve. This way, we maintain graph's completeness and know during build time that all objects "can be resolved" when graph is built.
2. `Invoke` is somewhat impromptu execution of methods that are not part of the graph and needs to pull parameters from dig. Since its an impromptu invocation, `dig` graph can't control graph completeness and should resolve missing dependencies in real time (using zero values). 

Making everything zero value effectively allowing dig to not have control over graph's completeness. We will lead to dig container into unstable states very often (think just executing without Providing anything to dig), and I am not very sure we are ready for it.

glibsm:
Closing in favor of #66 approach.

We're going to go with embedding Param and Return objects to support optionals, named types, and potentially scoping in the future.

PR #66, opened 2017-05-24
[Prototype] Optional type

glibsm:
See `optional.go` for an illustration of what's going on. `type2` is optional and therefor there is no error during `Invoke`, since it gets a nil.

This adds a new interface called `DigOptional` and a type `dig.Optional` which can be embedded into a struct to signal an optional parameter.

This obviously doesn't address things like optional slices or maps (unless they are typealiased and this interface is satisfied).

Thoughts/comments?

### DIFF
--- a/examples/optional/optional.go
+++ b/examples/optional/optional.go
@@ -19,7 +19,7 @@ func main() {
 	c.Provide(&type1{})
 	if err := c.Invoke(func(t1 *type1, t2 *type2) {
 		fmt.Println("t1", t1) // will get actual instance of type1
-		fmt.Println("t2", t2) // note nil, yet dig doesn't not error
+		fmt.Println("t2", t2) // note nil and dig doesn't error out, just provides zero value
 	}); err != nil {
 		panic(err)
 	}
--- a/examples/optional/optional.go
+++ b/examples/optional/optional.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"fmt"
+
+	"go.uber.org/dig"
+)
+
+// type1 is a "required" dig type by default
+type type1 struct{}
+
+// type2 embeds a dig.Optional object to signal it's ok to ignore in params
+type type2 struct {
+	dig.Optional
+}
+
+func main() {
+	c := dig.New()
+	c.Provide(&type1{})
+	if err := c.Invoke(func(t1 *type1, t2 *type2) {
+		fmt.Println("t1", t1) // will get actual instance of type1
+		fmt.Println("t2", t2) // note nil, yet dig doesn't not error
+	}); err != nil {
+		panic(err)
+	}
+}
--- a/container.go
+++ b/container.go
@@ -50,6 +50,13 @@ type Container struct {
 	graph.Graph
 }
 
+// Optional struct implementing the DigOptional interface
+// Signaling to dig that it's ok to pass zero value into this parameter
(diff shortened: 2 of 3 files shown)


alsamylkin:
With this approach existing constructors that can accept nil/optionals should be updated.
The problem is that non user controlled types need to be aliased and correctly injected in the graph together with original types :(

examples/optional/optional.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"go.uber.org/dig"
+)
+
+// type1 is a "required" dig type by default
+type type1 struct{}
+
+// type2 embeds a dig.Optional object to signal it's ok to ignore in params
+type type2 struct {
+	dig.Optional
```
anuptalwalkar: This would work if struct definition is in control of module owners, but providing this struct with `dig.Optional` will be complicated since modules will not have control over where the objects coming from. For example, even `zap.Logger` can be optional in http module, and switched to std logger if it is not provided.

glibsm:
The main target audience for this would be module authors that expose types such as `Backend`, `[]Middleware`, etc.

You can still apply this approach on a per-constructor basis. Consider this:

```go
func constructorWithOpts(t1 *type1, t2 *type2, opts struct {
	dig.Optional

	l *zap.Logger
	s *tally.Scope
}) {
	// my opts struct has optional zap logger and tally scope
}
```

akshayjshah:
@glibsm I think I'm missing something in your example - since the final argument is an anonymous type, how would it be provided into the container?

To me, this approach encourages modules and libraries to depend on dig directly to solve a very common problem. I'd prefer to make all args optional and provide a separate validation helper.

glibsm:
@akshayjshah In the last example, we wouldn't provide the optional struct through the graph. We would *inject* the struct from the types available in dig, i.e. in this case it would look for a zap.Logger and tally.Scope.

> I'd prefer to make all args optional and provide a separate validation helper.

That approach is prototyped in another diff, this illustrates a different idea. 

The main con of that approach for is that it completely swaps the default for what it means to be a function parameter. Module functions are no longer regular Go, they have some very specific conventions and are callable with all-zero values if you use them incorrectly. In other words, I expect a lot of crashes as people try to call functions on nil parameters.

> To me, this approach encourages modules and libraries to depend on dig directly to solve a very common problem

In this sentence, what do you refer to as the "common problem".

akshayjshah:
Even after the update to embed the logger and tally scope, does that example actually work today? As I understand it, it would require dig to reflect over the *exported fields* of optional structs, using the graph to supply them. Does that code exist today?

Let's chat tomorrow about this. It's a really interesting approach, but I'm personally a bit uncomfortable with the tradeoffs.

breerly:
Super interesting pattern, pumped to talk about this and other approaches tomorrow.

glibsm:
Discussion tomorrow sounds good, but to answer your question: no, currently that pattern would not work without additional functionality of introspecting into anonymous structs that implement di.Object. Confusingly, it was an illustration of what *could be* to allow control of optionals on the constructor-level in addition to type-level

akshayjshah:
Ah, I see - I get it now 🤦‍♂️ 

Definitely interesting if we're also planning to support anonymous structs like this.

glibsm:
Logical continuation of this is in #71

PR #67, opened 2017-05-24
[Prototype] Graph-wide optional types

anuptalwalkar:
With this approach, dig container exposes a variadic API `dig.Optionals` to register any optional parameters. This API can be wrapped inside fx.App stack, so that fxstack.Optional() can be called by users, or any module that uses app during `Load`.

1. Any type can be `Optional`
2. services and modules get to define when to use Optionals
3. No existing API changes

In fx.App:
```go
//Optionals is used to present optional variables to dig
func (s *App) Optionals(i ...interface{}) error {
	s.container.Optionals(i...)
	return nil
}
```
In modules:
```go
// Load sets the multiple constructors needed to use this module
func Load(app *fx.App) {
	app.Optionals(SomeOptionalStuff...)
	app.Provide(NewSetup, NewStuff)
}
```

### DIFF
--- a/examples/optionals/optionals.go
+++ b/examples/optionals/optionals.go
@@ -1,3 +1,23 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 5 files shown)


glibsm:
Just relaying my in-person comment here.

My main concern with this approach is that parameters become optional/non-optional for the entire graph and that more often than not doesn't reflect reality. What can be optional for one constructor (for instance it knows how to not use metrics if they are not actually provided), the other constructor (for example Jaeger) will absolutely 100% require it. Yet because optionals are now container-scoped, you still need some sort of validation support on top.

breerly:
@anuptalwalkar is it possible to make this the default behavior?

For example, without increasing the API surface, make the following work:


```go
func main() {
  c := dig.New()

  // we're not providing SomeType, so it should be the 0-value
  // c.Provide(&&SomeType{})

  c.Invoke(Demo)
}

func Demo(someType *SomeType) {
   fmt.Println(someType) // THIS IS NIL!!!
}

type SomeType struct {}
```

If we had the above working in a branch, then we could try this API out for a few day to see how it feels.

glibsm:
@breerly for that behavior you can just use the zero value branch :)

breerly:
@glibsm really!? what branch specifically? I'll try it out now...

examples/optionals/optionals.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Type1 type
+type Type1 struct{}
+
+// Type2 type
+type Type2 struct{}
+
+func main() {
+	c := dig.New()
+	c.Optionals(&Type1{})
```
madhuravi: It becomes the provider’s responsibility to define optional. It should be the receiver (ie the module) that defines whether something is optional or not. For eg. same argument could be optional on one module and required on another.

anuptalwalkar: Its a feature! :) This functionality requires optional objects defined for the service, and not on individual methods. 
For example, would `zap.Logger` be optional only for a certain module or the entire service? The answer to this question relies on stack definitions, and not how module writers implement various modules.

anuptalwalkar:
Closing for greater good.

PR #68, opened 2017-05-25
[Prototype] Use zero values for unknown types

akshayjshah:
Just for discussion :taco:

This PR makes a simple change: instead of returning errors when `Read`/`Resolving` unknown types, it supplies the zero value of the type instead. This puts the burden of validating inputs on constructor authors.

### DIFF
--- a/container_test.go
+++ b/container_test.go
@@ -102,7 +102,7 @@ func TestResolve(t *testing.T) {
 				var p1 *Parent1
 				return c.Resolve(&p1)
 			},
-			"type *dig.Parent1 is not registered",
+			"",
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


glibsm:
How is this any different from #63 minus logging and prototyping what `vallie` library might look like?

We should consolidate all the discussion in one place, since this is the exact same approach

akshayjshah:
Sure; this approach is (to me) a simpler change, but you're right that the user-facing result is the same.

glibsm:
To make the discussion easier, I'll remove logging and push `Validate` into a new package purely for discussion sake. I don't know if any of the `[Prototype]` branches are ready to merge anyway, just used for illustration

PR #69, opened 2017-05-26
Fix deletion of cycled objects (was using array index)

glibsm:
Noticed this while working in another branch. We were trying to delete by the array index out of the nodes map, rather than the type.


internal/graph/graph.go
```diff
@@ -128,7 +128,7 @@ func (g *Graph) InsertConstructor(ctor interface{}) error {
 	// object needs to be part of the container to properly detect cycles
 	if cycleErr := g.recursiveDetectCycles(&n, nil); cycleErr != nil {
 		// if the cycle was detected delete from the container
-		for objType := range objTypes {
+		for _, objType := range objTypes {
```
alsamylkin: Tests?

glibsm:
this PR Is now obsolete

PR #70, opened 2017-05-26
[1/2] Add tests for user-facing behavior

akshayjshah:
# ☮️ 🙏 😇 🙇   I come in peace  :bow:😇 🙏 ☮️ 

TL;DR: This PR adds integration tests, which fail. Fixing the tests requires
enough refactoring that I separated it into in a follow-on PR.

As I was working on Fx modules that use dig, I noticed some behavior that
surprised me. In exploring dig's code, I realized that many of my assumptions
about how the library should behave aren't covered by dig's testing suite.

In this PR, I've added a bunch of end-to-end tests in the top-level dig package.
Those tests *only* use the public `Provide` and `Invoke` APIs to verify that the
behaviors Fx needs work correctly. I haven't altered any of dig's production
code.

Unfortunately, these tests exposed a number of bugs:

```
$ go test . | grep FAIL

    # Container.Provide(nil) panics.
--- FAIL: TestCantProvideUntypedNil (0.00s)

    # The container allows us to provide errors, but only via some code paths.
--- FAIL: TestCantProvideErrors (0.00s)

    # All types in the graph must have their deps satisfied, even if nobody's
    # using them.
--- FAIL: TestIncompleteGraphIsOkay (0.00s)

--- FAIL: TestEndToEndSuccess (0.00s)
    # We can't provide slices or maps from standard constructor functions, but
    # we can provide instances. Elaborate constructors like `func() []Foo, int,
    # error` also work correctly.
    --- FAIL: TestEndToEndSuccess/slice_constructor (0.00s)
    --- FAIL: TestEndToEndSuccess/map_constructor (0.00s)

    # We can't provide both []*Foo and *Foo, since dig considers them the same
    # type.
    --- FAIL: TestEndToEndSuccess/collections_and_instances_of_same_type (0.00s)
```

This is concerning to me, since dig has 89% code coverage. Fixing these bugs and
structuring the code so that we can confidently refactor without breaking
user-facing behavior was a larger change, so I've broken it out into a separate
PR.

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -340,13 +340,20 @@ func TestInvokesUseCachedObjects(t *testing.T) {
 	t.Parallel()
 
 	c := New()
-	calls := 0
-	buf := &bytes.Buffer{}
-	require.NoError(t, c.Provide(func() *bytes.Buffer { return buf }))
 
+	constructorCalls := 0
+	buf := &bytes.Buffer{}
+	require.NoError(t, c.Provide(func() *bytes.Buffer {
+		assert.Equal(t, 0, constructorCalls, "constructor must not have been called before")
+		constructorCalls++
+		return buf
+	}))
+
+	calls := 0
 	for i := 0; i < 3; i++ {
 		assert.NoError(t, c.Invoke(func(b *bytes.Buffer) {
 			calls++
+			require.Equal(t, 1, constructorCalls, "constructor must be called exactly once")
 			require.Equal(t, buf, b, "invoke got different buffer pointer")
 		}), "invoke %d failed", i)
 		require.Equal(t, i+1, calls, "invoked function not called")
--- a/dig_test.go
+++ b/dig_test.go
@@ -265,7 +265,7 @@ func TestProvideRespectsConstructorErrors(t *testing.T) {
 }
 
 func TestCantProvideUntypedNil(t *testing.T) {
-	t.Fatalf("Calling Container.Provide(nil) crashes the main testing thread. Failing early.")
+	t.Skip("Calling Container.Provide(nil) crashes the main testing thread")
 	t.Parallel()
 	c := New()
 	assert.Error(t, c.Provide(nil))
--- a/dig_test.go
+++ b/dig_test.go
@@ -285,6 +285,34 @@ func TestCantProvideErrors(t *testing.T) {
 	assert.NoError(t, c.Provide(errors.New("foo")))
 }
 
+type someError struct{}
+
+var _ error = (*someError)(nil)
+
+func (*someError) Error() string { return "foo" }
+
+func TestCanProvideErrorLikeType(t *testing.T) {
+	t.Parallel()
+
+	tests := []interface{}{
+		func() *someError { return &someError{} },
+		func() (*someError, error) { return &someError{}, nil },
+		&someError{},
+	}
+
+	for _, tt := range tests {
+		t.Run(fmt.Sprintf("%T", tt), func(t *testing.T) {
+			c := New()
+			require.NoError(t, c.Provide(tt), "provide must not fail")
+
+			require.NoError(t, c.Invoke(
+				func(err *someError) {
+					assert.NotNil(t, err, "invoke received nil")
+				}), "invoke must not fail")
+		})
+	}
+}
+
 func TestProvideKnownTypesFails(t *testing.T) {
 	t.Parallel()
 
--- a/dig_test.go
+++ b/dig_test.go
@@ -35,9 +35,11 @@ func TestEndToEndSuccess(t *testing.T) {
 
 	t.Run("pointer", func(t *testing.T) {
 		c := New()
-		require.NoError(t, c.Provide(&bytes.Buffer{}), "provide failed")
-		require.NoError(t, c.Invoke(func(b *bytes.Buffer) {
-			require.NotNil(t, b, "invoke got nil buffer")
+		b := &bytes.Buffer{}
+		require.NoError(t, c.Provide(b), "provide failed")
+		require.NoError(t, c.Invoke(func(got *bytes.Buffer) {
+			require.NotNil(t, got, "invoke got nil buffer")
+			require.True(t, got == b, "invoke got wrong buffer")
 		}), "invoke failed")
 	})
 
@@ -54,11 +56,14 @@ func TestEndToEndSuccess(t *testing.T) {
 
 	t.Run("pointer constructor", func(t *testing.T) {
 		c := New()
+		var b *bytes.Buffer
 		require.NoError(t, c.Provide(func() *bytes.Buffer {
-			return &bytes.Buffer{}
+			b = &bytes.Buffer{}
+			return b
 		}), "provide failed")
-		require.NoError(t, c.Invoke(func(b *bytes.Buffer) {
-			require.NotNil(t, b, "invoke got nil buffer")
+		require.NoError(t, c.Invoke(func(got *bytes.Buffer) {
+			require.NotNil(t, got, "invoke got nil buffer")
+			require.True(t, got == b, "invoke got wrong buffer")
 		}), "invoke failed")
 	})
 
--- a/dig_test.go
+++ b/dig_test.go
@@ -330,7 +330,6 @@ func TestProvideKnownTypesFails(t *testing.T) {
 			for _, second := range provideArgs {
 				assert.Error(t, c.Provide(second), "second provide must fail")
 			}
-
 		})
 	}
 }
--- a/dig_test.go
+++ b/dig_test.go
@@ -23,6 +23,7 @@ package dig
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"io"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+	c := New()
+	assert.Error(t, c.Provide(nil))
+}
+
+func TestCantProvideErrors(t *testing.T) {
+	t.Parallel()
+	c := New()
+
+	assert.Error(t, c.Provide(func() error { return errors.New("foo") }))
+	// TODO: This is another case where we're losing type information, which
```
alsamylkin: What about a regular type comparison, that we are using:
https://github.com/uber-go/dig/blob/master/container.go#L71
Is it a bug then? https://play.golang.org/p/N0c4JiwR4W

What makes me to think about adding a test, that has a struct, that implements the error interface..

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+func TestProvideCycleFails(t *testing.T) {
+	t.Parallel()
+
+	// A <- B <- C
+	// |         ^
+	// |_________|
+	type A struct{}
+	type B struct{}
+	type C struct{}
+	newA := func(C) A { return A{} }
```
alsamylkin: Looks like the test fails because returned types are not pointers..

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+		require.NoError(t, c.Invoke(func(bs []*bytes.Buffer) {
+			require.Equal(t, 2, len(bs), "invoke got unexpected number of buffers")
+		}), "invoke failed")
+	})
+
+	t.Run("array", func(t *testing.T) {
+		c := New()
+		bufs := [1]*bytes.Buffer{{}}
+		require.NoError(t, c.Provide(bufs), "provide failed")
+		require.NoError(t, c.Invoke(func(bs [1]*bytes.Buffer) {
```
akshayjshah: IMO, we want to work with whatever is provided. It's up to the user to decide which they want.

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+		}
+		consumer := func(b *bytes.Buffer, nums []int) {
+			assert.NotNil(t, b, "invoke got nil buffer")
+			assert.Equal(t, 1, len(nums), "invoke got empty slice")
+		}
+		require.NoError(t, c.Provide(constructor), "provide failed")
+		require.NoError(t, c.Invoke(consumer), "invoke failed")
+	})
+
+	t.Run("collections and instances of same type", func(t *testing.T) {
```
akshayjshah: The test points out that dig's pointer and map support is currently a little broken - it's not possible to provide both `[]*Foo` and `*Foo`. Obviously, we *should* support providing both.

dig_test.go
akshayjshah: The type comparison works fine for constructors that explicitly return `error`. It doesn't work if the constructor returns a type that implements the error interface. It definitely doesn't work for instances, since we lose interface information by passing instances to `Provide` as `interface{}` (see https://research.swtch.com/interfaces for details). We *could* drop all types that implement `error`, but that's a larger change.

The intention of these two PRs isn't to fix all bugs in dig, but to get to a place where everyone knows what dig's behavior actually is.

dig_test.go
akshayjshah: Good point! Changed.

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+	"errors"
+	"io"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEndToEndSuccess(t *testing.T) {
+	t.Parallel()
```
prashantv: Do these tests take long enough to make `Parallel` worth it? It makes debugging failures a little trickier so I wouldn't add it unless we're seeing a significant win.

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+	t.Run("pointer", func(t *testing.T) {
+		c := New()
+		require.NoError(t, c.Provide(&bytes.Buffer{}), "provide failed")
+		require.NoError(t, c.Invoke(func(b *bytes.Buffer) {
+			require.NotNil(t, b, "invoke got nil buffer")
```
prashantv: since you're providing a specific value, can you assert that the value that got passed in here matches?

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+		}), "invoke failed")
+	})
+
+	t.Run("pointer constructor", func(t *testing.T) {
+		c := New()
+		require.NoError(t, c.Provide(func() *bytes.Buffer {
+			return &bytes.Buffer{}
+		}), "provide failed")
+		require.NoError(t, c.Invoke(func(b *bytes.Buffer) {
+			require.NotNil(t, b, "invoke got nil buffer")
```
prashantv: can we assert that `Invoke` gets passed in the same value the constructor returned

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+		}), "invoke failed")
+	})
+
+	t.Run("struct", func(t *testing.T) {
+		t.Skip("Not yet supported.")
+		c := New()
+		buf := &bytes.Buffer{}
```
prashantv: why create a pointer only to dereference it below? you can just do `var buf bytes.Buffer` or `buf := bytes.Buffer{}`

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+		require.NoError(t, c.Provide(*buf), "provide failed")
+		require.NoError(t, c.Invoke(func(b bytes.Buffer) {
+			// ensure we're getting back the buffer we put in
+			require.Equal(t, "foo", buf.String(), "invoke got new buffer")
+		}), "invoke failed")
+	})
+
+	t.Run("slice", func(t *testing.T) {
+		c := New()
+		bufs := []*bytes.Buffer{{}, {}}
```
prashantv: if we're testing a slice, can we keep it simpler and do something like `[]int{1,2}` -- or did you want to test slice of struct pointers specifically?

using a simple slice and verifying the contents of the slice seems simpler to understand and provides more validation.

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+		var called bool
+		err := c.Invoke(func(b *bytes.Buffer) { called = true })
+		assert.Contains(t, err.Error(), "oh no", "expected to bubble up constructor error")
+		assert.False(t, called, "shouldn't call invoked function when deps aren't available")
+	})
+}
+
+func TestCantProvideUntypedNil(t *testing.T) {
+	t.Fatalf("Calling Container.Provide(nil) crashes the main testing thread. Failing early.")
```
prashantv: why do some tests use `Skip` while some use Fatal? Seems like this should probably be skip as well?

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+	t.Parallel()
+	c := New()
+	assert.Error(t, c.Provide(nil))
+}
+
+func TestCantProvideErrors(t *testing.T) {
+	t.Parallel()
+	c := New()
+
+	assert.Error(t, c.Provide(func() error { return errors.New("foo") }))
```
prashantv: Just curious about this behaviour -- what happens if the func returns a type that is not `error`, but the type returned implements `error`?

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+	// provide types that *implement* error, but let's discuss a holistic
+	// solution.
+	assert.NoError(t, c.Provide(errors.New("foo")))
+}
+
+func TestProvideKnownTypesFails(t *testing.T) {
+	t.Parallel()
+	c := New()
+	assert.NoError(t, c.Provide(func() *bytes.Buffer { return nil }))
+	assert.Error(t, c.Provide(func() *bytes.Buffer { return nil }))
```
prashantv: can you verify that it fails even if the value is provided. this might need to be a loop test like:
```
provideArgs := []interface{}{
  func() *bytes.Buffer { return nil },
  func() (*bytes.Buffer, error) { return nil, nil },
  &bytes.Buffer{},
}

for _, firstProvide := range provideArgs {
  assert.NoError(t, c.Provide(firstProvide), ...)
  for _, secondProvide := range provideArgs {
    assert.Error(t, c.Provide(secondProvide), ...)
  }
}
```

that way we ensure that once a type is known (regardless of how it was provided), other provides fail.

dig_test.go
```diff
@@ -0,0 +1,360 @@
...
+	assert.Error(t, c.Provide(func(*bytes.Buffer) {}))
+}
+
+func TestInvokesUseCachedObjects(t *testing.T) {
+	t.Parallel()
+
+	c := New()
+	calls := 0
+	buf := &bytes.Buffer{}
+	require.NoError(t, c.Provide(func() *bytes.Buffer { return buf }))
```
prashantv: as an extra safety, can we have the provided `func` increment a counter and ensure that the value is 0 initially, and then 1 the loop

dig_test.go
akshayjshah: Not really, but all the existing tests were marked parallel. In the interest of minimizing change, I stuck with that convention.

dig_test.go
akshayjshah: Can do.

dig_test.go
akshayjshah: Yep, good point.

dig_test.go
akshayjshah: I'd prefer not to, since there's an ongoing discussion about whether `dig` should allow primitive values at all. I'd prefer to stick to some more-realistic type here, so that future changes don't also need to alter these high-level types.

This test should also check that invoke receives the same pointer that provide returned, though.

dig_test.go
abhinav: Since the check is against the exact error type, that'll just work. Added a test for it.

dig_test.go
```diff
@@ -0,0 +1,424 @@
...
+
+	for _, first := range provideArgs {
+		t.Run(fmt.Sprintf("%T", first), func(t *testing.T) {
+			c := New()
+			require.NoError(t, c.Provide(first), "first provide must not fail")
+
+			for _, second := range provideArgs {
+				assert.Error(t, c.Provide(second), "second provide must fail")
+			}
+
```
prashantv: super nit: remove blank before `}`

abhinav: true

PR #71, opened 2017-05-26
Support Parameter Objects

abhinav:
This adds support for parameter objects to Dig. Parameter objects are arguments
to a constructor which contain zero or more fields that are the actual
dependencies for the constructor. Users opt-in for a struct to be interpreted
as a parameter object by embedding the `dig.Param` type into it.

    type Config struct {
        dig.Param

        In io.Reader
        Out io.Writer
    }

    func NewTransport(c Config) (*Transport, error) {
        // ..
    }

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -238,7 +238,7 @@ func (c *Container) constructorArgs(ctype reflect.Type) ([]reflect.Value, error)
 
 		t := ctype.In(i)
 		if t.Implements(_parameterObjectType) {
-			arg, err = getParameterObject(c, t)
+			arg, err = c.getParameterObject(t)
 		} else {
 			arg, err = c.get(t)
 		}
@@ -358,7 +358,7 @@ func getParameterDependencies(t reflect.Type) ([]reflect.Type, error) {
 	return deps, nil
 }
 
-func getParameterObject(c *Container, t reflect.Type) (reflect.Value, error) {
+func (c *Container) getParameterObject(t reflect.Type) (reflect.Value, error) {
 	dest := reflect.New(t).Elem()
 	result := dest
 	for t.Kind() == reflect.Ptr {
--- a/dig.go
+++ b/dig.go
@@ -28,8 +28,10 @@ import (
 )
 
 var (
-	_noValue reflect.Value
-	_errType = reflect.TypeOf((*error)(nil)).Elem()
+	_noValue             reflect.Value
+	_errType             = reflect.TypeOf((*error)(nil)).Elem()
+	_parameterObjectType = reflect.TypeOf((*parameterObject)(nil)).Elem()
+	_paramType           = reflect.TypeOf(Param{})
 )
 
 // A Container is a directed, acyclic graph of dependencies. Dependencies are
@@ -113,6 +115,9 @@ func (c *Container) provideInstance(val interface{}) error {
 	if vtype == _errType {
 		return errors.New("can't provide errors")
 	}
+	if vtype.Implements(_parameterObjectType) {
+		return errors.New("can't provide parameter objects")
+	}
 	if _, ok := c.nodes[vtype]; ok {
 		return errors.New("already in container")
 	}
@@ -129,6 +134,9 @@ func (c *Container) provideConstructor(ctor interface{}, ctype reflect.Type) err
 			// Don't register errors into the container.
 			continue
 		}
+		if rt.Implements(_parameterObjectType) {
+			return errors.New("can't provide parameter objects")
+		}
 		if _, ok := returnTypes[rt]; ok {
 			return fmt.Errorf("returns multiple %v", rt)
 		}
@@ -143,7 +151,10 @@ func (c *Container) provideConstructor(ctor interface{}, ctype reflect.Type) err
 
 	nodes := make([]node, 0, len(returnTypes))
 	for rt := range returnTypes {
-		n := newNode(rt, ctor, ctype)
+		n, err := newNode(rt, ctor, ctype)
+		if err != nil {
+			return err
+		}
 		nodes = append(nodes, n)
 		c.nodes[rt] = n
 	}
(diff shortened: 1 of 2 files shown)


dig.go
```diff
@@ -220,7 +231,18 @@ func (c *Container) remove(nodes []node) {
...
 	for i := 0; i < ctype.NumIn(); i++ {
-		arg, err := c.get(ctype.In(i))
+		var (
+			arg reflect.Value
+			err error
+		)
+
+		t := ctype.In(i)
+		if t.Implements(_parameterObjectType) {
+			arg, err = getParameterObject(c, t)
```
alsamylkin: Why `getParameterObject` is a function, not a method?

dig.go
```diff
@@ -275,3 +307,85 @@ func detectCycles(n node, graph map[reflect.Type]node, path []reflect.Type, seen
...
+		if f.PkgPath != "" {
+			continue // skip private fields
+		}
+
+		// Skip the embedded Param type.
+		if f.Anonymous && f.Type == _paramType {
+			continue
+		}
+
+		// THe user added a parameter object as a dependency. We don't recurse
```
alsamylkin: nit: The

dig.go
```diff
@@ -275,3 +307,85 @@ func detectCycles(n node, graph map[reflect.Type]node, path []reflect.Type, seen
...
+		// /yet/ so let's try to give an informative error message.
+		if f.Type.Implements(_parameterObjectType) {
```
alsamylkin: I've tried private implementations and it looks very confusing:
foo/foo.go:
```golang
package foo

type A struct {}

func (A) private() {
	println("foo")
}
```
main.go
```golang
package main

import (
	"fmt"
	"tmp/foo"
	"reflect"
)

type B struct{foo.A}
type private interface {
	private()
}

func main() {
	b := reflect.ValueOf(B{})
	v := (*private)(nil)
	fmt.Println(reflect.TypeOf(v).Elem())
	fmt.Println(b.Type().Implements(reflect.TypeOf(v).Elem()))
}
```
the output is:
```
main.private
true
```
May be I am doing something wrong, but it seems there could be conflicts if a type has parameterObject() method..

dig.go
```diff
@@ -275,3 +307,85 @@ func detectCycles(n node, graph map[reflect.Type]node, path []reflect.Type, seen
...
+			continue
+		}
+
+		v, err := c.get(f.Type)
+		if err != nil {
+			return result, fmt.Errorf(
+				"could not get field %v (type %v) of %v: %v", f.Name, f.Type, t, err)
+		}
+
+		dest.Field(i).Set(v)
```
alsamylkin: Do we need to check if the field is settable? ([In case dig will support non-pointer types](https://play.golang.org/p/tHhOy93wDu))

abhinav: The field will always be settable because the struct that contains it is addressable (because we created it with `reflect.New`).

dig.go
abhinav: No reason. Changed.

dig.go
abhinav: Interfaces with unexported methods may only be implemented by types defined in
that package. A type that implements a private interface may be embedded
inside another type (which may reside outside the package containing the
private interface). Because methods of embedded types are "inherited", the
type that embeds the other type now implements the private interface too.

That's exactly the behavior we want: Only types that embed `dig.Param`
implement `parameterObject`. Demonstration:

```go
// foo/foo.go
package foo

type A interface { private() }

func Matches(x interface{}) bool {
	_, ok := x.(A)
	return ok
}

// External types embedding this will match.
type EmbedMe struct{}

func (EmbedMe) private() {}
```

```go
// main.go
package main

import (
	"fmt"
	"tmp/foo"
)

type bad struct{}

func (bad) private() {}

type good struct{ foo.EmbedMe }

func main() {
	fmt.Println("bad:", foo.Matches(bad{}))
	fmt.Println("good:", foo.Matches(good{}))
	// Output:
	// bad: false
	// good: true
}
```

In your example, the other package doesn't define the interface, only the
method. And you embed it in a type inside the package where you define the
interface, so it somehow matches.

This will not work for types that simply implement `parameterObject()`.

alsamylkin: I got the trick you are doing, what I am arguing is that `Implements`(which you are using here) doesn't follow the same behavior as type assertion. If you replace Matches function with it:
```golang
func Matches(x interface{}) bool {
	return reflect.TypeOf(x).Implements(reflect.TypeOf((*A)(nil)).Elem())
}
```
Output is going to be:
```
bad: true
good: true
```

dig.go
```diff
@@ -113,6 +115,9 @@ func (c *Container) provideInstance(val interface{}) error {
 	if vtype == _errType {
 		return errors.New("can't provide errors")
 	}
+	if vtype.Implements(_parameterObjectType) {
+		return errors.New("can't provide parameter objects")
```
anuptalwalkar: parameter term is super generic. Can this error be more explicit? `can't provide parameterized object`

dig.go
```diff
@@ -220,7 +231,18 @@ func (c *Container) remove(nodes []node) {
...
 	for i := 0; i < ctype.NumIn(); i++ {
-		arg, err := c.get(ctype.In(i))
+		var (
+			arg reflect.Value
+			err error
+		)
+
+		t := ctype.In(i)
+		if t.Implements(_parameterObjectType) {
+			arg, err = c.getParameterObject(t)
```
anuptalwalkar: nit: getting parameterObject can be part of `c.get(t)`. Then you don't need to declare variables and assign. implementation wise we just want an object back from container.

dig.go
```diff
@@ -275,3 +307,85 @@ func detectCycles(n node, graph map[reflect.Type]node, path []reflect.Type, seen
...
+// Returns dependencies introduced by a parameter object.
+func getParameterDependencies(t reflect.Type) ([]reflect.Type, error) {
+	for t.Kind() == reflect.Ptr {
+		t = t.Elem()
+	}
+
+	var deps []reflect.Type
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		if f.PkgPath != "" {
```
anuptalwalkar: nit: refactor? pkgPath and Anonymous is been checked in multiple places.

dig.go
abhinav: Oh, wow. That's a bug in reflect. Good catch!

abhinav: @alsamylkin Created https://github.com/golang/go/issues/20541. Thanks!

dig.go
abhinav: I've been using the term "parameter object" for types that embed `dig.Param` in all the documentation and errors so far. That's the standard name for objects which represent all the parameter of a function, created solely to avoid having too many arguments on the function. See http://wiki.c2.com/?ParameterObject, https://refactoring.com/catalog/introduceParameterObject.html.

We can call this something else, maybe give it a dig-specific name, but "parameterized object" has a whole different meaning.

dig.go
abhinav: Good call. Updating the other PR with this.

dig.go
abhinav: Unfortunately that's unavoidable. These are two different entry points. This check is necessary to make sure we don't muck with private fields.

PR #73, opened 2017-05-26
dig.Param: Recurse into nested parameter objects

abhinav:
This change adds support to parameter objects for recursing into fields that
are also parameter objects. That is, this change allows the following,

    type OutputConfig struct {
        dig.Param

        Writer io.Writer
    }

    type Config struct {
        dig.Param

        Input io.Reader
        Output OutputConfig
    }

    func NewTransport(cfg Config) (*Transport, error) { .. }

Besides the usefulness of treating fields similar to constructor parameters,
there's another motivating use case for this: Aliasing `dig.Param`. Without
this change, we have special handling for the `dig.Param` type, opting it out
of resolution from `dig`. If a library author wants to use their own type in
place of `dig.Param` (for example, `fx.Param`), they can't quite do this
because any alias of `dig.Param` is a different type; one which the library
does not special-case.

    package fx

    type Param dig.Param

    type Config struct {
        Param

        Input io.Reader
    }

    // The above with fail because dig doesn't know how to resolve fx.Param.

With this change, Fx will be able to support the following.

    package fx

    type Param struct{ dig.Param }

    type Config struct {
        Param

        // ..
    }

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -181,7 +181,7 @@ func (c *Container) get(t reflect.Type) (reflect.Value, error) {
 	}
 
 	if t.Implements(_parameterObjectType) {
-		// No caching
+		// We do not want parameter objects to be cached.
 		return c.createParamObject(t)
 	}
 
--- a/dig.go
+++ b/dig.go
@@ -257,16 +257,7 @@ type node struct {
 func newNode(provides reflect.Type, ctor interface{}, ctype reflect.Type) (node, error) {
 	deps := make([]reflect.Type, 0, ctype.NumIn())
 	for i := 0; i < ctype.NumIn(); i++ {
-		t := ctype.In(i)
-		if t.Implements(_parameterObjectType) {
-			pdeps, err := getParameterDependencies(t)
-			if err != nil {
-				return node{}, err
-			}
-			deps = append(deps, pdeps...)
-		} else {
-			deps = append(deps, t)
-		}
+		deps = append(deps, getCtorParamDependencies(ctype.In(i))...)
 	}
 
 	return node{
@@ -275,6 +266,26 @@ func newNode(provides reflect.Type, ctor interface{}, ctype reflect.Type) (node,
 		ctype:    ctype,
 		deps:     deps,
 	}, nil
+}
+
+// Retrives the dependencies for the parameter of a constructor.
+func getCtorParamDependencies(t reflect.Type) (deps []reflect.Type) {
+	if !t.Implements(_parameterObjectType) {
+		deps = append(deps, t)
+		return
+	}
+
+	deps = make([]reflect.Type, 0, t.NumField())
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		if f.PkgPath != "" {
+			continue // skip private fields
+		}
+
+		deps = append(deps, getCtorParamDependencies(f.Type)...)
+	}
+
+	return
 }
 
 func cycleError(cycle []reflect.Type, last reflect.Type) error {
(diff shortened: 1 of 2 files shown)


abhinav:
TODO: getParameterObject can probably be absorbed into container.get

glibsm:
what's going on with the coverage?

dig.go
```diff
@@ -173,11 +172,17 @@ func (c *Container) isAcyclic(n node) error {
...
 }
 
+// Retrieve a type from the container
 func (c *Container) get(t reflect.Type) (reflect.Value, error) {
 	if v, ok := c.cache[t]; ok {
 		return v, nil
 	}
 
+	if t.Implements(_parameterObjectType) {
+		// No caching
```
glibsm: intentional, or just a TODO for now?

abhinav: Intentional. I'll clarify.

abhinav:
Coverage recovered. The remaining drop is because of the placeholder function `parameterObject()` which is not meant to ever be called.

image

The threshold may need to be adjusted.

abhinav:
derp. did not mean to merge #81 here

dig.go
```diff
@@ -283,6 +268,26 @@ func newNode(provides reflect.Type, ctor interface{}, ctype reflect.Type) (node,
 	}, nil
 }
 
+// Retrives the dependencies for the parameter of a constructor.
+func getCtorParamDependencies(t reflect.Type) (deps []reflect.Type) {
```
glibsm: Ctor -> Constructor
Param -> ""

dig.go
```diff
@@ -283,6 +268,26 @@ func newNode(provides reflect.Type, ctor interface{}, ctype reflect.Type) (node,
...
+	if !t.Implements(_parameterObjectType) {
+		deps = append(deps, t)
+		return
+	}
```
akshayjshah: Are named returns providing much value here? To me, this is less readable than `return []reflect.Type{t}`.

dig_test.go
```diff
@@ -286,6 +286,53 @@ func TestEndToEndSuccess(t *testing.T) {
...
+		}), "provide must not fail")
+
+		require.NoError(t, c.Invoke(func(p *someParam) {
+			require.NotNil(t, p, "someParam must not be nil")
+			require.NotNil(t, p.Buffer, "someParam must not be nil")
+			require.NotNil(t, p.Another, "someParam must not be nil")
+			require.NotNil(t, p.Another.Buffer, "someParam must not be nil")
+			require.True(t, p.Buffer == p.Another.Buffer, "buffer must be the same")
+		}), "invoke must not fail")
+	})
```
akshayjshah: Can we also validate that (1) we're getting the same buffer reference that the constructor returned, and (2) we're only calling the buffer constructor once? Same for test above.

dig_test.go
```diff
@@ -286,6 +286,53 @@ func TestEndToEndSuccess(t *testing.T) {
 		}))
 	})
 
+	t.Run("param wrapper", func(t *testing.T) {
+		c := New()
+		require.NoError(t, c.Provide(func() *bytes.Buffer {
+			return new(bytes.Buffer)
+		}), "provide failed")
+
+		type MyParam struct{ Param }
```
akshayjshah: Why not `type MyParam Param`?

dig_test.go
```diff
@@ -286,6 +286,53 @@ func TestEndToEndSuccess(t *testing.T) {
...
+		c := New()
+		require.NoError(t, c.Provide(func() *bytes.Buffer {
+			return new(bytes.Buffer)
+		}), "provide must not fail")
+
+		require.NoError(t, c.Invoke(func(p *someParam) {
+			require.NotNil(t, p, "someParam must not be nil")
+			require.NotNil(t, p.Buffer, "someParam must not be nil")
+			require.NotNil(t, p.Another, "someParam must not be nil")
+			require.NotNil(t, p.Another.Buffer, "someParam must not be nil")
```
akshayjshah: Leftover copypasta? Failure messages should be different, I think.

abhinav: my bad

PR #75, opened 2017-05-30
Add test for multi-return constructor being called once

glibsm:
It passes as-is, but covers an important case.

There is a similar test for counting a single-return constructor.
Just want to make sure we don't accidentally break this while migrating
the map to a composite key in the future

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -212,6 +212,27 @@ func TestEndToEndSuccess(t *testing.T) {
 		require.NoError(t, c.Invoke(consumer), "invoke failed")
 	})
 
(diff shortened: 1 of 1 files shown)


dig_test.go
```diff
@@ -212,6 +212,27 @@ func TestEndToEndSuccess(t *testing.T) {
 		require.NoError(t, c.Invoke(consumer), "invoke failed")
 	})
 
+	t.Run("multiple-type constructor is called once", func(t *testing.T) {
```
anuptalwalkar: Can you add test for `func() (*A, *A, error)` as well?

glibsm: Sure! Good idea

PR #76, opened 2017-05-31
Add edge tests around Provide and Invoke

anuptalwalkar:
Added some missing edge cases around provide - 
1. Multiple return provide returning same object more than once
2. Failure cases for combination of Provide ctor and instance of same type
3. Invocation inside invocation.

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -247,14 +247,6 @@ func TestEndToEndSuccess(t *testing.T) {
 		require.NoError(t, c.Invoke(func(a *A, b *B) {}), "AB invoke failed")
 		require.Equal(t, 1, count, "Constructor must be called once")
 	})
-	t.Run("multiple-type constructor returns multiple objects of same type", func(t *testing.T) {
-		c := New()
-		type A struct{}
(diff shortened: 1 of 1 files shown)


abhinav:
@anuptalwalkar can you add something to the PR description for more context?

dig_test.go
```diff
@@ -233,7 +233,35 @@ func TestEndToEndSuccess(t *testing.T) {
 		require.NoError(t, c.Invoke(func(a *A, b *B) {}), "AB invoke failed")
 		require.Equal(t, 1, count, "Constructor must be called once")
 	})
+	t.Run("multiple-type constructor returns multiple objects of same type", func(t *testing.T) {
+		c := New()
+		type A struct{}
+		constructor := func() (*A, *A, error) {
+			return &A{}, &A{}, nil
+		}
+		require.Error(t, c.Provide(constructor), "provide failed")
```
abhinav: I think this particular test function has only success cases (EndToEndSuccess). Perhaps move this out to a different test.

anuptalwalkar: good point. I will move it out.

PR #77, opened 2017-05-31
Add support for optional parameters

glibsm:
The refactor and Params made it quite trivial to add support and e2e
test for optional tags.

Reading the test explains the expected behavior quite well.

Stacked on #70, #72, #71

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -380,8 +380,15 @@ func (c *Container) getParameterObject(t reflect.Type) (reflect.Value, error) {
 
 		v, err := c.get(f.Type)
 		if err != nil {
-			return result, fmt.Errorf(
-				"could not get field %v (type %v) of %v: %v", f.Name, f.Type, t, err)
+			// TODO: once other tags are avilabled (named=foo),
+			// probably should serialize the tags better, or at least scan the string
+			tag := f.Tag.Get("dig")
+			if tag == "optional" {
+				v = reflect.Zero(f.Type)
+			} else {
+				return result, fmt.Errorf(
+					"could not get field %v (type %v) of %v: %v", f.Name, f.Type, t, err)
+			}
 		}
 
 		dest.Field(i).Set(v)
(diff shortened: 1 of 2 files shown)


dig_test.go
```diff
@@ -294,6 +294,34 @@ func TestEndToEndSuccess(t *testing.T) {
...
+			return &type1{}, &type3{}, &type4{}
+		}
+
+		c := New()
+		type param struct {
+			Param
+
+			T1 *type1 // regular 'ol type
+			T2 *type2 `dig:"optional,useless_tag"`   // optional type NOT in the graph
+			T3 *type3 `unrelated:"foo=42, optional"` // type in the graph with unrelated tag
```
abhinav: can we add one with the dig tag and a space after the comma?

dig.go
```diff
@@ -380,12 +381,27 @@ func getParameterObject(c *Container, t reflect.Type) (reflect.Value, error) {
...
 		dest.Field(i).Set(v)
 	}
 
 	return result, nil
 }
+
+func containsString(source []string, s string) bool {
+	res := false
+	for _, v := range source {
+		if v == s {
```
abhinav: `strings.TrimSpace(v)`?

dig.go
```diff
@@ -380,10 +381,8 @@ func getParameterObject(c *Container, t reflect.Type) (reflect.Value, error) {
 
 		v, err := c.get(f.Type)
 		if err != nil {
-			// TODO: once other tags are avilabled (named=foo),
-			// probably should serialize the tags better, or at least scan the string
-			tag := f.Tag.Get("dig")
-			if tag == "optional" {
+			tags := strings.Split(f.Tag.Get("dig"), ",")
+			if containsString(tags, "optional") {
```
anuptalwalkar: "dig:optional" should be case agnostic?

abhinav: Actually, I don't think we should be case insensitive. Standard struct tags are case sensitive.

https://play.golang.org/p/KejIp28lR0

anuptalwalkar: ah makes sense. I was thinking `"dig:optional"` or `"dig:Optional"` should work either way.

PR #78, opened 2017-06-01
Non invasive optional params

alsamylkin:
We can let users to introduce their own optional parameters, that are not necessarily present in the container by splitting a user provided function that has `n` arguments into `n` functions that take 1 argument. This lets us to introduce optional parameters to dig without changing the Invoke logic.

Note: this change is just to introduce a proof of concept and doesn't have to be in DIG, same with tests, if the direction is good I'll polish everything.


param.go
```diff
@@ -0,0 +1,72 @@
+package dig
+
+import (
+	"errors"
+	"reflect"
+)
+
+// Param represents an argument to invoked function.
+type Param struct {
```
alsamylkin: I guess it can be named `Optional` instead, but we can extend the behaviour(e.g. with validation, etc) and rely only on the field.

anuptalwalkar: Interesting. But now caller will have to infer and pass in `reflect.Type`, `reflect.Value` as well. How about `Param` just be a variadic interface and we do the heavy loading of finding Type and Value?

glibsm:
I'm a little confused here. How does this let *modules* control their optional parameters?

anuptalwalkar:
Yeah, these can't be on a method. The way I read it, I was assuming a call  `module.OptionalParams() []interface{}` so that modules can setup Optional parameters somewhere in the code.

param.go
alsamylkin: Well, they are passing a reflect.Type and actual default value. I can try to infer type from the value, the only thing is that users have to have typed nils :(

alsamylkin:
Yeah, I misunderstood the problem, let me check if we can specify optionals in the function signature itself without tags.

alsamylkin:
Looks like we are diving in the tag world.

PR #79, opened 2017-06-01
optional tag: Support "yes"

abhinav:
We may want a more sophisticated truthiness check in the future.

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -655,6 +655,27 @@ func TestInvokeFailures(t *testing.T) {
 		assert.Error(t, c.Invoke(func(*bytes.Buffer) {}))
 	})
 
+	t.Run("unmet required dependency", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


dig_test.go
```diff
@@ -368,13 +369,15 @@ func TestEndToEndSuccess(t *testing.T) {
 			T2 *type2 `optional:"true" useless_tag:"false"` // optional type NOT in the graph
 			T3 *type3 `unrelated:"foo=42, optional"`        // type in the graph with unrelated tag
 			T4 *type4 `optional:"true"`                     // optional type present in the graph
+			T5 *type5 `optional:"yes"`                      // optional type NOT in the graph with "yes"
```
glibsm: we should probably have an `optional:"no"` case, too.

glibsm:
ah, there is the missing coverage!

PR #81, opened 2017-06-01
param: Pointers are not parameter objects

abhinav:
This narrows down the definition of parameter objects to only structs
embedding `dig.Param`. Pointers to structs embedding `dig.Param` are
treated like any other struct pointer.

---

This is a PR against #73

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -116,7 +116,7 @@ func (c *Container) provideInstance(val interface{}) error {
 	if vtype == _errType {
 		return errors.New("can't provide errors")
 	}
-	if vtype.Implements(_parameterObjectType) {
+	if isParameterObject(vtype) {
 		return errors.New("can't provide parameter objects")
 	}
 	if _, ok := c.nodes[vtype]; ok {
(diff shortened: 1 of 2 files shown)


dig_test.go
```diff
@@ -539,15 +524,19 @@ func TestCantProvideParameterObjects(t *testing.T) {
...
-		})
-		require.Error(t, err, "provide should fail")
-		require.Contains(t, err.Error(), "can't provide parameter objects")
+		require.NoError(t, c.Provide(func() (*Args, error) {
+			return args, nil
+		}), "provide failed")
+		require.NoError(t, c.Invoke(
+			func(a *Args) {
+				require.True(t, args == a, "args must match")
+			}), "invoke failed")
```
glibsm: what's going on with whitespace here

dig.go
```diff
@@ -332,17 +332,14 @@ type parameterObject interface {
...
 // populated from the dig container.
 func (c *Container) createParamObject(t reflect.Type) (reflect.Value, error) {
 	dest := reflect.New(t).Elem()
-	result := dest
-	for t.Kind() == reflect.Ptr {
-		t = t.Elem()
-		dest.Set(reflect.New(t))
-		dest = dest.Elem()
-	}
-
```
glibsm: +1
was wondering what this code block is all about

dig_test.go
abhinav: I try to start anonymous functions on new lines but I see that the rest of the code doesn't do that. Fixing.

abhinav:
Copy of this PR in #83. Will be landed on dev once #73 lands.

PR #83, opened 2017-06-01
param: Pointers are not parameter objects

abhinav:
Copy of #81. Will land as a separate commit on dev once #73 lands.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -116,7 +116,7 @@ func (c *Container) provideInstance(val interface{}) error {
 	if vtype == _errType {
 		return errors.New("can't provide errors")
 	}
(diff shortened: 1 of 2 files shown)


dig.go
```diff
@@ -357,12 +354,12 @@ func (c *Container) createParamObject(t reflect.Type) (reflect.Value, error) {
 			case "true", "yes":
```
akshayjshah: Side question: why are we supporting both of these values? If we want to support multiple values, why don't we just use `strconv.ParseBool` and accept all the usual truthy strings?

abhinav: Good call. Will change in a different PR.

PR #86, opened 2017-06-03
Add basic string output of the container

glibsm:
Bringing back what was removed as part of the refactor - an easy way to quickly see what's inside the container.

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -723,22 +723,3 @@ func TestInvokeFailures(t *testing.T) {
 		assert.Error(t, c.Invoke(func() (int, error) { return 42, errors.New("oh no") }))
 	})
 }
-
-func TestStringer(t *testing.T) {
-	c := New()
-	type A struct{}
-	type B struct{}
-	require.NoError(t, c.Provide(func() (*A, *B) { return &A{}, &B{} }))
-	require.NoError(t, c.Invoke(func(a *A) {}))
-
-	b := &bytes.Buffer{}
-	fmt.Fprintln(b, c)
-	s := b.String()
-
-	assert.Contains(t, s, "*dig.A ->")
-	assert.Contains(t, s, "*dig.B ->")
-	assert.Contains(t, s, "func() (*dig.A, *dig.B)")
-
-	assert.Contains(t, s, "*dig.A =>")
-	assert.Contains(t, s, "*dig.B =>")
-}
--- a/dig_test.go
+++ b/dig_test.go
@@ -723,3 +723,18 @@ func TestInvokeFailures(t *testing.T) {
 		assert.Error(t, c.Invoke(func() (int, error) { return 42, errors.New("oh no") }))
 	})
 }
+
+func TestStringer(t *testing.T) {
+	c := New()
+	type A struct{}
+	type B struct{}
+	require.NoError(t, c.Provide(func() (*A, *B) { return &A{}, &B{} }))
+
+	b := &bytes.Buffer{}
+	fmt.Fprintln(b, c)
+	s := b.String()
+
+	require.Contains(t, s, "*dig.A ->")
+	require.Contains(t, s, "*dig.B ->")
+	require.Contains(t, s, "func() (*dig.A, *dig.B)")
+}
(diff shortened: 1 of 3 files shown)


dig_test.go
```diff
@@ -723,3 +723,22 @@ func TestInvokeFailures(t *testing.T) {
 		assert.Error(t, c.Invoke(func() (int, error) { return 42, errors.New("oh no") }))
 	})
 }
+
+func TestStringer(t *testing.T) {
+	c := New()
```
alsamylkin: Add t.Parallel()?

stringer.go
```diff
@@ -0,0 +1,30 @@
+package dig
+
+import (
+	"bytes"
+	"fmt"
+)
+
+// String representation of the entire Container
+func (c *Container) String() string {
```
alsamylkin: Why these functions are in a separate file? The different receiver type((pointer and value) in a String methods makes me to look in the other methods implementations.

dig_test.go
```diff
@@ -723,3 +723,22 @@ func TestInvokeFailures(t *testing.T) {
...
+	b := &bytes.Buffer{}
+	fmt.Fprintln(b, c)
+	s := b.String()
+
+	assert.Contains(t, s, "*dig.A ->")
+	assert.Contains(t, s, "*dig.B ->")
+	assert.Contains(t, s, "func() (*dig.A, *dig.B)")
+
+	assert.Contains(t, s, "*dig.A =>")
+	assert.Contains(t, s, "*dig.B =>")
```
alsamylkin: No test for `deps` and `constructor`?

stringer.go
glibsm: This functionality is very self-contained even though it belongs to the same package.
I'd like to keep `dig.go` as tight as possible -- just the core functionality.

stringer_test.go
```diff
@@ -0,0 +1,39 @@
...
+	require.NoError(t, c.Provide(func() (*A, *B) { return &A{}, &B{} }))
+	require.NoError(t, c.Provide(func(*B) *C { return &C{} }))
+	require.NoError(t, c.Invoke(func(a *A) {}))
+
+	b := &bytes.Buffer{}
+	fmt.Fprintln(b, c)
+	s := b.String()
+
+	// all nodes are in the graph
+	assert.Contains(t, s, "*dig.A -> deps: []")
```
alsamylkin: nit: if you use assert many times, you can use: `assert := assert.New(t)`

stringer.go
```diff
@@ -0,0 +1,30 @@
...
+	b := &bytes.Buffer{}
+	fmt.Fprintln(b, "nodes: {")
+	for k, v := range c.nodes {
+		fmt.Fprintln(b, "\t", k, "->", v)
+	}
+	fmt.Fprintln(b, "}")
+
+	fmt.Fprintln(b, "cache: {")
+	for k, v := range c.cache {
+		fmt.Fprintln(b, "\t", k, "=>", v)
```
akshayjshah: Why the different arrows?

stringer.go
```diff
@@ -0,0 +1,30 @@
...
+	}
+	fmt.Fprintln(b, "}")
+
+	return b.String()
```
akshayjshah: This is *so* close to being JSON - why not implement `MarshalJSON` instead?

stringer.go
glibsm: separating graph and cache outputs, open to suggestions

stringer.go
glibsm: Yeah, but it's not the intention to have a JSON output :) The target audience is for people using dig, not someone parsing the output of the graph.

That said, having it be *actual* JSON might be beneficial if we ever wanted to actually try and parse it for some reason (dunno, codegen or something)

stringer.go
akshayjshah: Works for me, just making sure it's intentional.

stringer.go
akshayjshah: :+1: sure.

glibsm:
Going to merge as-is since it's actually pretty useful even while writing new features of dig and unit testing.

We can iterate and convert to *actually* producing JSON later if we feel like it

Issue #80, opened 2017-06-01
Add support for named instances


glibsm:
Discussion on the subject: #30 

Now that we have a concept of tagging special structs via embedding `dig.Param`, we can expand the concept and support named instances of the same type in the graph as well.

Goes something like this:

```go
type Param struct {
	dig.In

	A1 *A `named:"firstA"`
	A2 *A `named:"secondA" optional:"true"`
}

type Output struct {
	dig.Out

	// Two instances of B will be contributed to the graph
	// with different names
	B1 *B `named:"mainB"`
	B2 *B `named:"secondaryB"`
}

func Constructor(p Param) *Output {
	// use p.A1 and p.A2
	// ...
	// return a custom type
}
```

Rough plan of attack:

- [x] Rename `dig.Param` to `dig.In`
- [x] Add a new type `dig.Out`
- [ ] Migrate container to use a composite key (reflect.Type + string name)
- [ ] Add support for `named:"<name>"` on `dig.In` and `dig.Out` types

PR #89, opened 2017-06-05
[2/4] Add dig.Out type

glibsm:
Second step of #80 

Needs basic README docs, but so does `dig.In` usage. Will work on a separate PR for it

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -28,14 +28,9 @@ import (
 	"strconv"
 )
 
-var (
-	_noValue         reflect.Value
-	_errType         = reflect.TypeOf((*error)(nil)).Elem()
-	_inInterfaceType = reflect.TypeOf((*digInObject)(nil)).Elem()
-	_inType          = reflect.TypeOf(In{})
+const (
+	_optionalTag = "optional"
 )
-
-const _optionalTag = "optional"
 
 // A Container is a directed, acyclic graph of dependencies. Dependencies are
 // constructed on-demand and returned from a cache thereafter, so they're
--- a/dig.go
+++ b/dig.go
@@ -221,25 +221,6 @@ func traverseOutTypes(t reflect.Type, f func(t reflect.Type) error) error {
 	return nil
 }
 
-// DFS over values and types for all nested dig.Outs
-func traverseOutValues(v reflect.Value, t reflect.Type, f func(reflect.Type, reflect.Value)) {
-	// dig.Out objects are not acted upon directly, but rather their memebers are considered
-	if isOutObject(t) {
-		for i := 0; i < t.NumField(); i++ {
-			field := t.Field(i)
-
-			ft := field.Type
-			fv := v.Field(i)
-
-			// recurse into other embedded Out objects
-			traverseOutValues(fv, ft, f)
-		}
-	} else {
-		// tun the provided function on the object itself (no need to recurse in)
-		f(t, v)
-	}
-}
-
 func (c *Container) isAcyclic(n node) error {
 	return detectCycles(n, c.nodes, nil)
 }
@@ -278,11 +259,7 @@ func (c *Container) get(t reflect.Type) (reflect.Value, error) {
 
 	for _, con := range constructed {
 		ct := con.Type()
-		traverseOutValues(con, ct, func(ft reflect.Type, fv reflect.Value) {
-			if ct != _errType {
-				c.cache[ft] = fv
-			}
-		})
+		c.set(ct, con)
 	}
 	return c.cache[t], nil
 }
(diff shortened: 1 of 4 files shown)


glibsm:
There are a lot of similarities for how we traverse over all nested objects in `dig.In` and `dig.Out`, yet they use completely different code paths.

There are also a lot of similarities for when we traverse over just types vs when we go over types *AND* values after a constructor has been called, there is probably some DRY-ness there too.

I will see if all four cases can be united, because the only things that differ are:
- Conditions for recursion (bool function would work)
- Types only, or values are also present

stringer.go
```diff
@@ -0,0 +1,30 @@
...
+	fmt.Fprintln(b, "}")
+
+	return b.String()
+}
+
+func (n node) String() string {
+	return fmt.Sprintf(
+		"deps: %v, constructor: %v", n.deps, n.ctype,
+	)
+}
```
glibsm: I will change the base branch, or merge this in first so it goes away from this PR

glibsm:
Dropping the WIP tag, as it's now fully working e2e and I was able to refactor the traverse

dig.go
```diff
@@ -172,6 +150,77 @@ func (c *Container) provideConstructor(ctor interface{}, ctype reflect.Type) err
...
+				continue // skip private fields
+			}
+
+			// keep recursing to traverse all the embedded objects
+			traverseOutTypes(ft, f)
+		}
+	} else {
+		// call the provided function on non-Out type
+		if err := f(t); err != nil {
+			return err
```
abhinav: nit: invert condition

```
if !isOutObject(t) {
  return f(t)
}

// ..
```

dig.go
```diff
@@ -210,14 +259,68 @@ func (c *Container) get(t reflect.Type) (reflect.Value, error) {
...
+				fv    = v.Field(i)
+			)
+			// recurse into all embedded objects
+			c.set(ft, fv)
+		}
+	} else {
+		// do not cache error types
+		if t != _errType {
+			c.cache[t] = v
+		}
```
abhinav: nit: invert if condition to reduce nesting

```go
if !isOutObject(t) {
  if t != _errType {
    c.cache[t] = v
  }
  return
}

// ..
```

dig.go
```diff
@@ -210,14 +259,69 @@ func (c *Container) get(t reflect.Type) (reflect.Value, error) {
 
 	for _, con := range constructed {
 		ct := con.Type()
-		if ct == _errType {
-			continue
-		}
-		c.cache[ct] = con
+		c.set(ct, con)
```
akshayjshah: Can `set` take just `con` and handle getting the type internally? Prevents unintentional mismatches between the passed value and type.

dig.go
```diff
@@ -210,14 +259,69 @@ func (c *Container) get(t reflect.Type) (reflect.Value, error) {
...
+func (c *Container) set(t reflect.Type, v reflect.Value) {
+	if !isOutObject(t) {
+		// do not cache error types
+		if t != _errType {
+			c.cache[t] = v
+		}
+		return
+	}
+
+	// dig.Out objects are not acted upon directly, but rather their memebers are considered
```
akshayjshah: spelling: members

dig.go
```diff
@@ -210,14 +259,69 @@ func (c *Container) get(t reflect.Type) (reflect.Value, error) {
...
+	for i := 0; i < t.NumField(); i++ {
+		var (
+			field = t.Field(i)
+			ft    = field.Type
+			fv    = v.Field(i)
+		)
+		// recurse into all embedded objects
```
akshayjshah: nit: these are fields, but they're not necessarily embedded. https://golang.org/ref/spec#Struct_types

dig_test.go
```diff
@@ -427,6 +427,31 @@ func TestEndToEndSuccess(t *testing.T) {
...
+		}
+		type B struct {
+			name string
+		}
+		type Ret struct {
+			Out
+
+			foo string // private field to be ignored
+			A          // value type A
+			*B         // pointer type *B
```
akshayjshah: nit: embeds at the top.

dig_test.go
```diff
@@ -427,6 +427,31 @@ func TestEndToEndSuccess(t *testing.T) {
...
+		}
+
+		c := New()
+		require.NoError(t, c.Provide(func() Ret {
+			return Ret{A: A{"string A"}, B: &B{"string B"}}
+		}), "provide for the Ret struct should succeed")
+		require.NoError(t, c.Invoke(func(a A, b *B) {
+			assert.Equal(t, a.name, "string A", "value type should work for dig.Out")
+			assert.Equal(t, b.name, "string B", "pointer sohuld work for dig.Out")
```
akshayjshah: spelling: "should"

Also, can we be nit-picky and also validate that we're getting the same ref returned from the constructors?

dig.go
glibsm: memebears!

PR #90, opened 2017-06-05
Fix detectCyles duplicate dependencies bug

amckinney:
This refactors the cycle detection `seen` map to use a linear search over the path. There was previously an error when multiple dependencies depended on the same types, as reflected in the added unit test.

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -435,8 +435,8 @@ func TestEndToEndSuccess(t *testing.T) {
 		c := New()
 
 		require.NoError(t, c.Provide(func() A { return A{} }), "provide failed")
-		require.NoError(t, c.Provide(func(_ A) B { return B{} }), "provide failed")
-		require.NoError(t, c.Provide(func(_ A, _ B) C { return C{} }), "provide failed")
+		require.NoError(t, c.Provide(func(A) B { return B{} }), "provide failed")
+		require.NoError(t, c.Provide(func(A, B) C { return C{} }), "provide failed")
 	})
 }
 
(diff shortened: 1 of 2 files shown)


dig_test.go
```diff
@@ -426,6 +426,18 @@ func TestEndToEndSuccess(t *testing.T) {
...
+	t.Run("nested dependencies", func(t *testing.T) {
+		type A struct{}
+		type B struct{}
+		type C struct{}
+
+		c := New()
+
+		require.NoError(t, c.Provide(func() A { return A{} }), "provide failed")
+		require.NoError(t, c.Provide(func(dep0 A) B { return B{} }), "provide failed")
+		require.NoError(t, c.Provide(func(dep0 A, dep1 B) C { return C{} }), "provide failed")
```
akshayjshah: small nit: rather than using throwaway names, you can do `func(_ A, _ B) C`.

dig_test.go
```diff
@@ -426,6 +426,18 @@ func TestEndToEndSuccess(t *testing.T) {
...
+	t.Run("nested dependencies", func(t *testing.T) {
+		type A struct{}
+		type B struct{}
+		type C struct{}
+
+		c := New()
+
+		require.NoError(t, c.Provide(func() A { return A{} }), "provide failed")
+		require.NoError(t, c.Provide(func(_ A) B { return B{} }), "provide failed")
+		require.NoError(t, c.Provide(func(_ A, _ B) C { return C{} }), "provide failed")
```
glibsm: even smaller nit: you actually don't need the `_` at all. `func(A, B)` will also work.

akshayjshah: orly? Learned something new today.

PR #91, opened 2017-06-06
Update the changelog

glibsm:
It's quite stale

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,12 @@
 
 
 ## v0.4 (unreleased)
-- Add support form `map`, `slice` and `array` to dig
-- Remove `Must*` funcs to greatly reduce API surface area.
-- Remove `*All` funcs to reduce API surface area. Available methods on
-container are `Provide`, `Resolve` and `Invoke`
+
+- **[Breaking]** Remove `Must*` funcs to greatly reduce API surface area.
+- **[Breaking]** Restrict the API surface to only `Provide` and `Invoke`.
 - Providing constructors with common returned types results in an error.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,12 +2,12 @@
...
-- Add support form `map`, `slice` and `array` to dig
-- Remove `Must*` funcs to greatly reduce API surface area.
-- Remove `*All` funcs to reduce API surface area. Available methods on
-container are `Provide`, `Resolve` and `Invoke`
+
+- **[Breaking]** Remove `Must*` funcs to greatly reduce API surface area.
+- **[Breaking]** Restrict the API surface to only `Provide` and `Invoke`.
 - Providing constructors with common returned types results in an error.
-- Update `Provide` and `Resolve` methods to accept variadic arguments.
+- **[Breaking]** Update `Provide` method to accept variadic arguments.
```
alsamylkin: I think dig is not concurrent safe anymore.

CHANGELOG.md
```diff
@@ -2,12 +2,12 @@
 
 
 ## v0.4 (unreleased)
-- Add support form `map`, `slice` and `array` to dig
```
alsamylkin: Why do you delete entries, not append?

CHANGELOG.md
glibsm: technically that was never in the changelog in the first place, i.e. there is no reversal of previous feature :P

CHANGELOG.md
glibsm: Because we don't *explicitly* support only map and slice anymore, but rather all pointer-like and value types that are no primitives

PR #92, opened 2017-06-06
Add test covering primitives

glibsm:
Out of curiosity I wanted to see if we supported primitives, might as well check it in.

This looks silly right now (since you'd only be able to have one string type in the graph), however in combination with the incoming named instances through `dig.Out` will be quite powerful without the need to typealias constantly

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -444,11 +444,13 @@ func TestEndToEndSuccess(t *testing.T) {
 		c := New()
 		require.NoError(t, c.Provide(func() string { return "piper" }), "string provide failed")
 		require.NoError(t, c.Provide(func() int { return 42 }), "int provide failed")
+		require.NoError(t, c.Provide(func() int64 { return 24 }), "int provide failed")
 		require.NoError(t, c.Provide(func() time.Duration {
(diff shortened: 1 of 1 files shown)


dig_test.go
```diff
@@ -438,6 +438,16 @@ func TestEndToEndSuccess(t *testing.T) {
 		require.NoError(t, c.Provide(func(A) B { return B{} }), "provide failed")
 		require.NoError(t, c.Provide(func(A, B) C { return C{} }), "provide failed")
 	})
+
+	t.Run("primitives", func(t *testing.T) {
+		c := New()
+		require.NoError(t, c.Provide(func() string { return "piper" }), "string provide failed")
+		require.NoError(t, c.Provide(func() int { return 42 }), "int provide failed")
```
alsamylkin: Maybe it is out of scope, but can you add a case for time.Duration, which is an alias to int64?

alsamylkin: If you replace the line above with:
```golang
require.NoError(t, c.Provide(func() int64 { return 42 }), "int64 provide failed")
```
It will be even more awesome.

PR #94, opened 2017-06-06
optional: Fix support for constructors

abhinav:
The current implementation of optionals has a bug which causes failures
if a constructor for a provided type has `dig.In` struct with optional
parameters.

That is, given,

    type param struct {
        In

        Buffer *bytes.Buffer `optional:"true"`
    }

    c.Provide(func(p param) *Something { ... })

The following will fail, complaining that `*bytes.Buffer` isn't
available in the container.

    c.Invoke(func(*Something) { .. })

This fixes this issue by storing more information about node
dependencies: specifically, whether they are required or optional.

Test Plan: A failing test case was added before the fix was implemented.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -262,57 +262,26 @@ type dep struct {
 }
 
 func newNode(provides reflect.Type, ctor interface{}, ctype reflect.Type) (node, error) {
-	deps := make([]dep, 0, ctype.NumIn())
-	for i := 0; i < ctype.NumIn(); i++ {
-		t := ctype.In(i)
-		if isInObject(t) {
-			newDeps, err := getInDependencies(t)
-			if err != nil {
-				return node{}, err
-			}
-			deps = append(deps, newDeps...)
-		} else {
-			deps = append(deps, dep{Type: t})
-		}
-	}
-
+	deps, err := getConstructorDependencies(ctype)
 	return node{
 		provides: provides,
 		ctor:     ctor,
 		ctype:    ctype,
 		deps:     deps,
-	}, nil
+	}, err
 }
 
-// Retrieves the dependencies for a dig.In object
-func getInDependencies(t reflect.Type) ([]dep, error) {
-	deps := make([]dep, 0, t.NumField())
-	for i := 0; i < t.NumField(); i++ {
-		f := t.Field(i)
-		if f.PkgPath != "" {
-			continue // skip private fields
-		}
-
-		if isInObject(f.Type) {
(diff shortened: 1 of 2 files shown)


dig.go
```diff
@@ -270,21 +284,36 @@ func newNode(provides reflect.Type, ctor interface{}, ctype reflect.Type) (node,
...
 }
 
-// Retrives the dependencies for the parameter of a constructor.
-func getConstructorDependencies(t reflect.Type) []reflect.Type {
-	if !isInObject(t) {
-		return []reflect.Type{t}
-	}
-
-	deps := make([]reflect.Type, 0, t.NumField())
+// Retrives the dependencies for a dig.In object
```
breerly: Retrieves*

dig_test.go
```diff
@@ -709,6 +733,34 @@ func TestInvokeFailures(t *testing.T) {
 		require.Contains(t, err.Error(), "dig.type2 isn't in the container")
 	})
 
+	t.Run("unmet constructor dependency", func(t *testing.T) {
```
breerly: unmet?

abhinav: Yeah, as in a dependency/requirement that wasn't met. Same wording is used by the previous test.

dig.go
```diff
@@ -253,13 +253,27 @@ type node struct {
 	provides reflect.Type
 	ctor     interface{}
 	ctype    reflect.Type
-	deps     []reflect.Type
+	deps     []dep
+}
+
+type dep struct {
```
glibsm: you're half way to swapping out `map[reflect.Type]` for `map[key]` to support named types :)

dig.go
```diff
@@ -253,13 +253,27 @@ type node struct {
...
+	Optional bool
 }
 
 func newNode(provides reflect.Type, ctor interface{}, ctype reflect.Type) (node, error) {
-	deps := make([]reflect.Type, 0, ctype.NumIn())
+	deps := make([]dep, 0, ctype.NumIn())
 	for i := 0; i < ctype.NumIn(); i++ {
-		deps = append(deps, getConstructorDependencies(ctype.In(i))...)
+		t := ctype.In(i)
+		if isInObject(t) {
```
glibsm: You can fold all this into the `getInDependencies` method, similar to `traverseOutTypes`.

Maybe even call it something similar as well, up to you

abhinav: Good call. Added a `traverseInTypes` function. Greatly simplifies this code.

PR #96, opened 2017-06-08
Invoke: Return errors as-is

abhinav:
This changes Invoke to return the error returned by the function as-is
rather than wrapping it in Errorf. As a caller of Invoke, I would expect
my function's returned error to be returned back to me. Otherwise, I'm
losing a lot of information if my error object was richer than
`fmt.Errorf`'s string.

Further, the existing message didn't really add a lot of information: it
added the address of the function to the error message.

Depends on #95 so that tests don't fail.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -102,8 +102,10 @@ func (c *Container) Invoke(function interface{}) error {
 	if len(returned) == 0 {
 		return nil
 	}
-	if last := returned[len(returned)-1]; last.Type() == _errType && last.Interface() != nil {
-		return fmt.Errorf("failed to execute %v (type %v): %v", function, ftype, last.Interface())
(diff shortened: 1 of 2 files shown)


dig.go
```diff
@@ -102,8 +102,10 @@ func (c *Container) Invoke(function interface{}) error {
...
 		return nil
 	}
-	if last := returned[len(returned)-1]; last.Type() == _errType && last.Interface() != nil {
-		return fmt.Errorf("failed to execute %v (type %v): %v", function, ftype, last.Interface())
+	if last := returned[len(returned)-1]; last.Type() == _errType {
+		if err, _ := last.Interface().(error); err != nil {
+			return err
+		}
 	}
 	return nil
```
breerly: Actually surprised this path isn't exercised in any tests.

PR #102, opened 2017-06-12
Add licenses and `make license` target, part of lint

glibsm:


Makefile
```diff
@@ -31,7 +31,7 @@ else
 endif
 
 .PHONY: lint
-lint:
+lint: license
```
abhinav: does check_license not take care of this?

glibsm: Huh. Well clearly something is afoot, because builds have not been failing and we've been missing headers.

I will look into it more and consolidate the two

Makefile
```diff
@@ -44,10 +44,9 @@ ifdef SHOULD_LINT
 	@$(foreach dir,$(PKGS),golint $(dir) 2>&1 | tee -a lint.log;)
 	@echo "Checking for unresolved FIXMEs..."
 	@git grep -i fixme | grep -v -e vendor -e Makefile | tee -a lint.log
-	@echo "Ensuring generated doc.go are up to date"
-	$(ECHO_V)$(MAKE) gendoc
 	@echo "Checking for license headers..."
-	@./check_license.sh | tee -a lint.log
+	@DRY_RUN=1 ./check_license.sh | tee -a lint.log
+	@$(MAKE) gendoc
```
abhinav: Why is lint changing code? Lint should really just check that the code is up to date.

glibsm: Open to suggestions for a better approach (I just moved this line, not introduced it in the PR).
How else do we ensure docs are up to date?

abhinav: This isn't a major concern, TBH. One way is obviously to check if the git working tree is clean. Another would be to make md-to-godoc support a `--dry-run`/`--diff` flag which reports what changes would be made rather than actually changing.

PR #103, opened 2017-06-13
[WIP] [3/4] Swap the map key for a custom struct

glibsm:
Third pr in the #80 series. Getting there!

This no longer uses `reflect.Type` to uniquely identify an object in the graph, all the options are considered (right now only the optional true/false, but soon a string to identify the name as well).

At the moment, this has a side-effect: `optional Type1` and `Type1` are considered different nodes and can co-exist. I haven't thought all the way through if this should be fixed, or if it's actually a good thing. Let me know what you think.

This PR also removed the `dep struct`, as node dependencies now know the exact key of their dependencies.

Looks rather invasive, but in most cases it's pretty straight forward swap from `reflect.Type` usage to `nodeKey`

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -299,11 +299,11 @@ func (c *Container) set(v reflect.Value) {
 	}
 }
 
-func (c *Container) contains(deps []dep) error {
+func (c *Container) contains(keys []nodeKey) error {
 	var missing []reflect.Type
-	for _, d := range deps {
-		if _, ok := c.nodes[key(d.Type)]; !ok && !d.Optional {
-			missing = append(missing, d.Type)
+	for _, k := range keys {
+		if _, ok := c.nodes[k]; !ok && !k.optional {
+			missing = append(missing, k.t)
 		}
 	}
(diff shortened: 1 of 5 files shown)


dig.go
```diff
@@ -346,15 +350,16 @@ func newNode(provides reflect.Type, ctor interface{}, ctype reflect.Type) (node,
 		ctor:     ctor,
 		ctype:    ctype,
 		deps:     deps,
+		key:      key(provides),
```
glibsm: hm, this looks wrong. `newNode` should take in the key which is being constructed, not derive one itself

glibsm:
Tagging as WIP for now. Feel free to take a look to make sure I'm going down the right path, but looks like there are a few more function signatures that need to change

glibsm:
derp. This means that a single name instance in the graph would only be retrievable by name.

Going to leave this branch up and prototype not swapping out the main key, but rather add options to the `get()` method.

glibsm:
This won't work. 
Will open a few PRs in a new direction

PR #105, opened 2017-06-14
Fold the caching mechanism into node struct

glibsm:
Inching ever closer to #80.

This removes the double-map nature of dig, and stores the cache values on the node directly. While it might seem like a meaningless diff, it actually paves the way to extend the graph to have multiple objects of the same type, tagged of course.

The nodes map will soon have to be switched to `nodes map[reflect.Type][]node` type, which obviously makes the cache very difficult to deal with. In this world, at least only one data structure has to be operated on at a time, without the need to sync with the cache.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -37,14 +37,12 @@ const (
 // effectively singletons.
 type Container struct {
 	nodes map[reflect.Type]node
-	cache map[reflect.Type]reflect.Value
 }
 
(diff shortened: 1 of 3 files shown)


glibsm:
I'll open a large e2e PR first, can be chopped up later. It's too confusing to juggle these branches before everything is working

PR #106, opened 2017-06-14
Swap out dep for key, and a few other refactors

glibsm:
A couple of more small hops towards #80 

Simplify the handling of optional tags by putting that responsibility
onto get.

Start documentation of the get behavior, before it gets expanded even
further to support named types.

PR against #105 , can be rebased after that lands or merged first.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -202,8 +202,35 @@ func (c *Container) isAcyclic(n node) error {
 	return detectCycles(n, c.nodes, nil)
 }
 
-// Retrieve a type from the container
-func (c *Container) get(t reflect.Type) (reflect.Value, error) {
+type keyOpt func(*key)
+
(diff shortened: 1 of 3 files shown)


dig.go
```diff
@@ -202,15 +202,45 @@ func (c *Container) isAcyclic(n node) error {
 	return detectCycles(n, c.nodes, nil)
 }
 
-// Retrieve a type from the container
-func (c *Container) get(t reflect.Type) (reflect.Value, error) {
+type keyOpt func(*key)
```
glibsm: can technically specify `getOpt` and later `setOpt`, but they will both operate on keys and look somewhat similar.

glibsm:
I'll open a large e2e PR first, can be chopped up later. It's too confusing to juggle these branches before everything is working

I will open the small `isOptional` refactor on it's own

PR #108, opened 2017-06-14
[WIP] Named instances

glibsm:
Consolidating all PRs into one WIP, since it's really only useful to see the atomic addition.

Needs a rebase, but comment away on the approach.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -204,8 +204,35 @@ func (c *Container) isAcyclic(n node) error {
 	return detectCycles(n, c.nodes, nil)
 }
 
-// Retrieve a type from the container
-func (c *Container) get(t reflect.Type) (reflect.Value, error) {
+type keyOpt func(*key)
+
+// Specifies if absence of the object in the graph is ok on retrieval.
+func optional(opt bool) keyOpt {
+	return func(k *key) {
+		k.optional = opt
+	}
+}
+
+// Represents an edge in the graph.
+type key struct {
+	t        reflect.Type
+	optional bool
+}
+
+// Custom map get method for the underlying nodes.
+//
+// Behaves in the following ways:
+//     - get without any options will return a value from the graph, or an error
+//     - get for type not in the map with optional=true returns zero value for type.
+//     - get with a dig.In embed will construct a new object, and populate all the
+//       members based on the memebers, including a parse of the struct tags that
+//       drives the aforementioned behaviors.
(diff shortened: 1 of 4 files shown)


dig.go
```diff
@@ -206,10 +204,33 @@ func (c *Container) isAcyclic(n node) error {
...
 }
 
-// Retrieve a type from the container
-func (c *Container) get(t reflect.Type) (reflect.Value, error) {
-	if v, ok := c.cache[t]; ok {
-		return v, nil
+type keyOpt func(*key)
+
+// Specifies if absence of the object in the graph is ok on retrieval.
+func optional(opt bool) keyOpt {
```
abhinav: Optional: This pattern is probably not required here. Seems to me that `key`/`edge` expresses all the requirements for the query for a node. So you could probably just write `get(key) (reflect.Value, error)`.

dig_test.go
```diff
@@ -462,6 +462,35 @@ func TestEndToEndSuccess(t *testing.T) {
 			require.NotNil(t, a, "*A should be part of the container through Ret2->Ret1")
 		}))
 	})
+
+	t.Run("named instances", func(t *testing.T) {
```
alsamylkin: More tests are in order? I'll think about circular dependency resolution(named instances could have a different dependency graphs that are valid, but could look like a loop if you drop names), what should we return if there is only a named instance(or two) for the unnamed type and reverse.

dig.go
```diff
@@ -206,10 +204,33 @@ func (c *Container) isAcyclic(n node) error {
...
+	return func(k *key) {
+		k.optional = opt
+	}
+}
+
+// Represents an edge in the graph.
+type key struct {
+	t        reflect.Type
+	optional bool
```
glibsm: no, it's a property of the constructor

dig_test.go
glibsm: was tagged as WIP for a reason

dig.go
```diff
@@ -37,14 +37,12 @@ const (
 // effectively singletons.
 type Container struct {
 	nodes map[reflect.Type]node
-	cache map[reflect.Type]reflect.Value
```
glibsm: they do, because `reflect.Type` alone doesn't uniquely identify an object anymore

glibsm:
Closed in favor of #109

PR #109, opened 2017-06-14
Named instances

### LINKED ISSUE #80: Add support for named instances

glibsm:
On the third time of asking, here's one atomic PR that solves the rest of what's needed for #80 

Object are now uniquely identified by `reflect.Type:string`, everything else stems from there. Far cleaner than the lasts two attempts to get this in.

Optional edges are prepended with `~` now in the output, i.e. required A, named B and optional C look like this now: `deps: [A, B:foo, ~C]`

Fixes #80

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -30,21 +30,36 @@ import (
 
 const (
 	_optionalTag = "optional"
+	_nameTag     = "name"
 )
+
+// Unique identification of an object in the graph.
+type key struct {
+	t    reflect.Type
+	name string
+}
 
 // A Container is a directed, acyclic graph of dependencies. Dependencies are
 // constructed on-demand and returned from a cache thereafter, so they're
 // effectively singletons.
 type Container struct {
-	nodes map[reflect.Type]node
-	cache map[reflect.Type]reflect.Value
+	nodes map[key]*node
+	cache map[key]reflect.Value
+	// TODO: for advanced use-case, add an index
+	// This will allow retrieval of a single type, without specifying the exact
+	// tag, provided there is only one object of that given type
+	//
+	// It will also allow library owners to create a "default" tag for their
+	// object, in case users want to provide another type with a different name
+	//
+	// index map[reflect.Type]key
 }
 
 // New constructs a ready-to-use Container.
 func New() *Container {
 	return &Container{
-		nodes: make(map[reflect.Type]node),
-		cache: make(map[reflect.Type]reflect.Value),
+		nodes: make(map[key]*node),
+		cache: make(map[key]reflect.Value),
 	}
 }
 
@@ -106,19 +121,19 @@ func (c *Container) Invoke(function interface{}) error {
 }
 
 func (c *Container) provide(ctor interface{}, ctype reflect.Type) error {
-	returnTypes, err := c.getReturnTypes(ctor, ctype)
+	keys, err := c.getReturnKeys(ctor, ctype)
 	if err != nil {
 		return fmt.Errorf("unable to collect return types of a constructor: %v", err)
 	}
 
-	nodes := make([]node, 0, len(returnTypes))
-	for rt := range returnTypes {
-		n, err := newNode(rt, ctor, ctype)
+	nodes := make([]*node, 0, len(keys))
+	for k := range keys {
+		n, err := newNode(k, ctor, ctype)
 		if err != nil {
 			return err
 		}
 		nodes = append(nodes, n)
-		c.nodes[rt] = n
+		c.nodes[k] = n
 	}
 
 	for _, n := range nodes {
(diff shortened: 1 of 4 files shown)


dig.go
```diff
@@ -235,13 +250,13 @@ func (c *Container) get(t reflect.Type) (reflect.Value, error) {
 	// Provide-time validation ensures that all constructors return at least
 	// one value.
 	if err := constructed[len(constructed)-1]; err.Type() == _errType && err.Interface() != nil {
-		return _noValue, fmt.Errorf("constructor %v for type %v failed: %v", n.ctype, t, err.Interface())
+		return _noValue, fmt.Errorf("constructor %v for type %v failed: %v", n.ctype, e.t, err.Interface())
 	}
 
 	for _, con := range constructed {
-		c.set(con)
+		c.set(key{t: con.Type()}, con)
```
abhinav: This doesn't seem right. If I did `get(key{name: "foo", t: someType})`, this will `set(key{t: someType})`.

abhinav: Nvm, I get it. Named get means that `con.Type()` is always a `dig.Out` object; and for that we'll recurse into it. Consider adding a comment explaining this.

dig.go
```diff
@@ -175,55 +190,55 @@ func (c *Container) getReturnTypes(
...
 		// We do not want parameter objects to be cached.
-		return c.createInObject(t)
+		return c.createInObject(e.t)
 	}
 
-	n, ok := c.nodes[t]
+	n, ok := c.nodes[e.key]
 	if !ok {
-		return _noValue, fmt.Errorf("type %v isn't in the container", t)
+		return _noValue, fmt.Errorf("type %v isn't in the container", e.t)
```
abhinav: what about `e.optional`? you're not using it here.

dig.go
```diff
@@ -259,7 +274,7 @@ func (c *Container) createInObject(t reflect.Type) (reflect.Value, error) {
 			return dest, err
 		}
 
-		v, err := c.get(f.Type)
+		v, err := c.get(edge{key: key{t: f.Type, name: f.Tag.Get(_nameTag)}})
```
abhinav: `edge{..., optional: isOptional}`? currently the lines below are applying the optional property.

dig.go
```diff
@@ -275,28 +290,30 @@ func (c *Container) createInObject(t reflect.Type) (reflect.Value, error) {
...
 		// do not cache error types
-		if t != _errType {
-			c.cache[t] = v
+		if k.t != _errType {
+			c.cache[k] = v
 		}
 		return
 	}
 
 	// dig.Out objects are not acted upon directly, but rather their memebers are considered
```
abhinav: typo: members

dig_test.go
```diff
@@ -462,6 +462,103 @@ func TestEndToEndSuccess(t *testing.T) {
...
+			return Ret2{
+				Ret1: Ret1{
+					A1: A{1},
+				},
+				A2: A{2},
+			}
+		}))
+		require.NoError(t, c.Invoke(func(p param) {
+			assert.Equal(t, 1, p.A1.idx)
+			assert.Equal(t, 2, p.A2.idx)
```
abhinav: can you add another test where you `Invoke` on a type which depends on named parameters?

```
type fooArgs struct {
  dig.In

  A1 A `name:"first"`
  A2 A `name:"second"`
}

c.Provide(func(args fooArgs) (*foo, error) { .. })

c.Invoke(func(f *foo) {
 ..
})
```

dig.go
glibsm: yeah we can fold this into a get, now that it takes an edge as a param

dig_test.go
glibsm: 👍

dig.go
glibsm: That's right, the entry point is either a specific type only, or In/Out which then further narrow it down. I will try to convey that through a comment

dig.go
glibsm: It's used outside of `.get` right now. I will move that code in here

dig.go
```diff
@@ -132,36 +147,36 @@ func (c *Container) provide(ctor interface{}, ctype reflect.Type) error {
...
 			}
-			if _, ok := returnTypes[rt]; ok {
-				return fmt.Errorf("returns multiple %v", rt)
+			if _, ok := returnTypes[k]; ok {
+				return fmt.Errorf("returns multiple %v", k.t)
 			}
-			if _, ok := c.nodes[rt]; ok {
-				return fmt.Errorf("provides type %v, which is already in the container", rt)
+			if _, ok := c.nodes[k]; ok {
+				return fmt.Errorf("provides type %v, which is already in the container", k.t)
```
amckinney: We might want to update the error messages here. Now that we are giving users the ability to register multiple instances of the same type, the error message could be something along the lines of:
`return fmt.Errorf("provides type %v with name %s, which is already in the container", k.t, k.name)`

The same applies to the error message above, too:
`return fmt.Errorf("returns multiple %v with name %s", k.t, k.name)`

glibsm: That's actually a good point - error messaging.

But I think we can do it even better by just including the `key` to identify which object fails (that already handles for printing out the name as well, if it's present)

glibsm:
As of this PR, and error messaging changes I'm proposing a convention.

Objects are now identified either by type directly (unnamed), such as `*bytes.Buffer`, or in conjunction with the name `*bytes.Buffer:primary`.

dig_test.go
```diff
@@ -727,6 +876,21 @@ func TestInvokeFailures(t *testing.T) {
 		require.Contains(t, err.Error(), "dig.type2 isn't in the container")
 	})
 
+	t.Run("unmet named dependency", func(t *testing.T) {
```
alsamylkin: Could you add a similar test, but that has a type without a name provided?
Also reverse, there is a named type, but we ask for a default.

glibsm: That is not supported yet. See the TODO about an index.
I would prefer to defer that to a new PR

dig.go
```diff
@@ -30,21 +30,36 @@ import (
 
 const (
 	_optionalTag = "optional"
+	_nameTag     = "name"
 )
 
+// Unique identification of an object in the graph.
+type key struct {
+	t    reflect.Type
+	name string
```
alsamylkin: is it case sensitive?

glibsm: Yes, added a test.
We can change in the future if for reason we think we should be case insensitive

PR #113, opened 2017-06-16
Manual sentinel value detection

abhinav:
This changes our detection of `dig.In` and `dig.Out` structs from
relying on Go's automatic wrapper method generation to a manual check
for that embedded value. We do this by simply performing a breadth first
search of embedded fields of a struct.

Note about performance: There usually won't be a lot of embedding in
constructor struct types so the function should return very quickly. If
performance ever becomes an issue, we can memoize types that are known
to embed `dig.In` or `dig.Out` in a global cache similar to how
`encoding/json` caches reflection information for types.

By not relying on wrapper method generation, we are able to support
struct types generated with `reflect.StructOf`, as a documented
limitation of these structs is,

> StructOf currently does not generate wrapper methods for embedded
> fields. This limitation may be lifted in a future version.

This opens up more complex use cases where reflection is required to
generate the `Invoke`d or `Provide`d function.

---

Also note that I added tip to .travis.yml so that we can verify this
behavior with Go 1.9 and newer.

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -676,7 +676,7 @@ func TestEndToEndSuccess(t *testing.T) {
 		})
 
 		fn := reflect.MakeFunc(
-			reflect.FuncOf(nil /* params */, []reflect.Type{outType}, false /* variadict */),
+			reflect.FuncOf(nil /* params */, []reflect.Type{outType}, false /* variadic */),
 			func([]reflect.Value) []reflect.Value {
 				result := reflect.New(outType).Elem()
 				result.Field(1).Set(reflect.ValueOf(&A{Value: 1}))
--- a/dig_test.go
+++ b/dig_test.go
@@ -27,6 +27,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


dig_test.go
```diff
@@ -590,6 +591,115 @@ func TestEndToEndSuccess(t *testing.T) {
...
+			},
+			{
+				Name: "Bar",
+				Type: reflect.TypeOf(&A{}),
+				Tag:  `name:"bar"`,
+			},
+		})
+
+		fn := reflect.MakeFunc(
+			reflect.FuncOf(nil /* params */, []reflect.Type{outType}, false /* variadict */),
```
glibsm: typo: variadic

dig_test.go
```diff
@@ -590,6 +591,115 @@ func TestEndToEndSuccess(t *testing.T) {
...
+
+		require.NoError(t, c.Provide(new1), "failed to provide constructor")
+
+		// We generate a struct that embeds dig.In.
+		//
+		// Note that the fix for https://github.com/golang/go/issues/18780
+		// requires that StructField.Name is always set but versions of Go
+		// older than 1.9 expect Name to be empty for embedded fields.
+		//
+		// We use utils_for_go19_test and utils_for_pre_go19_test with build
```
glibsm: nice

types.go
```diff
@@ -42,47 +52,52 @@ var (
 //      T *Thingy `optional:"true"`
 //    }
 //
-type In struct{}
+type In struct{ digSentinel }
```
glibsm: sentinel?

abhinav: I figured since `dig.In` are like sentinel nodes in the graph of embedded types, that terminology could be used.


https://en.wikipedia.org/wiki/Sentinel_node

abhinav: To clarify: digSentinel isn't used at all but I want dig.In and dig.Out as having unexported or private fields rather than a plain `struct{}`.

glibsm: Ah, I thought you've been busy playing No Man's Sky.
I think the terminology is appropriate here.

dig_test.go
```diff
@@ -590,6 +591,115 @@ func TestEndToEndSuccess(t *testing.T) {
...
+
+				require.True(t, args[0].Field(2).IsNil(), "type2 must be nil")
+				return nil
+			},
+		)
+
+		require.NoError(t, c.Invoke(fn.Interface()), "invoke failed")
+	})
+
+	t.Run("dynamically generated dig.Out", func(t *testing.T) {
```
glibsm: now we're talking...

PR #114, opened 2017-06-20
Allow optional dependencies to have missing deps

### LINKED ISSUE #111: Missing dependencies for optional dependencies should not blow up
If A has an optional dependency on B, and B has a dependency on C,

    A --opt--> B -------> C

And C is not present in the graph, currently, we will fail to resolve A.

Instead, we should invoke A and pretend like B isn't present in the graph
because all its dependencies aren't present.

Relevant failing test:

```go
type C struct{}

type B struct{ C *C }

newB := func(c *C) *B { return &B{C: c} }

type A struct{ B *B }

type AParams struct {
  dig.In
  
  B *B `optional:"yes"`
}

newA := func(p AParams) *A { return &A{B: p.B} }

// We Provide A and B but not C.
c := dig.New()
c.Provide(newA, newB)

assert.NoError(t, c.Invoke(func(a *A) {
  assert.Nil(t, a.B)
}), "invoke failed")
```

akshayjshah:
If an Invoke takes an optional dependency on type `A`, but `A`'s
provided constructor has unmet dependencies, we should provide the zero
value. This makes optional params work as most users would expect.

Fixes #111.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -284,9 +284,10 @@ func (c *Container) createInObject(t reflect.Type) (reflect.Value, error) {
 
 		e := edge{key: key{t: f.Type, name: f.Tag.Get(_nameTag)}, optional: isOptional}
 		v, err := c.get(e)
-		if err != nil {
-			return dest, fmt.Errorf(
-				"could not get field %v (edge %v) of %v: %v", f.Name, e, t, err)
+		if isOptional && err != nil {
+			v = reflect.Zero(f.Type)
+		} else if err != nil {
+			return dest, fmt.Errorf("could not get field %v (edge %v) of %v: %v", f.Name, e, t, err)
 		}
 
 		dest.Field(i).Set(v)
--- a/dig.go
+++ b/dig.go
@@ -33,6 +33,11 @@ const (
 	_nameTag     = "name"
 )
 
+// Sentinel error for missing constructors.
+type errMissing string
+
+func (e errMissing) Error() string { return string(e) }
+
 // Unique identification of an object in the graph.
 type key struct {
 	t    reflect.Type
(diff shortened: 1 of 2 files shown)


akshayjshah:
Fixed.

dig.go
```diff
@@ -241,7 +246,7 @@ func (c *Container) get(e edge) (reflect.Value, error) {
 	}
 
 	if err := c.contains(n.deps); err != nil {
-		return _noValue, fmt.Errorf("missing dependencies for %v: %v", e.key, err)
+		return _noValue, errMissing(fmt.Sprintf("missing dependencies for %v: %v", e.key, err))
```
abhinav: Optional: IMO it would be preferable for `errMissing` to hold a list of missing dependencies and construct the message in `Error()`.

dig.go
```diff
@@ -285,10 +290,12 @@ func (c *Container) createInObject(t reflect.Type) (reflect.Value, error) {
 		e := edge{key: key{t: f.Type, name: f.Tag.Get(_nameTag)}, optional: isOptional}
 		v, err := c.get(e)
 		if err != nil {
-			return dest, fmt.Errorf(
-				"could not get field %v (edge %v) of %v: %v", f.Name, e, t, err)
+			if _, ok := err.(errMissing); ok {
+				v = reflect.Zero(f.Type)
```
abhinav: Wouldn't you rather do this in `get()` where you're returning `errMissing`?

```go
if err := c.contains(n.deps); err != nil {
  if e.optional { .. }
  else { .. }
}
```

That would actually let us drop the error type altogether.

akshayjshah: Yep, sounds better.

PR #118, opened 2017-06-29
Add typealiases for Invoke and Provide parameters.

glibsm:
This is going to allow us to better document these concepts with
examples. I think the biggest benefit of these is going to come from
migrating `fx` to use these aliases in it's Provide and Invoke functions
meaning the developers would be able to access the docs much easier.
Right now we just say "go find dig docs".

Bikeshed away on the `Extractor` name. I was trying to think through
signaling that it's the flip side of a `Constructor`, initially having it named
`dig.Function`, but that maked godocs very awkward: `dig.Function is a function that...`
reads very weird.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -55,6 +55,20 @@ type Container struct {
 	// index map[reflect.Type]key
 }
 
+// Constructor is a function that takes any number of parameters (dependencies)
+// which will be supplied by the Container on demand, and returns any number
+// of objects that will be inserted into the dependency graph.
+//
+// Constructor must return at least one non-error value, all of which are
+// then available in the Container.
+type Constructor interface{}
+
+// Extractor is a function that takes any number of arguments which supplied
+// from the underlying Container. This action will result in the resolution
+// of the dependency graph for the parameters, calling their constructors,
+// and recursing into the dependencies.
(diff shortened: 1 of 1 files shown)


akshayjshah:
I'd really prefer not to do this; I don't see much value in typedefs that don't add *behavior*, especially when they're only used in one place.

Put differently, I'd prefer to see excellent docs for the `Provide` method instead of splitting the docs between the `Provide` method and the `Constructor` type.

glibsm:
The idea for these typealiases is that Fx can now change it's signatures to use these dig types. Yes, within the confines of this dig repository it doesn't make too much sense, but in the context of `fx.Provide`, for instance, it would make a great deal of sense to be able to jump to the `dig.Constructor` parameter type and check out the docs.

Given that both of your arguments hinge on the *one place*, do you still feel like it's one place if signatures of Fx can benefit from it?

akshayjshah:
I think I see what you're saying, but I'm still not in favor. The behavior being documented isn't a property of the type, it's a property of the provide and invoke methods.

glibsm:
I still think there typealiases would be a huge win in documenting the *fx* side of things, especially in the method declarations there.

PR #119, opened 2017-06-30
Add variadic options parameter to public APIs

### LINKED ISSUE #117: New needs to take varargs Options
Before we get further into an RC series, `New` needs to take a variadic number of `Option`s, even if we don't yet have any concrete `Option` implementations. Adding this after 1.0 is a breaking change.

akshayjshah:
We're fairly confident that the core APIs meet our current needs, but we
should leave space for future expansion. To avoid breaking changes,
let's introduce functional options even though we don't currently have
any implementations.

Fixes #117.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -42,16 +42,20 @@ type key struct {
 // Option configures a Container. It's included for future functionality;
 // currently, there are no concrete implementations.
 type Option interface {
-	apply(*Container)
+	unimplemented()
 }
 
 // A ProvideOption modifies the default behavior of Provide. It's included for
 // future functionality; currently, there are no concrete implementations.
-type ProvideOption interface{}
+type ProvideOption interface {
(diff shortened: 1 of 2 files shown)


dig.go
```diff
@@ -39,6 +39,20 @@ type key struct {
...
 
+// Option configures a Container. It's included for future functionality;
+// currently, there are no concrete implementations.
+type Option interface {
+	apply(*Container)
+}
+
+// A ProvideOption modifies the default behavior of Provide. It's included for
+// future functionality; currently, there are no concrete implementations.
+type ProvideOption interface{}
```
alsamylkin: should we have an actual private interface? Type aliasing to interface{} widens the DIG API too much :)

akshayjshah: Makes sense.

Issue #120, opened 2017-07-06
Treat varargs as optional?


akshayjshah:
Typically, variadic arguments are optional - the function needs to handle `len(varargs)==0` already. Making dig automatically treat varargs as optional would make a wider variety of plain-Go functions work out of the box.

abhinav:
This needs further discussion.

Currently, constructors that have variadic arguments cannot be called from
dig. For `NewFoo(...Option) *Foo`, you cannot provide `[]Option` or `Option`.

So before we make variadic arguments optional, we have to figure out how, if
at all, users will be able to `Provide` them to constructors.

After discussions with @alsamylkin and @akshayjshah, there are two solutions
to this:

-   Disallow providing variadic arguments to constructors. A function
    `NewFoo(*Bar, ...Option) *Foo` may be `Provide`d into the container but we
    will supply no method of filling in that `...Option`. The constructor will
    be called with just the `*Bar`. We are categorizing supplying variadic
    arguments as a complex use case, the kind for which we expect people to
    use `dig.In`, `dig.Out`, or wrapper functions. An advantage of this
    solution is that we can add support for providing variadic arguments later
    if a better solution becomes possible.
    
    @alsamylkin suggested that under this solution, if necessary, `fx` could
    provide a function like `fx.Expand` that lets users explicitly opt into
    vararg-slice conversion. As in,
    
    ```go
    fx.Provide(fx.Expand(NewFoo))
    ```
    
    Is equivalent to,
    
    ```go
    fx.Provide(func(b *Bar, opts []Option) *Foo { return NewFoo(b, opts...) })
    ```

-   Allow providing variadic arguments as slices by default (optional as per
    this issue). A function `NewFoo(*Bar, ...Option)` may be `Provide`d.
    Additionally, a `[]Option` may optionally be provided. The function will
    be called with that slice expanded as if with `opts...`. Arguably, this is
    how a user would expect to provide values for variadic arguments. Another
    advantage is that most existing constructors users write will Just Work
    without any wrapping.

---

CC @alsamylkin @akshayjshah @breerly @glibsm

akshayjshah:
I'm fine with the first option above - treating all variadic arguments as unmet optional dependencies - in the short term. If we can't come to agreement quickly, let's land that change and continue discussion.

[...]

PR #122, opened 2017-07-10
Support constructors with variadic arguments

abhinav:
This adds support for functions with a variadic number of arguments to
be Provided and Invoked. We will declare a dependency on a slice of the
variadic type when a variadic function is encountered.

In a follow up PR, we will make this dependency optional rather than required.

See: #120

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -352,13 +352,24 @@ func (c *Container) remove(nodes []*node) {
 }
 
 func (c *Container) constructorArgs(ctype reflect.Type) ([]reflect.Value, error) {
+	numArgs := ctype.NumIn()
+
 	args := make([]reflect.Value, 0, ctype.NumIn())
-	for i := 0; i < ctype.NumIn(); i++ {
+	for i := 0; i < numArgs; i++ {
 		arg, err := c.get(edge{key: key{t: ctype.In(i)}})
 		if err != nil {
 			return nil, fmt.Errorf("couldn't get arguments for constructor %v: %v", ctype, err)
 		}
-		args = append(args, arg)
+
+		// If the function is variadic, the last argument is a slice that must
+		// be expanded into the argument list.
+		if i == numArgs-1 && ctype.IsVariadic() {
+			for j := 0; j < arg.Len(); j++ {
+				args = append(args, arg.Index(j))
+			}
+		} else {
+			args = append(args, arg)
+		}
 	}
 	return args, nil
 }
--- a/dig_test.go
+++ b/dig_test.go
@@ -700,6 +700,49 @@ func TestEndToEndSuccess(t *testing.T) {
 			assert.Nil(t, p.Baz, "Baz must be unset")
 		}), "invoke failed")
 	})
+
+	t.Run("variadic arguments invoke", func(t *testing.T) {
+		c := New()
+
+		type A struct{ i int }
+
+		require.NoError(t, c.Provide(func() []*A {
+			return []*A{{1}, {2}, {3}}
+		}), "failed to provide A slice")
+
+		require.NoError(t, c.Invoke(func(as ...*A) {
+			require.Len(t, as, 3, "length must match")
+			assert.Equal(t, &A{1}, as[0], "as[0] must match")
+			assert.Equal(t, &A{2}, as[1], "as[1] must match")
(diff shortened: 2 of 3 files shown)


dig.go
```diff
@@ -352,13 +352,24 @@ func (c *Container) remove(nodes []*node) {
...
 		arg, err := c.get(edge{key: key{t: ctype.In(i)}})
 		if err != nil {
 			return nil, fmt.Errorf("couldn't get arguments for constructor %v: %v", ctype, err)
 		}
-		args = append(args, arg)
+
+		// If the function is variadic, the last argument is a slice that must
+		// be expanded into the argument list.
+		if i == numArgs-1 && ctype.IsVariadic() {
+			for j := 0; j < arg.Len(); j++ {
```
alsamylkin: Can we use variadic functions as constructors?
If yes, then, the size should be limited here by 0 or 1. Otherwise we should error, because we are going to copy the same variable(this is how I think about variadic arguments, not as slices)

dig_test.go
```diff
@@ -700,6 +700,49 @@ func TestEndToEndSuccess(t *testing.T) {
...
+	t.Run("variadic arguments invoke", func(t *testing.T) {
+		c := New()
+
+		type A struct{ i int }
+
+		require.NoError(t, c.Provide(func() []*A {
+			return []*A{{1}, {2}, {3}}
+		}), "failed to provide A slice")
+
+		require.NoError(t, c.Invoke(func(as ...*A) {
```
alsamylkin: variadic types are different from slices and I don't think that we should mix them.
Consider this case: type is missing from the graph and we invoke a variadic function that expands this type. There should be no error, right? I expect this test to pass:
```golang
t.Run("alsam", func(t *testing.T) {
	c := New()
	require.NoError(t, c.Invoke(func(n ...int) {
	assert.Empty(t, n)
}), "failed to invoke")
```

akshayjshah: This should be addressed by a PR to fix #120.

dig.go
akshayjshah: Alex, can you clarify this comment? I'm not understanding why we'd want to limit the number of variadic args to 0 or 1, since that greatly reduces the value of supporting varargs.

Varargs are slices in Go, so this seems like the least surprising behavior to me.

abhinav:
@alsamylkin The purpose of this PR is to make it possible to provide
constructors with variadic arguments. The arguments will be made optional in a
follow-up PR.

As for your concern about slices, I don't see many solutions to this. If I
provide `func(...Option) *Type` and I *really* wanted to provide the options,
I would expect to provide `func() []Option`, not a single `func() Option`.
Requiring that you provide only one value for a variadic constructor seems
broken to me.

I think we can do two sensible things for variadic arguments:

- Allow providing them as slices.
- Make it impossible to provide them. Constructors with variadic arguments
  will be called with none.
  
After thinking about this a bit more, I'm leaning towards the second solution.
It's a bit limiting but it lets us figure out a better strategy in the future
without breaking anything.

dig.go
alsamylkin: I think there is a distinction between them: [varargs are passed as slices](https://golang.org/ref/spec#Passing_arguments_to_..._parameters), but they are not slices. If someone wants to pass slice elements as args, they need to expand it by adding `...` after slice.
```golang
package main

func print(nums ...int) {
	for _, n := range nums {
		println(n)
	}
}

func main() {
	print([]int{3,2,1}...)
	println("now print argument without slices:")
	print(7)
	print(6,5,4)
}
```

Which leads to a following case: if we have variable x and y (`var T x` and `var []T y`) in the graph, how should we handle invoke of of some `f(v ...T)`? I argue that we should use `x`, not `y`, because in your program you can write `f(x)`. `f(y)` will lead to compilation error, because you need **explicitly** expand y: `f(y...)`.

My suggestion is based on the above argument and I see varargs as optionals in the end of argument list for a function, not as slice from the container. Dig allows us to store only one instance per type, that's why I think the length could be only 0 or 1. I don't think that we can pass named instances as slices, because names carry important information about instances and can't be blindly applied.

alsamylkin:
@abhinav The second solution sounds good to me too!

abhinav:
More discussion needed. Moving conversation to https://github.com/uber-go/dig/issues/120#issuecomment-314256075

PR #123, opened 2017-07-11
Call variadic constructors without arguments

abhinav:
Dig constructors may now be variadic functions. These functions will be
called with their dependencies as usual and no arguments will be passed
for the variadic arguments.

So a constructor `New(*Foo, *Bar, ...Option) *Baz` will be treated as
`New(*Foo, *Bar) *Baz`.

See #120 for more discussion.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -352,16 +352,10 @@ func (c *Container) remove(nodes []*node) {
 }
 
 func (c *Container) constructorArgs(ctype reflect.Type) ([]reflect.Value, error) {
-	numArgs := ctype.NumIn()
-	if ctype.IsVariadic() {
-		// NOTE: If the function is variadic, we skip the last argument
-		// because we're not filling variadic arguments yet. See #120.
-		numArgs--
-	}
-
-	args := make([]reflect.Value, 0, numArgs)
-	for i := 0; i < numArgs; i++ {
-		arg, err := c.get(edge{key: key{t: ctype.In(i)}})
+	argTypes := getConstructorArgTypes(ctype)
+	args := make([]reflect.Value, 0, len(argTypes))
+	for _, t := range argTypes {
+		arg, err := c.get(edge{key: key{t: t}})
 		if err != nil {
 			return nil, fmt.Errorf("couldn't get arguments for constructor %v: %v", ctype, err)
 		}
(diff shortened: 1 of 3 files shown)


dig.go
```diff
@@ -389,8 +396,15 @@ func newNode(k key, ctor interface{}, ctype reflect.Type) (*node, error) {
 
 // Retrieves the dependencies for a constructor
 func getConstructorDependencies(ctype reflect.Type) ([]edge, error) {
+	numArgs := ctype.NumIn()
```
glibsm: lets extract into a function since the code and comment appears twice

dig_test.go
```diff
@@ -700,6 +700,59 @@ func TestEndToEndSuccess(t *testing.T) {
...
+			return gaveA
+		}), "failed to provide A")
+
+		require.NoError(t, c.Provide(func() []*A {
+			panic("[]*A constructor must not be called.")
+		}), "failed to provide A slice")
+
+		require.NoError(t, c.Invoke(func(a *A, as ...*A) {
+			require.NotNil(t, a, "A must not be nil")
+			require.True(t, a == gaveA, "A must match")
```
alsamylkin: Curious, why not `require.Equal`?

dig_test.go
```diff
@@ -700,6 +700,59 @@ func TestEndToEndSuccess(t *testing.T) {
...
+		type A struct{}
+
+		var gaveA *A
+		require.NoError(t, c.Provide(func() *A {
+			gaveA = &A{}
+			return gaveA
+		}), "failed to provide A")
+
+		require.NoError(t, c.Provide(func() []*A {
+			panic("[]*A constructor must not be called.")
```
alsamylkin: `require.FailNow()`?

abhinav: I was sort of trying to stay consistent with existing tests that use panic.

Plus with `require.FailNow`, you now have to have an empty return statement to satisfy the compiler.

I can change this if you feel strongly about this.

dig_test.go
abhinav: require.Equal on pointers dereferences and compares value. Because this is an empty type (`struct{}`), that will always be true. `==` on pointers compares the pointer value to verify that it's the same instance of the object.

dig_test.go
alsamylkin: Not really, panic is fine too.

PR #124, opened 2017-07-18
Test that invoke returns are not inserted

glibsm:
This PR should also bring the dig coverage to 100%.


glibsm:
@breerly we went through some cycles with this. 

At first, no returns were allowed. Then, an optional error was allowed. Then, we allowed *anything* and error, and anything would be inserted into the graph. Then we arrived at only optional error and ignoring other returns to allow for more functions to be invoked.

If we have a change of heart and only require an `error?`, then this test will be modified as we see fit.

breerly:
Ok - I'm fine with allowing anything but ignoring! Makes sense - allow more to be executed.

PR #126, opened 2017-07-19
Export isInObject and isOutObject as IsIn and IsOut

glibsm:
This allows other libraries to authoritatively check if a type is
dig.In/Out without duplicating the logic (and potentially having it
drift apart in the future releases).

https://github.com/uber-go/fx/issues/540

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
   variadic options parameter to all public APIs.
 - Added support for functions with variadic arguments. These functions will be
   called without supplying their variadic arguments.
-- Export `IsIn` and `IsOut` package level functions.
+- Exported `IsIn` and `IsOut` functions to check if types are `dig.In` or
+  `dig.Out` compatible.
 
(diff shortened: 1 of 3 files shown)


types.go
```diff
@@ -66,11 +66,15 @@ func isError(t reflect.Type) bool {
 	return t.Implements(_errType)
 }
 
-func isInObject(t reflect.Type) bool {
+// IsIn returns true if passed in type embeds dig.In either directly
+// or through another struct field.
```
abhinav: or through another embedded field.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
   variadic options parameter to all public APIs.
 - Added support for functions with variadic arguments. These functions will be
   called without supplying their variadic arguments.
+- Export `IsIn` and `IsOut` package level functions.
```
abhinav: Use same voice as other entries

> Added functions to check if types are `dig.In` or `dig.Out` compatible.

glibsm:
Yeah the nits make sense, will change.

PR #130, opened 2017-07-20
Change exported IsIn/IsOut to use interface{}

glibsm:
This makes external experience a lot better to check when something is
and isn't an In/Out type.

Use the previous iteration of the function under the hood.

Changelog not needed, since this hasn't been released.

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -956,15 +956,22 @@ func TestTypeCheckingEquality(t *testing.T) {
 		A
 	}
 	type in struct {
+		In
 		A
 	}
 	type out struct {
 		B
 	}
-	i := in{}
-	o := out{}
-	require.Equal(t, IsIn(i), isInType(reflect.TypeOf(i)))
-	require.Equal(t, IsOut(o), isOutType(reflect.TypeOf(o)))
+	tt := []interface{}{
+		in{},
+		out{},
+		A{},
+		B{},
+	}
+	for _, s := range tt {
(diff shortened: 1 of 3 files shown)


abhinav:
I'm okay with the API change. We should do two things differently,

-   Handle `nil`s. A nil value means we are dealing with a non-struct type
    which can't be a dig.In anyway (yet; see #131).
-   Attempt to cast the `interface{}` to `reflect.Type` and if it casts, use
    it directly. This lets us still use reflection APIs and avoid hard-to-
    debug bugs from loss of type safety.

types.go
```diff
@@ -68,18 +68,36 @@ func isError(t reflect.Type) bool {
...
+	}
+
+	var t reflect.Type
+
+	// maybe it's already reflect.Type
+	if mt, ok := i.(reflect.Type); ok {
+		t = mt
+	} else {
+		// if not just take the type
+		t = reflect.TypeOf(i)
```
abhinav:
```
if t == nil {
  return false
}
```

And a test for it

dig_test.go
```diff
@@ -949,6 +949,35 @@ func TestProvideFuncsWithoutReturnsFails(t *testing.T) {
...
+	}
+	tests := []struct {
+		item  interface{}
+		isIn  bool
+		isOut bool
+	}{
+		{in{}, true, false},
+		{out{}, false, true},
+		{A{}, false, false},
+		{B{}, false, true},
```
abhinav: Add

```
{nil, false, false}
```

PR #135, opened 2017-07-26
Basic tests for Go 1.9 aliases

### LINKED ISSUE #134: Validate behavior with Go 1.9 aliases
Didn't think of this earlier, but we *absolutely* need to nail down what dig does with the aliases introduced in Go 1.9 - they should be treated as the same type (only one can be in the container, &c).

Let's at least explore this before we cut a 1.0 of fx - Go 1.9 is now in RC1 and will land in August, and we don't want a ton of things break immediately post-launch.

abhinav:
This verifies the behavior of Dig with Go 1.9 aliases.

I had to change the Travis setup to run lint only for Go 1.9 because it fails
syntax check for other versions.

Resolves #134

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,19 @@
 language: go
 sudo: false
-go:
-  - 1.7
-  - 1.8
-  - tip
 go_import_path: go.uber.org/dig
+
+matrix:
+  include:
+    - go: 1.7
+    - go: 1.8
+    - go: tip
+      env: LINT=1
+
 cache:
   directories:
     - vendor
 install:
   - make dependencies
 script:
-  - make lint
+  - test "$LINT" -eq 1 && make lint
   - make ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,5 +15,5 @@ cache:
 install:
   - make dependencies
 script:
-  - test "$LINT" -eq 1 && make lint
+  - test "$LINT" -eq 1 && make lint || echo "Skipping lint"
   - make ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ matrix:
   include:
     - go: 1.7
     - go: 1.8
+    # TODO: Switch to Go 1.9 when released
     - go: tip
       env: LINT=1
 
--- a/dig_go19_test.go
+++ b/dig_go19_test.go
@@ -48,6 +48,20 @@ func TestEndToEndSuccessWithAliases(t *testing.T) {
 		}), "invoke failed")
 	})
 
+	t.Run("duplicate provide", func(t *testing.T) {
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -1,15 +1,19 @@
...
+      env: LINT=1
+
 cache:
   directories:
     - vendor
 install:
   - make dependencies
 script:
-  - make lint
+  - test "$LINT" -eq 1 && make lint || echo "Skipping lint"
```
glibsm: why do we lint on tip all of the sudden, as opposed to the latest release

glibsm:
Could you add some error cases?

```
type A struct {}
type B = A
require.NoError provide A
require.Error duplicate provide B
```

.travis.yml
abhinav: From description:

> I had to change the Travis setup to run lint only for Go 1.9 because it fails
syntax check for other versions.

glibsm: Am I missing something? :(

It looks like we're linting on tip not 1.9 according to the matrix

abhinav: Go 1.9 is unreleased. tip == go 1.9

glibsm: Ugh, I can't read.

Ok, lets remember to swap it out for 1.9 when it's released.

abhinav: Added TODO

dig_go19_test.go
```diff
@@ -0,0 +1,114 @@
...
+		type B = A
+
+		c := New()
+		require.NoError(t, c.Provide(func() A { 
+			return A{}
+		}), "A should not fail to provide")
+
+		require.Error(t, c.Provide(func() B {
+			return B{}
+		}), "B should fail to provide")
```
glibsm: The interesting point that raises for me: is the error message going to state that B is already in the graph, or A already in the graph?

I'd maybe `assert.Contains(t, err, "blah blah already in the graph")`

dig_go19_test.go
```diff
@@ -0,0 +1,114 @@
...
+			return b
+		}), "provide failed")
+
+		require.NoError(t, c.Invoke(func(got Buffer) {
+			require.NotNil(t, got, "invoke got nil buffer")
+			require.True(t, got == b, "invoke got wrong buffer")
+		}), "invoke failed")
+	})
+
+	t.Run("duplicate provide", func(t *testing.T) {
```
glibsm: `t.Run("glib test")`

dig_go19_test.go
abhinav: Since the aliases aren't retained at runtime, I'm gonna guess it will say "A". I'll add a test.

akshayjshah:
Awesome - thanks!

Issue #131, opened 2017-07-20
Can't embed pointers to dig.In/Out


glibsm:
The following e2e test doesn't work:

```go
	t.Run("embedding out pointers works", func(t *testing.T) {
		c := New()
		type A struct{}
		type B struct {
			Out
			A
		}
		type C struct {
			*B // pointer to B which is a dig.In with A
		}
		require.NoError(t, c.Provide(func() C { return C{} }))
		require.NoError(t, c.Invoke(func(A) {}))
	})
```

It works if type C is written to embed type B directly, rather than the pointer to B:

```go
type C struct {
  B // direct embed of type B
}
```

Question is... do we care? `embedsType` function seems to solely focus on structs:

```go
if t.Kind() != reflect.Struct {
  continue
}
```

Should be easy to modify to look at `Elems()` of a pointer instead.

abhinav:
The original implementation tried to stay as narrow as possible in the range
of types it allowed to be `dig.In` structs.

I don't feel strongly about whether we allow pointers except it may introduce
corner cases we haven't accounted for.  We can always add it post 1.0 if
someone has a valid use case for it.

glibsm:
The main corner case that this is going to add is ability to be stuck in a cycle while trying to detect when something is or isn't an In/Out type.

Because it's impossible to declare a cyclical struct right now it's not a problem. However the following code would recurse indefinitely:

```go
type A struct {
  dig.Out
  *A
}
```

I don't think it's worth to spend cycles on this now, but I'm going to leave this as an open issue for a bit until we decide if it's something we want to tackle for 1.0.

glibsm:
The main issue for me is that it's a *silent* failure. In the aforementioned test, it will register `*B` into the graph, which can be really confusing.

We may want to consider either forbidding it outright (to prevent massive cycles), or at least error out.

abhinav:
I didn't realize it was a silent failure. Yeah, we should definitely error out in this cas.e

bayesianmind:
+1 to either supporting this or erroring out. 
We can also improve our messaging for the dig.In case:
Today if have this:
```
type Z  struct{
    dig.In
    SomeType A
}

func Invoke(b *B){}
```
 Invoke will fail saying that *B isn't in the graph. For someone not familar with Dig this seems like dig.In is broken. I'd like to have a better error message directing users to use `B` instead of `*B`.

PR #137, opened 2017-07-27
Error on private fields inside dig.In/Out

### LINKED ISSUE #136: Error when dig.In/Out structs contains un-exported fields
Moving to appropriate repo from https://github.com/uber-go/fx/issues/565

glibsm:
As discussed in the issue, we no longer accept private fields on these structs as valid input. I tried to make the error message informative, pointing out the exact field and suggesting an export.

Original issue from Fx repo https://github.com/uber-go/fx/issues/565

Fixes #136

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## v1.0.0-rc3 (unreleased)
 
-- `Provide` and `Invoke` will now error out if `dig.In` or `dig.Out` structs
-  contain private fields. Previously these fields were ignored which often led
-  to confusion.
+- `Provide` and `Invoke` will now fail if `dig.In` or `dig.Out` structs
+  contain unexported fields. Previously these fields were ignored which often
+  led to confusion.
 
 ## v1.0.0-rc2 (21 Jul 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## v1.0.0-rc3 (unreleased)
 
-- No changes yet.
+- `Provide` and `Invoke` will now error out if `dig.In` or `dig.Out` structs
+  contain private fields. Previously these fields were ignored which often led
+  to confusion.
 
 ## v1.0.0-rc2 (21 Jul 2017)
 
--- a/dig.go
+++ b/dig.go
@@ -300,8 +300,16 @@ func (c *Container) createInObject(t reflect.Type) (reflect.Value, error) {
 	dest := reflect.New(t).Elem()
 	for i := 0; i < t.NumField(); i++ {
 		f := t.Field(i)
+
+		if f.Type == _inType {
+			// skip over the dig.In embed itself
+			continue
+		}
+
 		if f.PkgPath != "" {
-			continue // skip private fields
+			return dest, fmt.Errorf(
+				"private fields not allowed in dig.In, did you mean to export %q?",
+				fmt.Sprintf("%v %v", f.Name, f.Type))
 		}
 
 		isOptional, err := isFieldOptional(t, f)
--- a/dig.go
+++ b/dig.go
@@ -222,8 +222,15 @@ func traverseOutTypes(k key, f func(key) error) error {
 		field := k.t.Field(i)
 		ft := field.Type
 
+		if field.Type == _outType {
+			// do not recurse into dig.Out itself, it will contain digSentinel only
+			continue
+		}
+
 		if field.PkgPath != "" {
-			continue // skip private fields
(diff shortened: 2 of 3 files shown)


dig.go
```diff
@@ -222,8 +222,15 @@ func traverseOutTypes(k key, f func(key) error) error {
...
+		if field.Type == _outType {
+			// do not recurse into dig.Out itself, it will contain digSentinel only
+			continue
+		}
+
 		if field.PkgPath != "" {
-			continue // skip private fields
+			return fmt.Errorf(
+				"private fields not allowed in dig.Out, did you mean to export %q?",
+				fmt.Sprintf("%v %v", field.Name, field.Type))
```
abhinav: Sprintf inside an Errorf?

```
"every field of dig.Out structs must be exported: field %q (%v) of %v is not exported", field.Name, field.Type, k.t
```

dig.go
```diff
@@ -293,8 +300,16 @@ func (c *Container) createInObject(t reflect.Type) (reflect.Value, error) {
...
+		if f.Type == _inType {
+			// skip over the dig.In embed itself
+			continue
+		}
+
 		if f.PkgPath != "" {
-			continue // skip private fields
+			return dest, fmt.Errorf(
+				"private fields not allowed in dig.In, did you mean to export %q?",
+				fmt.Sprintf("%v %v", f.Name, f.Type))
```
abhinav: same as above

CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## v1.0.0-rc3 (unreleased)
 
-- No changes yet.
+- `Provide` and `Invoke` will now error out if `dig.In` or `dig.Out` structs
+  contain private fields. Previously these fields were ignored which often led
```
abhinav: Go docs usually refer to these as exported vs not exported. There's no mention of private or unexported on https://golang.org/ref/spec. Rephrase to,

> `Provide` and `Invoke` will now fail if `dig.In` or `dig.Out` structs contain fields that are not exported. Previously [..]

dig.go
glibsm: I wanted to quote over the name+type so that was the only way to do it.
I'm ok with the suggested format as a compromise

dig_test.go
```diff
@@ -1293,4 +1303,41 @@ func TestInvokeFailures(t *testing.T) {
...
+		require.NoError(t, c.Provide(func() A { return A{} }))
+
+		err := c.Invoke(func(i in) { assert.Fail(t, "should never get in here") })
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "private fields not allowed in dig.In")
+		assert.Contains(t, err.Error(), `"a2" (dig.A)`)
+		assert.Contains(t, err.Error(), "did you mean to export")
+	})
+
+	t.Run("embedded private member gets an error", func(t *testing.T) {
```
alsamylkin: what about private embedding as a pointer and value ?
```golang
type a struct {
  Name string
}

type b struct {
  (*)a
}
```
?

CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## v1.0.0-rc3 (unreleased)
 
-- No changes yet.
+- `Provide` and `Invoke` will now fail if `dig.In` or `dig.Out` structs
```
alsamylkin: nit: shouldn't it be a breaking change?

glibsm: I think breaking refers to _API_ changes, not behavior.

dig_test.go
glibsm: We don't do pointer embeds. See https://github.com/uber-go/dig/issues/131

CHANGELOG.md
glibsm: Also it's _maybe breaking_ at best, if that's a thing. It only affects customers who have been misusing this in the first place and not getting those values in and out of the conatienr.

bayesianmind:
This is great. I am deeply aware that dig.In structs need to have all exported members, and have been bitten by it multiple times, but I just now did it AGAIN and spent a bunch of time debugging why my members weren't being set. When can we get this in FX?

glibsm:
@bayesianmind Not sure what you are asking about exactly. This will automatically be picked up by the next Fx release.

If you're really itching to get on the bleeding edge of this, you should be able to lock your dig version to master, or this sha, to use it with RC1 Fx. At this point it might be worthwhile to wait, though. Up to you.

bayesianmind:
Got it, thanks!

breerly:
@bayesianmind fx master is now pinned to dig master - that might make life easier :)

PR #139, opened 2017-07-28
Increased error checking for embedding In/Out pointers

### LINKED ISSUE #131: Can't embed pointers to dig.In/Out

glibsm:
Well this sure is messy, but there is massive complexity in just how many cases there are:

- Embedding pointers to In/Out
- Using pointers as params/returns
- Embedding pointers to objects that embed In/Out
- Embedding pointers to objects that embed pointers to In/Out
- ????

I wrote some e2e tests that in general I expect to fail and pass and reversed it from there. This obviously adds some tech debt that should be cleaned up.

Fixes #131

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -1101,7 +1101,6 @@ func TestProvideFailures(t *testing.T) {
 		}
 
 		err := c.Provide(func() out { return out{String: "foo"} })
-		fmt.Println(c)
 		require.Error(t, err)
 		assert.Contains(t, err.Error(), "can't embed *dig.Out pointers")
 	})
--- a/dig_test.go
+++ b/dig_test.go
@@ -1078,6 +1078,33 @@ func TestProvideFailures(t *testing.T) {
 		assert.Contains(t, err.Error(), `"a2" (dig.A)`)
 		assert.Contains(t, err.Error(), "did you mean to export")
 	})
+
+	t.Run("providing pointer to out should fail", func(t *testing.T) {
+		c := New()
+		type out struct {
+			Out
+
+			String string
+		}
+		err := c.Provide(func() *out { return &out{String: "foo"} })
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "dig.out is a pointer to dig.Out")
+	})
+
+	t.Run("embedding pointer to out should fail", func(t *testing.T) {
+		c := New()
+
+		type out struct {
+			*Out
+
+			String string
+		}
+
+		err := c.Provide(func() out { return out{String: "foo"} })
+		fmt.Println(c)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "can't embed *dig.Out pointers")
+	})
 }
 
 func TestInvokeFailures(t *testing.T) {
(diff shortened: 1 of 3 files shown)


dig_test.go
```diff
@@ -883,17 +883,14 @@ func TestCantProvideParameterObjects(t *testing.T) {
...
-		c := New()
-		require.NoError(t, c.Provide(func() (*Args, error) {
-			return args, nil
-		}), "provide failed")
-		require.NoError(t, c.Invoke(func(a *Args) {
-			require.True(t, args == a, "args must match")
-		}), "invoke failed")
+		err := c.Provide(func() (*Args, error) { return args, nil })
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "can't provide pointers to parameter objects")
```
akshayjshah: This test has somehow gotten weird - why are we writing test cases that *return* `fx.In`? I think that the intention was to test that we can write constructors that return pointers. If that's not covered elsewhere, let's add a test with a constructor that returns `*byte.Buffer`.

Separately, we should add whatever test we need for pointers to In/Out structs.

dig_test.go
```diff
@@ -1081,6 +1078,33 @@ func TestProvideFailures(t *testing.T) {
...
+		c := New()
+
+		type out struct {
+			*Out
+
+			String string
+		}
+
+		err := c.Provide(func() out { return out{String: "foo"} })
+		fmt.Println(c)
```
akshayjshah: Stray `println`.

types.go
```diff
@@ -26,10 +26,12 @@ import (
...
-	_noValue reflect.Value
-	_errType = reflect.TypeOf((*error)(nil)).Elem()
-	_inType  = reflect.TypeOf((*In)(nil)).Elem()
-	_outType = reflect.TypeOf((*Out)(nil)).Elem()
+	_noValue    reflect.Value
+	_errType    = reflect.TypeOf((*error)(nil)).Elem()
+	_inPtrType  = reflect.TypeOf((*In)(nil))
+	_inType     = _inPtrType.Elem()
+	_outPtrType = reflect.TypeOf((*Out)(nil))
+	_outType    = _outPtrType.Elem()
```
akshayjshah: Perhaps clearer as `reflect.TypeOf(Out{})`?

breerly:
You mention this takes on some debt - can you create `// TODO` comments everywhere you feel some ickiness has been introduced?

dig_test.go
glibsm: There are plenty of tests that cover the pointers, like `*byte.Buffer`. This tries to cover the mis-use case when `In` is used instead of `Out`.

I just modified the previous test.

PR #142, opened 2017-08-07
Rewrite godocs

abhinav:
This rewrites most of the godocs of this library and documents
undocumented features (dig.In, dig.Out, etc.).

The package-level docs give a general overview of the library and
showcase different dig.In/dig.Out features by use case.

Entity-level docs for dig.In and dig.Out act as the authoritative
reference list of all tags supported by those structs and how they
behave.

---

Here's a rendering of the docs in PDF form: [dig - The Go Programming Language.pdf](https://github.com/uber-go/dig/files/1205574/dig.-.The.Go.Programming.Language.pdf)

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -57,12 +57,11 @@ type InvokeOption interface {
 	unimplemented()
 }
 
-// A Container is a directed, acyclic graph of dependencies. Dependencies are
-// constructed on-demand and returned from a cache thereafter, so they're
-// effectively singletons.
+// Container is a directed acyclic graph of types and their dependencies.
 type Container struct {
 	nodes map[key]*node
 	cache map[key]reflect.Value
+
 	// TODO: for advanced use-case, add an index
 	// This will allow retrieval of a single type, without specifying the exact
 	// tag, provided there is only one object of that given type
@@ -73,7 +72,7 @@ type Container struct {
 	// index map[reflect.Type]key
 }
 
-// New constructs a ready-to-use Container.
+// New constructs a Container.
 func New(opts ...Option) *Container {
 	return &Container{
 		nodes: make(map[key]*node),
@@ -81,18 +80,20 @@ func New(opts ...Option) *Container {
 	}
 }
 
-// Provide teaches the Container how to construct one or more new types.
+// Provide teaches the container how to build values of one or more types.
 //
-// Any function passed to Provide is assumed to be a constructor. Constructors
-// can take any number of parameters, which will be supplied by the Container
-// on demand. They must return at least one non-error value, all of which are
-// then available in the Container. If the last returned value is an error, the
-// Container inspects it to determine whether the constructor succeeded or
-// failed. Regardless of position, returned errors are never put into the
-// Container's dependency graph.
+// The first argument of Provide is a function which accepts zero or more
+// parameters and returns one or more results. The function may optionally
+// return an error to indicate that it failed to build the value. This
+// function will be treated as the constructor for all the types it returns.
+// This function will be called AT MOST ONCE when a type produced by it, or a
+// type which depends on this function's output, is requested via Invoke. If
+// the same types are requested multiple times, the previously produced value
+// will be reused.
 //
-// All non-functions (including structs, pointers, Go's built-in collections,
-// and primitive types like ints) are inserted into the Container as-is.
+// In addition to accepting vanilla constructors, Provide also accepts
+// constructors which accept dig.In structs as dependencies and/or return
+// dig.Out results.
 func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) error {
 	ctype := reflect.TypeOf(constructor)
 	if ctype == nil {
@@ -107,12 +108,12 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 	return nil
 }
 
-// Invoke runs a function, supplying its arguments from the Container. If the
-// function's last return value is an error, that error is propagated to the
-// caller. All other returned values (if any) are ignored.
+// Invoke runs the given function immediately, supplying its arguments from
+// the container. All types requested by the function will be instantiated,
+// and their dependencies will be instantiated in the process.
 //
-// Passing anything other than a function to Invoke returns an error
-// immediately.
+// The provided function may return an error to indicate failure. The error
+// will be returned to the caller as-is.
 func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
 	ftype := reflect.TypeOf(function)
 	if ftype == nil {
--- a/dig.go
+++ b/dig.go
@@ -80,7 +80,8 @@ func New(opts ...Option) *Container {
 	}
 }
 
-// Provide teaches the container how to build values of one or more types.
+// Provide teaches the container how to build values of one or more types and
+// expresses their dependency types.
 //
 // The first argument of Provide is a function which accepts zero or more
 // parameters and returns one or more results. The function may optionally
@@ -108,12 +109,13 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 	return nil
 }
 
-// Invoke runs the given function immediately, supplying its arguments from
-// the container. All types requested by the function will be instantiated,
-// and their dependencies will be instantiated in the process.
+// Invoke runs the given function after instantiating its dependencies.
+// Any arguments that the function has are treated as its dependencies and
+// they are instantiated in an unspecified order along with any dependencies
+// that they might have.
 //
-// The provided function may return an error to indicate failure. The error
-// will be returned to the caller as-is.
+// The function may return an error to indicate failure. The error will be
+// returned to the caller as-is.
 func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
 	ftype := reflect.TypeOf(function)
 	if ftype == nil {
--- a/dig.go
+++ b/dig.go
@@ -83,17 +83,17 @@ func New(opts ...Option) *Container {
 // Provide teaches the container how to build values of one or more types and
 // expresses their dependency types.
 //
-// The first argument of Provide is a function which accepts zero or more
+// The first argument of Provide is a function that accepts zero or more
 // parameters and returns one or more results. The function may optionally
 // return an error to indicate that it failed to build the value. This
 // function will be treated as the constructor for all the types it returns.
 // This function will be called AT MOST ONCE when a type produced by it, or a
-// type which depends on this function's output, is requested via Invoke. If
-// the same types are requested multiple times, the previously produced value
-// will be reused.
+// type that consumes this function's output, is requested via Invoke. If the
+// same types are requested multiple times, the previously produced value will
+// be reused.
 //
 // In addition to accepting vanilla constructors, Provide also accepts
-// constructors which accept dig.In structs as dependencies and/or return
+// constructors that accept dig.In structs as dependencies and/or return
 // dig.Out results.
 func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) error {
 	ctype := reflect.TypeOf(constructor)
--- a/dig.go
+++ b/dig.go
@@ -92,9 +92,10 @@ func New(opts ...Option) *Container {
 // same types are requested multiple times, the previously produced value will
 // be reused.
 //
-// In addition to accepting vanilla constructors, Provide also accepts
-// constructors that accept dig.In structs as dependencies and/or return
-// dig.Out results.
+// In addition to accepting constructors that accept dependencies as separate
+// arguments and produce results as separate return values, Provide also
+// accepts constructors that specify dependencies as dig.In structs and/or
+// specify results as dig.Out structs.
 func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) error {
 	ctype := reflect.TypeOf(constructor)
 	if ctype == nil {
--- a/dig.go
+++ b/dig.go
@@ -81,7 +81,7 @@ func New(opts ...Option) *Container {
 }
 
 // Provide teaches the container how to build values of one or more types and
-// expresses their dependency types.
+// expresses their dependencies.
 //
 // The first argument of Provide is a function that accepts zero or more
 // parameters and returns one or more results. The function may optionally
--- a/dig.go
+++ b/dig.go
@@ -111,9 +111,10 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 }
 
 // Invoke runs the given function after instantiating its dependencies.
-// Any arguments that the function has are treated as its dependencies and
-// they are instantiated in an unspecified order along with any dependencies
-// that they might have.
+//
+// Any arguments that the function has are treated as its dependencies. The
+// dependencies are instantiated in an unspecified order along with any
+// dependencies that they might have.
 //
 // The function may return an error to indicate failure. The error will be
 // returned to the caller as-is.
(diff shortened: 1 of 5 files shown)


dig.go
```diff
@@ -73,26 +72,28 @@ type Container struct {
...
+// New constructs a Container.
 func New(opts ...Option) *Container {
 	return &Container{
 		nodes: make(map[key]*node),
 		cache: make(map[key]reflect.Value),
 	}
 }
 
-// Provide teaches the Container how to construct one or more new types.
+// Provide teaches the container how to build values of one or more types.
```
kriskowal: …and expresses their dependency types.

dig.go
```diff
@@ -107,12 +108,12 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 	return nil
 }
 
-// Invoke runs a function, supplying its arguments from the Container. If the
-// function's last return value is an error, that error is propagated to the
-// caller. All other returned values (if any) are ignored.
+// Invoke runs the given function immediately, supplying its arguments from
```
kriskowal: Specify "immediately".

- [ ] Before it returns?
- [ ] Order of instantiation of dependencies?

Describe how its arguments are dependencies and causes Dig to retain an instance of all its transitive dependencies. Consider well known terms like "roots" of the DAG.

doc.go
```diff
@@ -18,121 +18,235 @@
...
 // Status
 //
-// BETA. Expect potential API changes.
+// STABLE. No breaking changes will be made in this major version.
 //
 // Container
 //
-// package dig exposes type Container as an object capable of resolving a
-// directional dependency graph.
+// Dig exposes type Container as an object capable of resolving a directional
```
kriskowal: directed, though I think "dependency" implies directed.

doc.go
```diff
@@ -18,121 +18,235 @@
...
 //
+//   c := dig.New()
 //
-// To create one:
+// Provide
 //
-//   import "go.uber.org/dig"
+// Constructors for different types are added to the container by using the
+// Provide method. A constructor can declare a dependency on another type by
+// simply adding it as a function parameter. Dependencies for a type can be
```
kriskowal: x simply, adds nothing to meaning.

doc.go
```diff
@@ -18,121 +18,235 @@
...
+//     // ...
 //   }
 //
-// **All objects in the container are treated as a singletons**, meaning there can be
-// only one object in the graph of a given type.
+//   if err := c.Provide(newDBConnection); err != nil {
+//     // ...
+//   }
 //
+// Multiple constructors can rely on the same type. Types in the container are
```
kriskowal: The container creates a singleton for each retained type. (solves passive voice by making the container the subject of the sentence)

doc.go
```diff
@@ -18,121 +18,235 @@
...
-// // TODO: docs on dig.Out usage
+// If the invoked function returns an error, that error is propagated to the
+// caller.
 //
+// Parameter Objects
 //
-// Invoke
+// Constructors declare their dependencies as function parameters. This can
+// very quickly become unreadable if the constructor has a lot of
+// dependencies.
```
kriskowal: Using a parameter object also makes the dependencies order independent and allows the contract to add optional dependencies without a breaking change.

doc.go
```diff
@@ -18,121 +18,235 @@
...
 //
-// Invoke looks through the graph and resolves all the constructor parameters for execution.
+//   func NewHandler(p HandlerParams, l *log.Logger) *Handler {
+//     // ...
+//   }
+//
+// Result Objects
+//
+// Result objects are the flip side of parameter objects. These are structs
+// which represent multiple outputs from a single function as fields in the
```
kriskowal: As an easy rule to apply in practice, "that" is grammatically proper anywhere you can use it instead of "which" without it sounding wrong.

doc.go
```diff
@@ -18,121 +18,235 @@
...
+//    Posts    *PostGateway
+//  }
+//
+//  func SetupGateways(conn *sql.DB) (Gateways, error) {
+//    // ...
+//  }
+//
+// Optional Dependencies
+//
+// Often times constructors don't have a hard dependency on some types and
```
kriskowal: Oftentimes is one word, but just Often will do.

doc.go
```diff
@@ -18,121 +18,235 @@
...
-// • If an error is returned from an invoked function, it will be propagated to the caller
+//   func NewUserGateway(p UserGatewayParams, log *log.Logger) (*UserGateway, error) {
+//     if p.Cache != nil {
+//       log.Print("Logging disabled")
+//     }
+//     // ...
+//   }
 //
-// Here is a fully working somewhat real-world Invoke example:
+// Constructors which declare dependencies as optional MUST handle the case of
```
kriskowal: dat

doc.go
```diff
@@ -18,121 +18,235 @@
...
-//   	"go.uber.org/dig"
-//   	"go.uber.org/zap"
-//   )
+// Some use cases call for multiple values of the same type. Dig allows adding
+// multiple values of the same type to the container with the use of
+// `name:".."` tags on fields of dig.In and dig.Out structs.
 //
-//   func main() {
-//   	c := dig.New()
+// A constructor which produces a dig.Out struct can tag any field with
```
kriskowal: dat

types.go
```diff
@@ -41,45 +41,67 @@ type digSentinel interface {
...
+//               value. See Named Values for more information.
+type Out struct{ digSentinel }
 
 func isError(t reflect.Type) bool {
 	return t.Implements(_errType)
 }
 
-// IsIn returns true if passed in type embeds dig.In either directly
-// or through another embedded field.
+// IsIn checks if the given struct is a dig.In struct. A struct qualifies as a
```
kriskowal: "whether" is an expression; "if" is a statement.

dig.go
```diff
@@ -73,26 +72,28 @@ type Container struct {
...
-// on demand. They must return at least one non-error value, all of which are
-// then available in the Container. If the last returned value is an error, the
-// Container inspects it to determine whether the constructor succeeded or
-// failed. Regardless of position, returned errors are never put into the
-// Container's dependency graph.
+// The first argument of Provide is a function which accepts zero or more
+// parameters and returns one or more results. The function may optionally
+// return an error to indicate that it failed to build the value. This
+// function will be treated as the constructor for all the types it returns.
+// This function will be called AT MOST ONCE when a type produced by it, or a
```
alsamylkin: Isn't it going to be executed `exactly ONCE` because it is requested by Invoke?
Should we also add explicitly that this function will not be called if returned types are not requested?

dig.go
```diff
@@ -73,26 +72,28 @@ type Container struct {
...
+// type which depends on this function's output, is requested via Invoke. If
+// the same types are requested multiple times, the previously produced value
+// will be reused.
 //
-// All non-functions (including structs, pointers, Go's built-in collections,
-// and primitive types like ints) are inserted into the Container as-is.
+// In addition to accepting vanilla constructors, Provide also accepts
+// constructors which accept dig.In structs as dependencies and/or return
```
breerly: I would use "take" here instead of "accept", since you already used "accepting" earlier in the sentence.

dig.go
```diff
@@ -73,26 +72,28 @@ type Container struct {
...
+// dig.Out results.
```
alsamylkin: May be add a line about handling cycles?

doc.go
```diff
@@ -18,121 +18,235 @@
...
+//   if err != nil {
+//     // ...
+//   }
 //
-// // TODO: docs on dig.In usage
-// // TODO: docs on dig.Out usage
+// If the invoked function returns an error, that error is propagated to the
+// caller.
 //
+// Parameter Objects
```
alsamylkin: I believe we need to add a note on variadic constructors and invokes.

types.go
```diff
@@ -41,45 +41,67 @@ type digSentinel interface {
...
+// dig.In struct if it embeds the dig.In type or if any struct that it embeds
+// is a dig.In struct. The parameter may be the reflect.Type of the struct
+// rather than the struct itself.
 //
-// Parameter can be a struct directly, or reflect.Type of it.
+// A struct MUST qualify as a dig.In struct for its fields to support dig.In
```
alsamylkin: nit: dig.In structs are always supported, probably rephrase:
`A struct MUST qualify as a dig.In struct for its fields to make dig treat them as special
parameter structs.` ?

dig.go
```diff
@@ -73,26 +72,28 @@ type Container struct {
...
+// return an error to indicate that it failed to build the value. This
+// function will be treated as the constructor for all the types it returns.
+// This function will be called AT MOST ONCE when a type produced by it, or a
+// type which depends on this function's output, is requested via Invoke. If
+// the same types are requested multiple times, the previously produced value
+// will be reused.
 //
-// All non-functions (including structs, pointers, Go's built-in collections,
-// and primitive types like ints) are inserted into the Container as-is.
+// In addition to accepting vanilla constructors, Provide also accepts
```
glibsm: s/vanilla/?

I understand the term but we should be more explicit here

doc.go
```diff
@@ -18,121 +18,235 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-// Package dig is the dig: Dependency Injection Framework for Go.
-//
-// package dig provides an opinionated way of resolving object dependencies.
+// Package dig provides an opinionated way of resolving object dependencies.
```
glibsm: Editing this file is in vein unless Makefile gets fixed and `gendoc` is removed.
Right now it's auto generated

doc.go
```diff
@@ -18,121 +18,235 @@
...
+//   err := c.Invoke(func(server *http.Server) error {
+//     // ...
+//   })
+//   if err != nil {
+//     // ...
+//   }
 //
-// // TODO: docs on dig.In usage
-// // TODO: docs on dig.Out usage
+// If the invoked function returns an error, that error is propagated to the
```
alsamylkin: Do we need to add a special paragraph about handling pointer and value types?
It is obvious for us, but not really for users. Although it can be handled by examples.

types.go
```diff
@@ -41,45 +41,67 @@ type digSentinel interface {
...
 type In struct{ digSentinel }
 
 // Out is an embeddable type that signals to dig that the returned
 // struct should be treated differently. Instead of the struct itself
 // becoming part of the container, all members of the struct will.
-type Out struct{ digSentinel }
 
-// TODO: better usage docs
-// Try to add some symmetry for In-Out docs as well.
+// Out may be embedded into structs to request dig to treat them as special
```
glibsm: `as special.`, full stop maybe?

types.go
```diff
@@ -41,45 +41,67 @@ type digSentinel interface {
...
+// result structs. When a constructor produces such a struct as a result,
```
glibsm: s/as a result//

types.go
glibsm: i sort of see what you are trying to say, but this sentence confused me on first read

dig.go
abhinav: I've generally been trying to avoid graph theory terms in the documentation.

doc.go
abhinav: I don't think we specify the dependency instantiation contract anywhere. It has always been unspecified.

abhinav: Mentioned the breaking change thing in the Optional Dependencies section.

dig.go
abhinav: True. How about,

```
// In addition to accepting constructors that accept dependencies as separate
// arguments and produce results as separate return values, Provide also
// accepts constructors that specify dependencies as dig.In structs and/or
// specify results as dig.Out structs.
```

doc.go
abhinav: removed

dig.go
abhinav: This is for Provide. It will be at most once. The doc.go explains that things are only called if requested directly or transitively.

dig.go
abhinav: Added a note about the graph being acyclic

doc.go
abhinav: We actually don't specify anything except types anywhere in the code. I think we can iterate if there's confusion around it.

doc.go
abhinav: Added info about variadic constructors

types.go
abhinav: That doesn't sound quite right. Changed to,

> [..] for its fields to be treated specially by dig.

types.go
abhinav: "special" is an adjective here, not a noun. There are two words that are part of this sentence on the following line.

types.go
abhinav: changed to,

> When a constructor returns such a struct, [..]

types.go
abhinav: reworded

abhinav:
All comments should be addressed. If there are no objections, I'll be merging this today or tomorrow at the latest.

dig.go
```diff
@@ -73,26 +72,30 @@ type Container struct {
...
 func New(opts ...Option) *Container {
 	return &Container{
 		nodes: make(map[key]*node),
 		cache: make(map[key]reflect.Value),
 	}
 }
 
-// Provide teaches the Container how to construct one or more new types.
+// Provide teaches the container how to build values of one or more types and
+// expresses their dependency types.
```
akshayjshah: s/dependency/dependent?

dig.go
```diff
@@ -107,12 +110,13 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 	return nil
 }
 
-// Invoke runs a function, supplying its arguments from the Container. If the
-// function's last return value is an error, that error is propagated to the
-// caller. All other returned values (if any) are ignored.
+// Invoke runs the given function after instantiating its dependencies.
+// Any arguments that the function has are treated as its dependencies and
+// they are instantiated in an unspecified order along with any dependencies
+// that they might have.
```
akshayjshah: Break into two sentences?

```
...are treated as its dependencies. Dependencies are instantiated...
```

doc.go
```diff
@@ -18,121 +18,253 @@
...
+//     ReadFromConn *sql.DB `name:"ro" optional:"true"`
+//   }
+//
+//   func NewCommentGateway(p GatewayParams, log *log.Logger) (*CommentGateway, error) {
+//     if p.ReadFromConn == nil {
+//       log.Print("Warning: Using RW connection for reads")
+//       p.ReadFromConn = p.WriteToConn
+//     }
+//     // ...
+//   }
```
akshayjshah: Awesome - really appreciate the effort that went into this. Can we also file an issue to convert as many of these code samples as possible to testable examples?

types.go
```diff
@@ -41,45 +41,67 @@ type digSentinel interface {
...
 }
 
-// In is an embeddable object that signals to dig that the struct
-// should be treated differently. Instead of itself becoming an object
-// in the graph, memebers of the struct are inserted into the graph.
+// In may be embedded into structs to request dig to treat them as special
+// parameter structs. When a constructor accepts such a struct, instead of the
+// struct becoming a dependency for that constructor, all its fields become
+// dependencies instead. See the section on Parameter Objects above for more
+// information.
```
akshayjshah: Let's not presume that everyone's looking at the docs in a browser - can we say, "See the section on Parameter Objects in the package-level documentation for more details"?

dig.go
abhinav: s/dependency types/dependencies

doc.go
abhinav: Done https://github.com/uber-go/dig/issues/143

PR #144, opened 2017-08-29
Test Go 1.9

breerly:


.travis.yml
```diff
@@ -6,7 +6,7 @@ matrix:
   include:
     - go: 1.7
     - go: 1.8
-    # TODO: Switch to Go 1.9 when released
+    - go: 1.9
     - go: tip
       env: LINT=1
```
abhinav: can we move the lint to 1.9 rather than tip?

it was on tip before because of 1.9 specific syntax used in one of the files

PR #146, opened 2017-09-14
Add RootCause to retrieve original error

### LINKED ISSUE #145: Support for leaving constructor errors unchanged
dig should support an opt-in way of requesting that errors returned by user
constructors are propagated upwards as-is.

Currently, errors are wrapped with a lot of context:

```
failed to get arguments for [..]:
  couldn't get arguments for constructor [..]:
    constructor [..] failed:
      <actual error>
```

This is all useful in most scenarios except where the error is informative for
the rest of my program. For example, I could have a `parseCLIOptions() (*options, error)`
constructor which parses the command line arguments provided by the user. The
error returned by the function would usually contain the message that should
be printed to the user, but because of the wrapping, the error message is no
longer consumable by users of the CLI application.

I imagine this would take form as a simple Container option:

    dig.New(dig.WrapErrors(false))
    // Name TBD. Another option is WrapUserErrors which is very explicit about
    // what this is doing.
    
Note that this wouldn't apply to all error wrapping inside dig. Only errors
returned by user-provided functions on the Invoke path will be affected.

Note that Invoke already returns the error returned by the provided function
as-is.

    err := c.Invoke(func() error { return someError })
    fmt.Println(err == someError)
    // Output: true

I just want support for,

    c.Provide(func() (*type1, error) { return someError })
    err := c.Invoke(func(*type1) { panic("never called")})
    fmt.Println(err == someError)
    // Output: true

abhinav:
This adds a new `dig.RootCause(error) error` function which retrieves the
original error that caused an Invoke to fail.

See #145 for motivation but the gist of it is, this makes it possible to
retrieve and act upon the original error that a constructor of a dependency
returned when an Invoke call is made.

Implementation:
We use a custom error wrapper for `fmt.Errorf` calls which include the
original error at the end. So,

    return fmt.Errorf("something went wrong: %v", err)

Becomes,

    return errWrapf(err, "something went wrong")

The wrapper tracks the original `err` and `RootCause` is just an accessor for
it.

Because this came up offline: We're not using pkg/errors because we want to
return the error returned by the user-provided constructor. If we use pkg/
errors, and the user-provided constructor also uses pkg/errors, we won't
return what the user returned but rather, the root cause of that error.

Resolves #145

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -119,10 +119,6 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 // The function may return an error to indicate failure. The error will be
 // returned to the caller as-is.
 func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
-	return c.invoke(function, opts...)
-}
-
-func (c *Container) invoke(function interface{}, opts ...InvokeOption) error {
 	ftype := reflect.TypeOf(function)
 	if ftype == nil {
 		return errors.New("can't invoke an untyped nil")
--- a/dig.go
+++ b/dig.go
@@ -42,8 +42,13 @@ type key struct {
 // Option configures a Container. It's included for future functionality;
 // currently, there are no concrete implementations.
 type Option interface {
-	unimplemented()
+	applyOption(*Container)
 }
+
+// optionFunc is a function-based implementation of Option.
+type optionFunc func(*Container)
+
+func (f optionFunc) applyOption(c *Container) { f(c) }
 
 // A ProvideOption modifies the default behavior of Provide. It's included for
 // future functionality; currently, there are no concrete implementations.
@@ -62,6 +67,8 @@ type Container struct {
 	nodes map[key]*node
 	cache map[key]reflect.Value
 
+	wrapErrors bool
+
 	// TODO: for advanced use-case, add an index
 	// This will allow retrieval of a single type, without specifying the exact
 	// tag, provided there is only one object of that given type
@@ -74,10 +81,27 @@ type Container struct {
 
 // New constructs a Container.
 func New(opts ...Option) *Container {
-	return &Container{
-		nodes: make(map[key]*node),
-		cache: make(map[key]reflect.Value),
+	c := &Container{
+		nodes:      make(map[key]*node),
+		cache:      make(map[key]reflect.Value),
+		wrapErrors: true,
 	}
+
+	for _, opt := range opts {
+		opt.applyOption(c)
+	}
+	return c
+}
+
+// WrapErrors enables or disables wrapping of errors returned by user
+// functions.
+//
+// If enabled, errors returned by user-provided constructors will be wrapped
+// with more contextual information to aid in debugging the failure.
+//
+// Defaults to true.
+func WrapErrors(shouldWrap bool) Option {
+	return optionFunc(func(c *Container) { c.wrapErrors = shouldWrap })
 }
 
 // Provide teaches the container how to build values of one or more types and
@@ -105,7 +129,7 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 		return fmt.Errorf("must provide constructor function, got %v (type %v)", constructor, ctype)
 	}
 	if err := c.provide(constructor, ctype); err != nil {
-		return fmt.Errorf("can't provide %v: %v", ctype, err)
+		return errWrapf(err, "can't provide %v", ctype)
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


dig.go
```diff
@@ -74,10 +81,27 @@ type Container struct {
...
+	return c
+}
+
+// WrapErrors enables or disables wrapping of errors returned by user
+// functions.
+//
+// If enabled, errors returned by user-provided constructors will be wrapped
+// with more contextual information to aid in debugging the failure.
+//
+// Defaults to true.
```
glibsm: If it's true by default and the only thing to do is turn it off, maybe we can have an unnamed option? `dig.DisableErrorWrap()` or something like that.

dig.go
```diff
@@ -119,6 +143,15 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 // The function may return an error to indicate failure. The error will be
 // returned to the caller as-is.
 func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
+	err := c.invoke(function, opts...)
+	// If error wrapping wasn't requested, return the root cause.
+	if !c.wrapErrors {
```
glibsm: does `Provide` need the same logic too?

dig.go
```diff
@@ -119,6 +143,15 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
...
+		err = errRootCause(err)
```
glibsm: `func errCause(bool wrap, err)` might be an easier abstraction so you can just pass the `c.WrapErrors` in as opposed to having to write an if statement in each applicable case

dig.go
breerly: No - a constructor returning an error should always result in the program failing in the same way it does currently.

breerly: @abhinav you should specify on this PR and the issue that this change is intended to target `Invoke` only.

glibsm: I don't see how you can realistically differentiate between one but not the other. If `.New` fails with a constructor error it's going to be all muffled anyways, something like `failed to create type X: blah blah: blah blah`, which may or may not be desirable either.

That said, I think I'd prefer @prashantv 's suggestion of having an extract function. That way the user can make their own decision of what is and isn't desirable.

dig.go
abhinav: dig.RootCause makes this moot but the reason was that `dig.WrapErrors(true/false)` is pretty self-documenting.

dig.go
abhinav: I think @breerly's point was that functions don't get called on Provide so that path remains unchanged. Functions get called only on the Invoke path, so that's the only place where we needed to consider this option.

dig.go
abhinav: Moot now.

abhinav:
Updated with new API @breerly @glibsm @prashantv

error.go
```diff
@@ -0,0 +1,81 @@
...
+//   RootCause(errWrapf(errWrapf(err, ...), ...)) == err
+//
+// Use errWrapf in the rest of dig in place of fmt.Errorf if the message ends
+// with ": <original error>".
+func errWrapf(err error, msg string, args ...interface{}) error {
+	if err == nil {
+		return nil
+	}
+
+	rootCause := err
```
prashantv: can you do `rootCause := RootCause(err)`? we could probably inline that into the creation of `wrappedError`

error.go
```diff
@@ -0,0 +1,81 @@
...
+		err:       fmt.Errorf("%v: %v", msg, err),
+	}
+}
+
+// wrappedError is a wrapper around error that tracks the root cause of the
+// error.
+//
+// The root cause will be retained between errWrapf calls and retrievable by
+// using RootCause.
+type wrappedError struct {
```
prashantv: maybe move the type to the beginning since other functions use it

dig_test.go
```diff
@@ -1291,6 +1292,7 @@ func TestInvokeFailures(t *testing.T) {
 		})
 		require.Error(t, err, "expected invoke error")
 		assert.Contains(t, err.Error(), "couldn't get arguments for constructor", "unexpected error text")
```
prashantv: should we make sure `err.Error()` also includes `errFailed.Error()`?

error.go
abhinav: good call

dig.go
glibsm: Yeah, it makes sense - we want to bubble up user constructor errors back up the stack through `RootCause`.

I was just saying that the name `RootCause` in relation to just `Invoke` can be pretty confusing. Technically, `.New` errors when provides fail also have a "root cause": bad argument types, no returns, etc. etc.

PR #148, opened 2017-09-15
Better error messaging during mixed types usage

### LINKED ISSUE #147: Better error message when value/pointer types are mixed
Right now, if the user requires a value type and only a pointer is provided (or vice versa), the error message is the generic "container is missing [TYPE]"

It might be nice to detect this case and suggest the other variation.

glibsm:
This is something that has been mentioned several times - it's very easy
to request a wrong type on accident - is it a pointer? is it value type?
In a large enough framework base it's a giant mix of both. This adds a
much better error message when dig is able to detect an easily
correctable case.

Fixes #147

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -1403,4 +1403,60 @@ func TestInvokeFailures(t *testing.T) {
 		assert.Contains(t, err.Error(), "dig.in embeds *dig.In")
 		assert.Contains(t, err.Error(), "embed dig.In value instead")
 	})
+
+	t.Run("requesting a value or pointer when other is present", func(t *testing.T) {
+		type A struct{}
+		type outA struct {
+			Out
+
+			A `name:"hello"`
+		}
+		type B struct{}
+
+		cases := []struct {
+			name        string
+			provide     interface{}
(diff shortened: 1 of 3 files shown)


dig_test.go
```diff
@@ -1403,4 +1403,60 @@ func TestInvokeFailures(t *testing.T) {
...
+			name        string
+			provide     interface{}
+			invoke      interface{}
+			errContains []string
+		}{
+			{
+				name:        "value missing, pointer present",
+				provide:     func() *A { return &A{} },
+				invoke:      func(A) {},
+				errContains: []string{"dig.A is not in the container, did you mean to use *dig.A"},
```
breerly: This is phenomenal.

breerly: Might be worth adding a `?` to the end of the err.

glibsm: Totally agree, in fact I included it :) The question mark is a few words later:

`type *dig.A is not in the container, did you mean to use dig.A that is present?`

I wanted to signal that the second type is present in the container. I could use some help with this copy if you guys have better ideas

PR #149, opened 2017-09-27
Provide: Inspect constructor parameters upfront

abhinav:
As a first-step to DRYing up the graph resolution logic in dig, this
changes Provide to inspect constructors in more detail upfront.

Their dependencies, whether they are dig.In objects, etc. is now
known on the Provide path rather than Invoke.

This information is stored with the help of a `param` interface which
unionizes the following parameter types:

- paramList: List of parameters of a constructor
- paramSingle: Basic argument type. Nothing special about this
- paramObject: `dig.In` object whose fields are other `param`s

This opens room for more `param` types.

Note: The `Dependencies()` method will be removed in a future change. It
exists only to minimize the impact of this specific change.

Also note that no new errors are being surfaced to the user. All errors
returned here are copies of existing errors returned by the library. Of
note is the logic in `newParamObject` which returns an error when
private fields are encountered. In dig currently, that logic exists in
two places:

- https://github.com/uber-go/dig/blob/809ed57c66b4c4b74e882be9da77d997ec439cde/dig.go#L363-L367
- https://github.com/uber-go/dig/blob/809ed57c66b4c4b74e882be9da77d997ec439cde/dig.go#L532-L534

If you look at the implementations, one of them skips private fields
while the other returns an error. The one that errors out does so at
`Invoke` time rather than `Provide`. With this change, the error will be
returned at `Provide` time instead, which is preferable anyway. This is
not actually a breaking change because we're failing fast rather than
waiting for an Invoke.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -12,3 +12,8 @@ coverage:
                                #   option: "X%" a static target percentage to hit
         if_not_found: success  # if parent is not found report status as success, error, or failure
         if_ci_failed: error    # if ci fails report status as success, error, or failure
+
+    patch:
+      default:
+        enabled: yes
+        target: 95
--- a/dig.go
+++ b/dig.go
@@ -248,7 +248,7 @@ func traverseOutTypes(k key, f func(key) error) error {
 
 		if field.PkgPath != "" {
 			return fmt.Errorf(
-				"private fields not allowed in dig.Out, did you mean to export %q (%v) from %v",
+				"unexported fields not allowed in dig.Out, did you mean to export %q (%v) from %v",
 				field.Name, field.Type, k.t)
 		}
 
@@ -362,7 +362,7 @@ func (c *Container) createInObject(t reflect.Type) (reflect.Value, error) {
 
 		if f.PkgPath != "" {
 			return dest, fmt.Errorf(
-				"private fields not allowed in dig.In, did you mean to export %q (%v) from %v?",
+				"unexported fields not allowed in dig.In, did you mean to export %q (%v) from %v?",
 				f.Name, f.Type, t)
 		}
 
--- a/dig.go
+++ b/dig.go
@@ -316,7 +316,7 @@ func (c *Container) get(e edge) (reflect.Value, error) {
 		return _noValue, fmt.Errorf("type %v isn't in the container", e.key)
 	}
 
-	if err := c.contains(n.deps); err != nil {
+	if err := c.contains(n.Params.Dependencies()); err != nil {
 		if e.optional {
 			return reflect.Zero(e.t), nil
 		}
@@ -438,7 +438,9 @@ type node struct {
 
 	ctor  interface{}
 	ctype reflect.Type
-	deps  []edge
+
+	// Type information about constructor parameters.
+	Params paramList
 }
 
 type edge struct {
@@ -448,27 +450,17 @@ type edge struct {
 }
 
 func newNode(k key, ctor interface{}, ctype reflect.Type) (*node, error) {
-	deps, err := getConstructorDependencies(ctype)
-	return &node{
-		key:   k,
-		ctor:  ctor,
-		ctype: ctype,
-		deps:  deps,
-	}, err
-}
-
-// Retrieves the dependencies for a constructor
-func getConstructorDependencies(ctype reflect.Type) ([]edge, error) {
-	var deps []edge
-	for _, t := range getConstructorArgTypes(ctype) {
-		err := traverseInTypes(t, func(e edge) {
-			deps = append(deps, e)
-		})
-		if err != nil {
-			return nil, err
-		}
+	params, err := newParamList(ctype)
+	if err != nil {
+		return nil, err
 	}
-	return deps, nil
+
+	return &node{
+		key:    k,
+		ctor:   ctor,
+		ctype:  ctype,
+		Params: params,
+	}, err
 }
 
 // Retrieves the types of the arguments of a constructor in-order.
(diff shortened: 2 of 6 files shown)


param.go
```diff
@@ -0,0 +1,187 @@
...
+		Type   reflect.Type
+		Fields []paramObjectField
+
+		deps []edge
+	}
+
+	// paramObjectField is a single field of a dig.In struct.
+	paramObjectField struct {
+		// Name of the field specified via a name:".." tag.
+		Name string
```
willhug: We already cover this in the paramSimple field, do we need it here?

Or, if we want to go the other way, should we pull the Optional field to this level too, and remove the Name/Optional fields from the paramSimple?

param.go
```diff
@@ -0,0 +1,187 @@
...
+
+// newParamObject builds an paramObject from the provided type. The type MUST
+// be a dig.In struct.
+func newParamObject(t reflect.Type) (paramObject, error) {
+	op := paramObject{Type: t}
+
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		if f.Type == _inType {
+			// Skip over the dig.In embed.
```
willhug: uhhhhh, doesn't this break things?

param.go
```diff
@@ -0,0 +1,187 @@
...
+			continue
+		}
+
+		if f.PkgPath != "" {
+			return op, fmt.Errorf(
```
willhug: breaking change?

dig.go
```diff
@@ -438,7 +438,9 @@ type node struct {
 
 	ctor  interface{}
 	ctype reflect.Type
-	deps  []edge
+
+	// Type information about constructor parameters.
+	Params paramList
```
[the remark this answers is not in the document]
abhinav: As discussed offline: Yes. Generally speaking, it's a good idea to clarify API
boundaries even between unexported types.

This states that `Params` is part of the API of `node` for the rest of `dig`.

param.go
```diff
@@ -0,0 +1,187 @@
...
+//  paramSimple   An explicitly requested type.
+//  paramObject   dig.In struct where each field in the struct can be another
+//                param.
+type (
+	param interface {
+		param()
+
+		// Human-friendly name for this param type. We use this exclusively in
+		// error messages.
+		ParamTypeName() string
```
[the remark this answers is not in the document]
abhinav: `param()` is private to conform with the pattern of declaring an unexported
method on an interface which ensures that all implementations of that
interface are in that same package.

The rest are public because of the reason stated above.

param.go
abhinav: Documentation here was wrong. This is the name of the field in the struct. Fixed. Thanks for catching.

param.go
abhinav: This doesn't break nested dig.Ins. (Test added.)

This is skipping the embed of the actual `dig.In` type.

param.go
abhinav: No. Code here lifted from elsewhere in the library:

https://github.com/uber-go/dig/blob/809ed57c66b4c4b74e882be9da77d997ec439cde/dig.go#L250

.codecov.yml
```diff
@@ -12,3 +12,8 @@ coverage:
                                #   option: "X%" a static target percentage to hit
         if_not_found: success  # if parent is not found report status as success, error, or failure
         if_ci_failed: error    # if ci fails report status as success, error, or failure
+
+    patch:
+      default:
+        enabled: yes
+        target: 95
```
glibsm: you might want to drop the total 98 target as well

abhinav: The full project is still over 98 so we can keep it for now IMO. If we find that >2% of the project actually contains unused code, we can drop the coverage target there too.

param.go
```diff
@@ -0,0 +1,189 @@
...
+		if f.PkgPath != "" {
+			return op, fmt.Errorf(
+				"private fields not allowed in dig.In, did you mean to export %q (%v) from %v?",
+				f.Name, f.Type, t)
+		}
+
+		p, err := newParam(f.Type)
+		if err != nil {
+			return op, errWrapf(err, "bad field %q of %v", f.Name, t)
+		}
```
akshayjshah: Realize that we don't want to pull in `multierr` here, but it'd be really nice to validate all the fields before returning - if there are multiple errors in a single params struct, it's pretty annoying to have them flagged one at a time. Can we just string-concat an error for the whole struct?

param.go
```diff
@@ -0,0 +1,189 @@
...
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		if f.Type == _inType {
+			// Skip over the dig.In embed.
+			continue
+		}
+
+		if f.PkgPath != "" {
+			return op, fmt.Errorf(
+				"private fields not allowed in dig.In, did you mean to export %q (%v) from %v?",
```
akshayjshah: Can we stick to Go idiom and replace "private" with "unexported"?

param.go
```diff
@@ -0,0 +1,189 @@
...
+)
+
+// The param interface represents a dependency for a constructor.
+//
+// The following implementations exist:
+//  paramList     All arguments of the constructor.
+//  paramSimple   An explicitly requested type.
+//  paramObject   dig.In struct where each field in the struct can be another
+//                param.
+type (
```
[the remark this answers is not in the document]
abhinav: I was trying to go for the go/ast style grouping but I see your point. Changing.

param.go
abhinav: I agree. There are other places where the errors should be collected too. I would prefer to do that in a follow-up diff if you don't mind. That's a lot of new behavior for this diff.

PR #150, opened 2017-09-27
Implement String() using `param`s

abhinav:
This adds String implementations for the different `param` types and
uses those in `node.String`.

### DIFF
--- a/param.go
+++ b/param.go
@@ -33,7 +33,7 @@ import (
 //  paramObject   dig.In struct where each field in the struct can be another
 //                param.
 type param interface {
-	String() string
+	fmt.Stringer
 
 	// Comprehensive list of dependencies this parameter represents.
 	Dependencies() []edge
--- a/param.go
+++ b/param.go
@@ -33,6 +33,8 @@ import (
 //  paramObject   dig.In struct where each field in the struct can be another
 //                param.
 type param interface {
+	String() string
+
 	// Comprehensive list of dependencies this parameter represents.
 	Dependencies() []edge
 }
--- a/stringer.go
+++ b/stringer.go
@@ -23,6 +23,7 @@ package dig
 import (
 	"bytes"
 	"fmt"
+	"strings"
 )
 
 // String representation of the entire Container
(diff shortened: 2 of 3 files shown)


param.go
```diff
@@ -34,6 +34,8 @@ import (
 //                param.
 type (
 	param interface {
+		String() string
```
akshayjshah: This is a really well-known signature - can we just embed `fmt.Stringer` instead?

It'd be weird if a handler interface explicitly had `ServeHTTP(http.ResponseWriter, *http.Request)` in its definition instead of embedding `http.Handler`, and this feels the same way to me.

stringer.go
```diff
@@ -66,3 +54,34 @@ func (k key) String() string {
 	}
 	return k.t.String()
 }
+
+func (pl paramList) String() string {
+	args := make([]string, len(pl.Params))
+	for i, p := range pl.Params {
+		args[i] = p.String()
+	}
+	return fmt.Sprint(args)
```
akshayjshah: I'm restraining myself from golfing allocations here, but I want you to know that I know you're trolling me.

stringer.go
```diff
@@ -66,3 +54,34 @@ func (k key) String() string {
...
+
+	return fmt.Sprintf("%s%v%s", prefix, sp.Type, suffix)
+}
+
+func (op paramObject) String() string {
+	fields := make([]string, len(op.Fields))
+	for i, f := range op.Fields {
+		fields[i] = f.Param.String()
+	}
+	return strings.Join(fields, " ")
```
akshayjshah: so-what

What, we suddenly care about the overhead of `fmt.Sprint`? Srsly tho let's use one or the other.

stringer_test.go
```diff
@@ -39,7 +39,7 @@ func TestStringer(t *testing.T) {
 	type D struct{}
 	type param struct {
 		In
-		*D `named:"foo" optional:"true"`
+		*D `name:"foo" optional:"true"`
```
akshayjshah: Nice catch.

stringer.go
abhinav: Discussed offline. We're doing this to preserve current behavior exactly as-is. We can iterate on this in the future.

PR #151, opened 2017-09-27
Detect cycles in `param`

abhinav:
This updates the cycle detection logic to rely on `param` types.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -261,7 +261,7 @@ func traverseOutTypes(k key, f func(key) error) error {
 }
 
 func (c *Container) isAcyclic(n *node) error {
-	return detectCycles(n, c.nodes, nil)
+	return detectCycles(n.Params, c.nodes, []key{n.key})
 }
 
 // Retrieve a type from the container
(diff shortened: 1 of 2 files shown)


dig.go
```diff
@@ -492,21 +492,37 @@ func cycleError(cycle []key, last key) error {
...
-			continue
+			return nil
 		}
-		if err := detectCycles(depNode, graph, path); err != nil {
-			return err
+
+		return detectCycles(n.Params, graph, path)
+	case paramObject:
+		for _, f := range p.Fields {
+			if err := detectCycles(f.Param, graph, path); err != nil {
```
breerly: Worth making the test recurse at least once?

dig.go
```diff
@@ -492,21 +492,37 @@ func cycleError(cycle []key, last key) error {
...
+				return err
+			}
 		}
+	default:
+		panic(fmt.Sprintf("unknown param type %T", param))
```
akshayjshah: This is one of those times when I really miss sum types and compiler-enforced exhaustiveness when dealing with them :/

Can we make this error message clearly indicate that the user can't fix this panic? It's an error in dig, and the user should immediately file an issue.

```
panic(fmt.Sprintf("Sorry - you've found a bug in dig! Please file an issue at https://github.com/uber-go/dig/issues/. Details: unknown param type %T", param))
```

PR #152, opened 2017-09-27
Cycle Detection: Use special error type

abhinav:
Instead of the janky append-to-string and `errors.New`, this changes the
cycle detection logic to use an explicit error type.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -483,13 +483,18 @@ func getConstructorArgTypes(ctype reflect.Type) []reflect.Type {
 	return args
 }
 
-func cycleError(cycle []key, last key) error {
-	b := &bytes.Buffer{}
-	for _, k := range cycle {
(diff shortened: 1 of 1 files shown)


dig.go
```diff
@@ -483,13 +483,18 @@ func getConstructorArgTypes(ctype reflect.Type) []reflect.Type {
...
+
+func (e errCycleDetected) Error() string {
+	b := new(bytes.Buffer)
+	for _, k := range e.Path {
 		fmt.Fprintf(b, "%v ->", k.t)
 	}
-	fmt.Fprintf(b, "%v", last.t)
-	return errors.New(b.String())
+	fmt.Fprintf(b, "%v", e.Key.t)
+	return b.String()
```
akshayjshah: Come at me, I'm untrollable.

PR #153, opened 2017-09-27
Shallow dependency check for `param` types

abhinav:
This implements a shallow check (previously implemented by
`Container.contains`) in terms of `param`. This check just verifies that
constructors for all dependencies of the provided param are present in
the container.

Note that this is the last usage of `Dependencies()` and we should be
able to delete it now.

### DIFF
--- a/param.go
+++ b/param.go
@@ -45,6 +45,24 @@ var (
 	_ param = paramList{}
 )
 
+// Calls the provided function on all paramSingles in the given param tree.
+func forEachSimpleParam(param param, f func(paramSingle)) {
+	switch p := param.(type) {
+	case paramList:
+		for _, arg := range p.Params {
+			forEachSimpleParam(arg, f)
+		}
+	case paramSingle:
+		f(p)
+	case paramObject:
+		for _, field := range p.Fields {
+			forEachSimpleParam(field.Param, f)
+		}
+	default:
+		panic(fmt.Sprintf("unknown param type %T", param))
+	}
+}
+
 // newParam builds a param from the given type. If the provided type is a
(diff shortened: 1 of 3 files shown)


param.go
```diff
@@ -83,6 +83,24 @@ func (paramList) ParamTypeName() string   { return "parameter list" }
 func (paramSimple) ParamTypeName() string { return "simple type" }
 func (paramObject) ParamTypeName() string { return "parameter object" }
 
+// Calls the provided function on all paramSimples in the given param tree.
+func forEachSimpleParam(param param, f func(paramSimple)) {
```
glibsm: `forEachSimpleParam` name doesn't make sense since function is not restricted to simple params

abhinav: walkSimpleParams?

glibsm: the _simple_ param is what I care about, not the first part :)

The way I see it it walks/traverses _any_ param if the type switch to be believed, not just the simple ones, right?

abhinav: Yeah, but it calls the function only on the simple params. It's a "walk the params and call this function an all simpleParams" which is why I thought forEach made sense.

param.go
```diff
@@ -92,6 +92,24 @@ var (
...
+			forEachSimpleParam(arg, f)
+		}
+	case paramSingle:
+		f(p)
+	case paramObject:
+		for _, field := range p.Fields {
+			forEachSimpleParam(field.Param, f)
+		}
+	default:
+		panic(fmt.Sprintf("unknown param type %T", param))
```
akshayjshah: Same comment here about the panic message: let's make it absolutely clear that this is a bug in dig, and that the user can't possibly fix it.

PR #155, opened 2017-09-27
Build dependencies using `param`s

abhinav:
This moves the logic to build a `reflect.Value` given a `*Container`
from `Container.get` into individual `param` types.

- `paramList` builds a `[]reflect.Value` by calling its underlying
  `param`s.
- `paramSingle` builds a `reflect.Value` by fetching its cached value
  from the container or retrieving and calling its constructor.
- `paramObject` builds the corresponding `dig.In` struct and fills it
  with `reflect.Value`s retrieved from the underlying `param`s.

Besides breaking up the massive function into consumable chunks, this
lets us delete a number of things that were duplicating some of the
reflection logic.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -126,10 +126,17 @@ func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
 	if ftype.Kind() != reflect.Func {
 		return fmt.Errorf("can't invoke non-function %v (type %v)", function, ftype)
 	}
-	args, err := c.constructorArgs(ftype)
+
+	pl, err := newParamList(ftype)
+	if err != nil {
+		return err
+	}
+
+	args, err := pl.BuildParams(c)
 	if err != nil {
 		return errWrapf(err, "failed to get arguments for %v (type %v)", function, ftype)
 	}
+
 	returned := reflect.ValueOf(function).Call(args)
 	if len(returned) == 0 {
 		return nil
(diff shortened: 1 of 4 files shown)


dig.go
```diff
@@ -264,123 +271,6 @@ func (c *Container) isAcyclic(n *node) error {
...
-	if embedsType(e.t, _inPtrType) {
-		return _noValue, fmt.Errorf(
-			"%v embeds *dig.In which is not supported, embed dig.In value instead", e.t,
-		)
-	}
-
-	if e.t.Kind() == reflect.Ptr {
-		if IsIn(e.t.Elem()) {
-			return _noValue, fmt.Errorf(
-				"dependency %v is a pointer to dig.In, use value type instead", e.t,
```
abhinav: These checks are now part of newParam. So, they're done on `Provide` rather than `Invoke`.

dig.go
```diff
@@ -264,123 +271,6 @@ func (c *Container) isAcyclic(n *node) error {
...
-	n, ok := c.nodes[e.key]
-	if !ok {
-		// Unlike in the fallback case below, if a user makes an error requesting
-		// a mixed type for an optional parameter, a good error message "did you mean X?"
-		// will not be used and dig will return zero value.
-		if e.optional {
-			return reflect.Zero(e.t), nil
-		}
-
-		// If the type being asked for is the pointer that is not found,
```
abhinav: This code is now in `paramSimple.Build`.

dig.go
```diff
@@ -264,123 +271,6 @@ func (c *Container) isAcyclic(n *node) error {
...
-}
-
-// Returns a new In parent object with all the dependency fields
-// populated from the dig container.
-func (c *Container) createInObject(t reflect.Type) (reflect.Value, error) {
-	dest := reflect.New(t).Elem()
-
-	for i := 0; i < t.NumField(); i++ {
-		f := t.Field(i)
-
```
abhinav: This code is now in `paramObject.Build`.

dig.go
```diff
@@ -407,19 +297,6 @@ func (c *Container) remove(nodes []*node) {
...
-func (c *Container) constructorArgs(ctype reflect.Type) ([]reflect.Value, error) {
-	argTypes := getConstructorArgTypes(ctype)
-	args := make([]reflect.Value, 0, len(argTypes))
-	for _, t := range argTypes {
-		arg, err := c.get(edge{key: key{t: t}})
-		if err != nil {
-			return nil, errWrapf(err, "couldn't get arguments for constructor %v", ctype)
-		}
-		args = append(args, arg)
-	}
```
abhinav: This is now part of `paramList.BuildParams`.

dig.go
```diff
@@ -450,26 +321,6 @@ func newNode(k key, ctor interface{}, ctype reflect.Type) (*node, error) {
...
-// those values in yet.
-func getConstructorArgTypes(ctype reflect.Type) []reflect.Type {
-	numArgs := ctype.NumIn()
-	if ctype.IsVariadic() {
-		// NOTE: If the function is variadic, we skip the last argument
-		// because we're not filling variadic arguments yet. See #120.
-		numArgs--
-	}
-
-	args := make([]reflect.Type, numArgs)
```
abhinav: This is not needed anymore because we have already gathered this information in `newParamList`.

param.go
```diff
@@ -135,16 +143,111 @@ func newParamList(ctype reflect.Type) (paramList, error) {
...
+	args, err := n.Params.BuildParams(c)
+	if err != nil {
+		return _noValue, errWrapf(err, "couldn't get arguments for constructor %v", n.ctype)
+	}
+
+	constructed := reflect.ValueOf(n.ctor).Call(args)
+
+	// Provide-time validation ensures that all constructors return at least
+	// one value.
+	if errV := constructed[len(constructed)-1]; isError(errV.Type()) {
```
willhug: This will panic with a weird error message if I try to Provide an Invoke right (with no error)?  Or do we validate that provided constructors need to return _something_.  Or I guess it will just never be called, nbd.

param.go
```diff
@@ -135,16 +143,111 @@ func newParamList(ctype reflect.Type) (paramList, error) {
...
+	case IsOut(t) || (t.Kind() == reflect.Ptr && IsOut(t.Elem())) || embedsType(t, _outPtrType):
+		return nil, fmt.Errorf("cannot depend on result objects: %v embeds a dig.Out", t)
+	default:
+		return paramSimple{Type: t}, nil
 	}
-	return paramSimple{Type: t}, nil
+}
+
+func (sp paramSimple) Build(c *Container) (reflect.Value, error) {
+	k := key{name: sp.Name, t: sp.Type}
```
willhug: ordering here doesn't matter right?  We alternate the order on the "set"

abhinav: Yeah, the ordering is irrelevant here. We ensure that keys are unique across all constructors.

param.go
abhinav: Discussed offline but the comment above addresses this concern: There will always be at least one return value because we validate that earlier.

akshayjshah: FYI @willhug `Invoke` will happily run `func() (int, int, string)` - it just throws away all the returned values. :tada: It special-cases functions whose last return value is an error and interprets that error as a marker of invocation success/failure.

param.go
```diff
@@ -35,9 +35,17 @@ import (
 type (
 	param interface {
 		String() string
+
+		// Builds this dependency and any of its from the provided Container.
```
akshayjshah: Missing a word here - "and any of its ???"

param.go
```diff
@@ -129,16 +137,111 @@ func newParamList(ctype reflect.Type) (paramList, error) {
 		}
 		pl.Params = append(pl.Params, p)
 	}
+
 	return pl, nil
 }
 
+func (pl paramList) Build(*Container) (reflect.Value, error) {
+	panic("paramList.Build() must never be called")
```
akshayjshah: Same thing re: panic messages.

param.go
```diff
@@ -129,16 +137,111 @@ func newParamList(ctype reflect.Type) (paramList, error) {
...
+	case IsIn(t):
 		return newParamObject(t)
+	case embedsType(t, _inPtrType):
+		return nil, fmt.Errorf(
+			"%v embeds *dig.In which is not supported, embed dig.In value instead", t)
+	case t.Kind() == reflect.Ptr && IsIn(t.Elem()):
+		return nil, fmt.Errorf(
+			"dependency %v is a pointer to dig.In, use value type instead", t)
+	case IsOut(t) || (t.Kind() == reflect.Ptr && IsOut(t.Elem())) || embedsType(t, _outPtrType):
+		return nil, fmt.Errorf("cannot depend on result objects: %v embeds a dig.Out", t)
```
akshayjshah: Should this case be first? What do we expect to happen if a struct embeds both `dig.In` and `dig.Out`?

param.go
```diff
@@ -129,16 +137,111 @@ func newParamList(ctype reflect.Type) (paramList, error) {
...
-	if IsIn(t) {
+	switch {
+	case IsIn(t):
 		return newParamObject(t)
+	case embedsType(t, _inPtrType):
+		return nil, fmt.Errorf(
+			"%v embeds *dig.In which is not supported, embed dig.In value instead", t)
+	case t.Kind() == reflect.Ptr && IsIn(t.Elem()):
+		return nil, fmt.Errorf(
+			"dependency %v is a pointer to dig.In, use value type instead", t)
```
akshayjshah: I realize that you've just copied the existing error message, but this wording is really confusing to me - there's no pointer to `dig.In` involved here at all. Would this be better as `cannot depend on pointer to params struct, use a value instead`?

param.go
```diff
@@ -129,16 +137,111 @@ func newParamList(ctype reflect.Type) (paramList, error) {
...
+	k := key{name: sp.Name, t: sp.Type}
+	if v, ok := c.cache[k]; ok {
+		return v, nil
+	}
+
+	n, ok := c.nodes[k]
+	if !ok {
+		// Unlike in the fallback case below, if a user makes an error requesting
+		// a mixed type for an optional parameter, a good error message "did you mean X?"
+		// will not be used and dig will return zero value.
```
akshayjshah: What's a "mixed type" in this context? The only references I can Google are PHP-specific and don't make much sense for Go.

Do we even need a comment explaining that optional params are not required?

param.go
```diff
@@ -129,16 +137,111 @@ func newParamList(ctype reflect.Type) (paramList, error) {
...
+		if err, _ := errV.Interface().(error); err != nil {
+			return _noValue, errWrapf(err, "constructor %v for type %v failed", n.ctype, sp.Type)
+		}
+	}
+
+	for _, con := range constructed {
+		// Set the resolved object into the cache.
+		// This might look confusing at first like we're ignoring named types,
+		// but `con` in this case will be the dig.Out object, which will
+		// cause a recursion into the .set for each of it's memebers.
```
akshayjshah: "members" is misspelled.

param.go
abhinav: Oh, good catch. Moved to front.

param.go
abhinav: Not entirely sure. This was copied verbatim. Reworded to make more sense.

param.go
abhinav: Fixed.

param.go
glibsm: requesting *foo when foo is available
requesting foo when *foo is available is what i labeled "mixed type".

If you can figure out a better verbage - go for it. I couldn't

glibsm:
`Invoke will happily run func() (int, int, string)`

I'm not sure we should encourage that, since it's an implementation detail that can change right? It's not enforced in the API.

A very interesting talk on the subject: https://www.youtube.com/watch?v=OuT8YYAOOVI

akshayjshah:
I definitely wouldn't encourage it, but I don't think we can get any more restrictive in the 1.x Fx release series.

PR #156, opened 2017-10-02
Provide: Inspect constructor results upfront

abhinav:
Similar to #149, we now inspect constructor results upfront. These are
represented by the `result` interface which unionizes the following
types,

- resultList: All return values of a constructor
- resultSingle: Basic (optionally named) value
- resultObject: `dig.Out` object whose fields are other `result`s
- resultError: Sentinel type to indicate that an error is returned here.

The `result` interface has a `Produces()` method which returns a
comprehensive set of all values produces by the `result`.

### DIFF
--- a/result.go
+++ b/result.go
@@ -38,18 +38,20 @@ var (
 
 func newResult(t reflect.Type) (result, error) {
 	switch {
+	case IsIn(t) || (t.Kind() == reflect.Ptr && IsIn(t.Elem())) || embedsType(t, _inPtrType):
+		return nil, fmt.Errorf("cannot provide parameter objects: %v embeds a dig.In", t)
 	case isError(t):
 		return resultError{}, nil
 	case IsOut(t):
 		return newResultObject(t)
 	case embedsType(t, _outPtrType):
 		return nil, fmt.Errorf(
-			"%v embeds *dig.Out which is not supported, embed dig.Out value instead", t)
+			"cannot build a result object by embedding *dig.Out, embed dig.Out instead: "+
+				"%v embeds *dig.Out", t)
 	case t.Kind() == reflect.Ptr && IsOut(t.Elem()):
-		return nil, fmt.Errorf("%v is a pointer to dig.Out, use value type instead", t)
-		// Make sure we're not producing dig.In's either.
-	case IsIn(t) || (t.Kind() == reflect.Ptr && IsIn(t.Elem())) || embedsType(t, _inPtrType):
-		return nil, fmt.Errorf("cannot provide parameter objects: %v embeds a dig.In", t)
+		return nil, fmt.Errorf(
+			"cannot return a pointer to a result object, use a value instead: "+
+				"%v is a pointer to a struct that embeds dig.Out", t)
 	default:
 		return resultSingle{Type: t}, nil
 	}
--- a/result.go
+++ b/result.go
@@ -25,6 +25,14 @@ import (
 	"reflect"
 )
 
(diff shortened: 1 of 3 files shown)


result.go
```diff
@@ -0,0 +1,177 @@
...
+	case IsOut(t):
+		return newResultObject(t)
+	case embedsType(t, _outPtrType):
+		return nil, fmt.Errorf(
+			"%v embeds *dig.Out which is not supported, embed dig.Out value instead", t)
+	case t.Kind() == reflect.Ptr && IsOut(t.Elem()):
+		return nil, fmt.Errorf("%v is a pointer to dig.Out, use value type instead", t)
+		// Make sure we're not producing dig.In's either.
+	case IsIn(t) || (t.Kind() == reflect.Ptr && IsIn(t.Elem())) || embedsType(t, _inPtrType):
+		return nil, fmt.Errorf("cannot provide parameter objects: %v embeds a dig.In", t)
```
akshayjshah: Ordering again - what do we expect to happen if a struct embeds both `dig.In` and `dig.Out`?

result.go
```diff
@@ -0,0 +1,177 @@
...
+	switch {
+	case isError(t):
+		return resultError{}, nil
+	case IsOut(t):
+		return newResultObject(t)
+	case embedsType(t, _outPtrType):
+		return nil, fmt.Errorf(
+			"%v embeds *dig.Out which is not supported, embed dig.Out value instead", t)
+	case t.Kind() == reflect.Ptr && IsOut(t.Elem()):
+		return nil, fmt.Errorf("%v is a pointer to dig.Out, use value type instead", t)
```
akshayjshah: Same as the params struct comment on a previous PR - I'm finding this confusing because there's no pointer to `dig.Out` involved here. We're complaining because we want results structs to be values.

result.go
```diff
@@ -0,0 +1,177 @@
...
+// THE SOFTWARE.
+
+package dig
+
+import (
+	"fmt"
+	"reflect"
+)
+
+type (
```
akshayjshah: The ordering in this file is really confusing me. In the `params` file, I remember seeing all type defs, then all constructors, then all methods (grouped so that all implementations of the same method are together). Can we either do the same here or stick to the usual type def-constructor-methods order?

result.go
```diff
@@ -0,0 +1,177 @@
...
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		if f.Type == _outType {
+			// Skip over the dig.Out embed.
+			continue
+		}
+
+		if f.PkgPath != "" {
+			return ro, fmt.Errorf(
+				"private fields not allowed in dig.Out, did you mean to export %q (%v) from %v?",
```
akshayjshah: Can we use Go idioms here and replace "private" with "unexported"?

result.go
```diff
@@ -0,0 +1,177 @@
...
+				f.Name, f.Type, t)
+		}
+
+		if isError(f.Type) {
+			return ro, fmt.Errorf(
+				"cannot provide errors from dig.Out: field %q (%v) of %v is an error field",
```
akshayjshah: It'd be nice to suggest an alternative here - the user probably (maybe?) should be writing a `func() (ResultStruct, error)` instead.

Exact wording up to you.

result.go
abhinav: Switched to the usual in both places.

result.go
abhinav: Switched to,

    cannot return errors from dig.Out, return it from the constructor instead:

PR #158, opened 2017-10-02
Feed results into container using `result`s

abhinav:
This changes how values are fed into the container to rely on individual
`result` implementations.

Instead of a single place where we handle the different return values of
a constructor, each `result` type now extracts information it cares
about.

- resultList feeds individual values into the underlying results
- resultSingle stores the value into the container cache
- resultError retrieves and returns the error as-is
- resultObject extracts corresponding dig.Out fields and delegates to
  the underlying results

With this change, both the input and output paths have been refactored
to use this approach.

### DIFF
--- a/result.go
+++ b/result.go
@@ -36,6 +36,8 @@ import (
 type result interface {
 	// Extracts the values for this result from the provided value and
 	// stores them in the container.
+	//
+	// This MAY panic if the result does not consume a single value.
 	Extract(*Container, reflect.Value) error
 
 	Produces() map[key]struct{}
@@ -110,7 +112,10 @@ func newResultList(ctype reflect.Type) (resultList, error) {
 func (rl resultList) Produces() map[key]struct{} { return rl.produces }
 
 func (resultList) Extract(*Container, reflect.Value) error {
-	panic("resultList.Extract() must never be called")
+	panic("It looks like you have found a bug in dig. " +
+		"Please file an issue at https://github.com/uber-go/dig/issues/ " +
+		"and provide the following message: " +
+		"resultList.Extract() must never be called")
 }
 
 func (rl resultList) ExtractResults(c *Container, values []reflect.Value) error {
--- a/result.go
+++ b/result.go
@@ -118,7 +118,7 @@ func (resultList) Extract(*Container, reflect.Value) error {
 		"resultList.Extract() must never be called")
 }
 
-func (rl resultList) ExtractResults(c *Container, values []reflect.Value) error {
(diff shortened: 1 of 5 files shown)


result.go
```diff
@@ -175,3 +202,14 @@ func newResultObject(t reflect.Type) (resultObject, error) {
 }
 
 func (ro resultObject) Produces() map[key]struct{} { return ro.produces }
+
+func (ro resultObject) Extract(c *Container, v reflect.Value) error {
+	for _, f := range ro.Fields {
+		if err := f.Result.Extract(c, v.Field(f.Index)); err != nil {
+			// In reality, this will never fail because none of the fields of
+			// a resultObject can be resultError.
```
abhinav: To be clear, this is not possible to reproduce in a test because it requires creating an invalid resultObject. Yes, I could hand-build an invalid resultObject but that sort-of defeats the purpose.

I'm not sure if we have agreement around how to handle testing situations like this.

result.go
```diff
@@ -27,6 +27,10 @@ import (
 
 type (
 	result interface {
+		// Extracts the values for this result from the provided value and
+		// stores them in the container.
```
akshayjshah: For symmetry with `params`, should we document that this may panic?

result.go
```diff
@@ -96,6 +100,19 @@ func newResultList(ctype reflect.Type) (resultList, error) {
 
 func (rl resultList) Produces() map[key]struct{} { return rl.produces }
 
+func (resultList) Extract(*Container, reflect.Value) error {
+	panic("resultList.Extract() must never be called")
```
akshayjshah: Same note re: panics and user-facing error messages.

result.go
```diff
@@ -96,6 +100,19 @@ func newResultList(ctype reflect.Type) (resultList, error) {
...
+}
+
+func (rl resultList) ExtractResults(c *Container, values []reflect.Value) error {
```
akshayjshah: `ExtractAll` or `ExtractList`? These are all `result` implementations, so `Extract` and `ExtractResults` aren't obviously different to me.

result.go
akshayjshah: This doesn't seem any different from the unreachable cases in the type switches we've introduced; since we've already introduced a precedent for it, we can panic if this code is really unreachable.

akshayjshah:
What do we think about relaxing the codecov configuration? I'm all for keeping coverage high and asking why we're not covering individual lines, but setting the project-wide hard failure threshold at 95% seems sufficient to me. Since we're already meeting that bar, I don't think we need a per-diff coverage requirement.

PR #160, opened 2017-10-09
param: Replace forEachParamSingle with visitor

abhinav:
This drops forEachParamSingle in favor of a go/ast-style visitor. The
visitor is called on a param, and if it returns a non-nil visitor, the
returned visitor is called on the children of that param.

This allows us to replace param switches which just want to traverse the
tree. This also lets us ensure that we have only one place in the code
where we panic with "unknown param type".


dig_test.go
```diff
@@ -1009,6 +1009,7 @@ func TestProvideCycleFails(t *testing.T) {
 			In
 
 			B B
+			W io.Writer
```
willhug: Accident or purposeful?

abhinav: Purposeful. Covers a path not previously covered.

PR #161, opened 2017-10-09
result: Find results with resultVisitor

abhinav:
Similar to paramVisitor, this adds resultVisitor and walkResult. In
addition to visiting results, resultVisitor also knows which result
position or result object field it is visiting, allowing for more
detailed error messages.

For example, this lets us provide a lot more information in the case
when the constructor is providing the same type multiple times. Given,

    type out struct {
        dig.Out

        Foo struct {
            dig.Out

            Writer io.Writer
        }

        Bar struct {
            dig.Out

            Writer io.Writer
        }
    }

Instead of,

   returns multiple io.Writer

We will now get,

    cannot provide io.Writer from [0].Bar.Writer in constructor func() out:
        already provided by [0].Foo.Writer

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -156,7 +156,7 @@ func (c *Container) provide(ctor interface{}, ctype reflect.Type) error {
 		return err
 	}
 
-	keys, err := c.findConnections(n)
+	keys, err := c.findAndValidateResults(n)
 	if err != nil {
 		return err
 	}
@@ -176,14 +176,15 @@ func (c *Container) provide(ctor interface{}, ctype reflect.Type) error {
 	return nil
 }
 
-func (c *Container) findConnections(n *node) (map[key]struct{}, error) {
+// Builds a collection of all result types produced by this node.
+func (c *Container) findAndValidateResults(n *node) (map[key]struct{}, error) {
 	var err error
 	keyPaths := make(map[key]string)
 	walkResult(n.Results, connectionVisitor{
-		c:    c,
-		n:    n,
-		err:  &err,
-		keys: keyPaths,
+		c:        c,
+		n:        n,
+		err:      &err,
+		keyPaths: keyPaths,
 	})
 
 	if err != nil {
--- a/dig.go
+++ b/dig.go
@@ -232,12 +232,12 @@ type connectionVisitor struct {
 	currentResultPath []string
 }
 
-func (cv connectionVisitor) VisitField(f resultObjectField) resultVisitor {
+func (cv connectionVisitor) AnnotateWithField(f resultObjectField) resultVisitor {
 	cv.currentResultPath = append(cv.currentResultPath, f.FieldName)
 	return cv
 }
 
-func (cv connectionVisitor) VisitPosition(i int) resultVisitor {
+func (cv connectionVisitor) AnnotateWithPosition(i int) resultVisitor {
 	cv.currentResultPath = append(cv.currentResultPath, fmt.Sprintf("[%d]", i))
 	return cv
 }
--- a/dig.go
+++ b/dig.go
@@ -198,21 +198,47 @@ func (c *Container) findAndValidateResults(n *node) (map[key]struct{}, error) {
 	return keys, nil
 }
 
+// Visits the results of a node and compiles a collection of all the keys
+// produced by that node.
 type connectionVisitor struct {
-	c    *Container
-	n    *node
-	err  *error
-	keys map[key]string
-	path []string
+	c *Container
+	n *node
+
+	// If this points to a non-nil value, we've already encountered an error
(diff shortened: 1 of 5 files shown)


result_test.go
```diff
@@ -170,34 +138,6 @@ func TestNewResultObjectErrors(t *testing.T) {
...
-		{
-			desc: "name conflict",
-			give: struct {
-				Out
-
-				Reader        io.Reader
-				NamedWriter   io.Writer `name:"what"`
-				AnotherWriter io.Writer `name:"what"`
-			}{},
-			err: "returns multiple io.Writer:what",
```
glibsm: why is this case deleted?

abhinav: This validation is no longer done in newResult. Instead it's done in connectionVisitor.
Plus these error paths are already exercised by the integration tests.

result.go
```diff
@@ -72,19 +70,89 @@ func newResult(t reflect.Type) (result, error) {
...
+		w := v
+		for i, r := range res.Results {
+			if v := w.VisitPosition(i); v != nil {
+				walkResult(r, v)
+			}
+		}
+	default:
+		panic(fmt.Sprintf(
+			"It looks like you have found a bug in dig. "+
+				"Please file an issue at https://github.com/uber-go/dig/issues/ "+
```
willhug: not sure if indentation is intentional...

dig.go
```diff
@@ -170,6 +176,78 @@ func (c *Container) provide(ctor interface{}, ctype reflect.Type) error {
 	return nil
 }
 
+func (c *Container) findConnections(n *node) (map[key]struct{}, error) {
```
willhug: had to keep reminding myself what this function did during the course of this review.  Maybe a rename to something like `findAndValidateResults`?

dig.go
```diff
@@ -170,6 +176,78 @@ func (c *Container) provide(ctor interface{}, ctype reflect.Type) error {
...
+	}
+	return keys, nil
+}
+
+type connectionVisitor struct {
+	c    *Container
+	n    *node
+	err  *error
+	keys map[key]string
+	path []string
```
willhug: path can be ambiguous (I started thinking "import path").  Maybe `structLocation`, `currentAttributeIdentifier`? Something like that.

dig.go
```diff
@@ -170,6 +176,78 @@ func (c *Container) provide(ctor interface{}, ctype reflect.Type) error {
...
+		keys[k] = struct{}{}
+	}
+	return keys, nil
+}
+
+type connectionVisitor struct {
+	c    *Container
+	n    *node
+	err  *error
+	keys map[key]string
```
willhug: does not infer that this is a map from "key" to "path" (or whatever we decide to name path).  Might be easier for readability to change this to something like "keysToIdentifier" or something.

result.go
```diff
@@ -72,19 +70,89 @@ func newResult(t reflect.Type) (result, error) {
...
+
+	// VisitField is called on each field of a resultObject after visiting it
+	// but before walking its descendants.
+	//
+	// The same resultVisitor is used for all fields: the one returned upon
+	// visiting the resultObject.
+	//
+	// For each visited field, if VisitField returns a non-nil resultVisitor,
+	// it will be used to walk the result of that field.
+	VisitField(resultObjectField) resultVisitor
```
willhug: I almost want to call these "AnnotateWithField" and "AnnotateWithPosition" because they're not reaaaaaally 'visiting' those fields for any other purpose.  Especially VisitPosition, there's not much you could do with it other than annotate (since it's just an index).  If we did something with it we'd have to keep state in the visitor object itself which could be the grounds for lots of confusion...

dig.go
```diff
@@ -170,6 +176,78 @@ func (c *Container) provide(ctor interface{}, ctype reflect.Type) error {
...
+		n:    n,
+		err:  &err,
+		keys: keyPaths,
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	keys := make(map[key]struct{}, len(keyPaths))
```
willhug: we throw this away immediately,  should we just return the keyPaths?

abhinav: They have different purposes. This is the set that we're expecting in the caller and `keyPaths` is used for tracking by the connectionVisitor. I'll document the individual fields of connectionVisitor further.

dig.go
abhinav: Renamed to `currentResultPath` and documented that it's the path to the 'result' being visited.

result.go
abhinav: Good call. Renaming.

result.go
abhinav: gofmt insists on indenting that

PR #162, opened 2017-10-09
node: Track if already called

abhinav:
Now that nodes represent individual constructors in the graph rather
than the types they provide, we can track on the node whether it was
already called.

This has no value right now but we will need this once we introduce
value groups to ensure that constructors don't get called multiple
times.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -279,9 +279,13 @@ func (c *Container) isAcyclic(p param, k key) error {
 	return detectCycles(p, c.nodes, []key{k})
 }
 
+// node represents a single constructor in the graph.
 type node struct {
 	ctor  interface{}
 	ctype reflect.Type
+
+	// Whether this constructor was already run.
+	called bool
 
 	// Type information about constructor parameters.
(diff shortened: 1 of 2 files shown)


glibsm:
add a test?

dig.go
```diff
@@ -320,6 +328,7 @@ func (n *node) Call(c *Container) error {
 		return errWrapf(err, "constructor %v failed", n.ctype)
 	}
 
+	n.called = true
```
kriskowal: Consider moving this up to avoid multiple attempts to call an uncallable function.

abhinav: I originally agreed with this and made the change but I realized that then
we'll have to remember the failure if the function failed. IMO we might as
well keep calling the function if it's failing.

Trying to test that did expose a little bug: a failing function leaves the
container in a bad state: it remembers the original value returned by the
constructor even if it failed.

Fixing and adding test.

abhinav: #166

PR #163, opened 2017-10-09
graph: Support multiple providers

abhinav:
This changes how the graph is represented internally from storing a
`map[key]*node` to `map[key][]*node`. This essentially allows a key to
have a dependency on multiple nodes (constructors).

As part of this change, I also renamed this dependency map to
`providers` and the store for values from `cache` to `values`, which I
think is clearer.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -60,24 +60,18 @@ type InvokeOption interface {
 
 // Container is a directed acyclic graph of types and their dependencies.
 type Container struct {
-	nodes map[key]*node
-	cache map[key]reflect.Value
+	// Constructors for different values.
+	producers map[key][]*node
 
-	// TODO: for advanced use-case, add an index
-	// This will allow retrieval of a single type, without specifying the exact
-	// tag, provided there is only one object of that given type
-	//
-	// It will also allow library owners to create a "default" tag for their
-	// object, in case users want to provide another type with a different name
-	//
-	// index map[reflect.Type]key
+	// Values that have already been generated in the container.
+	values map[key]reflect.Value
 }
 
 // New constructs a Container.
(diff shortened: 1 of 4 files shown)


dig.go
```diff
@@ -60,24 +60,18 @@ type InvokeOption interface {
...
-	// TODO: for advanced use-case, add an index
-	// This will allow retrieval of a single type, without specifying the exact
-	// tag, provided there is only one object of that given type
-	//
-	// It will also allow library owners to create a "default" tag for their
-	// object, in case users want to provide another type with a different name
-	//
-	// index map[reflect.Type]key
+	// Constructors for different values.
+	producers map[key][]*node
```
glibsm: I find it odd that something that stores `constructors` (godoc above), and keyed via `key -> *node` is called `producers`.

abhinav: The reason I called them producers instead of constructors is that constructor
almost implies one constructor per value type, but with this new structure we
can have multiple constructors which produce the same type of value.

So it made more sense to me to call them producers since they produce values.
I'll update the docs to reflect this a bit more.

dig.go
```diff
@@ -60,24 +60,19 @@ type InvokeOption interface {
 
 // Container is a directed acyclic graph of types and their dependencies.
 type Container struct {
-	nodes map[key]*node
-	cache map[key]reflect.Value
+	// Mapping from key to all the nodes that can produce a value for that
+	// key.
+	producers map[key][]*node
```
kriskowal: Is `providers` a more fitting name here? It might improve readability to make "provider" and "producer" the same or obviously different.

abhinav: I like providers. SG

PR #164, opened 2017-10-09
Add support for value groups

### LINKED ISSUE #127: Allow "fan-in" for named types in dig graph
I've been talking to @abhinav about this, In order to support use cases where a dig parameter depends on multiple inputs of the same type we need the ability to aggregate dig types together.

So if we have the following outputs:

```go
  Output1 MyType `name:"output1"`
  Output2 MyType `name:"output2"`
```

I'd want to be able to grab all the "MyType"s in a single attribute.

### 1: Add another tag to indicate that the type should be a fan-in collection:

```go
type In struct {
  dig.In
  Outputs map[string]MyType `collection:"true"` // tag name TBD
}
```
The `string` key in the `Outputs` variable will be the name of the attributes that were injected into the graph.

### 2: Add a `dig.Key` (or similar) type for map keys.  Any map of these types will be recognized as a fan-in collection.

```go
type In struct {
  dig.In
  Outputs map[dig.Key]MyType
}
```
The dig.Key would be a dig struct where we could embed information on the type.  Currently that would only be the "name", but we could add more information later.

### YARPC Middleware

Currently we don't have a good answer for middleware to ensure that we can take multiple middleware inputs and effectively construct a middleware graph without forcing the yarpcfx module to have a hard dependency on every middleware library (which gets out of control).  With this system we'd be able to grab all middleware that was inserted into the graph.  Since this is a map, the yarpcfx module would need to have an ordered list of middleware names it was expecting beforehand (all names should be optional) and it can construct the middleware chain itself.  Of course this would only support names that yarpcfx explicitly expects, but that is necessary for the middleware ordering.

### YARPC Specs:  

The yarpc config options allow inserting custom Transport/PeerList/Updater specs into the yarpc config builder.  Having support for this would allow us to detach the yarpc fx module from the implementations of each transport.  For instance, a user could inject a "grpcfx" module into the graph which provided a "TransportSpec" for grpc that integrated seamlessly into the yarpc configuration.

abhinav:
This adds support for value groups to dig. In addition to supporting
named values with `dig.In/Out`, we now supported unordered groups of
named values.

To produce a named value, a constructor will tag a `dig.Out` field with
`group:".."`. Constructors which rely on value groups can consume these
values by declaring a slice tagged with `group:".."`.

For example, consider the YARPC TransportSpec use case. The constructor
that produces the TransportSpec will do,

    type TransportSpecResult struct {
        dig.Out

        TransportSpec config.TransportSpec `group:"yarpc.transports"`
    }

    func NewTransportSpec() TransportSpecResult

The constructor that consumes the TransportSpecs will do,

    type Param struct {
        dig.In

        TransportSpecs []config.TransportSpec `group:"yarpc.transports"`
    }

    func NewThing(Param) *Thing

Resolves #127

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -429,7 +429,7 @@ func detectCycles(par param, graph map[key][]*node, path []key) error {
 }
 
 // Checks if a field of an In struct is optional.
-func isFieldOptional(parent reflect.Type, f reflect.StructField) (bool, error) {
+func isFieldOptional(f reflect.StructField) (bool, error) {
 	tag := f.Tag.Get(_optionalTag)
 	if tag == "" {
 		return false, nil
@@ -438,8 +438,8 @@ func isFieldOptional(parent reflect.Type, f reflect.StructField) (bool, error) {
 	optional, err := strconv.ParseBool(tag)
 	if err != nil {
 		err = errWrapf(err,
-			"invalid value %q for %q tag on field %v of %v",
-			tag, _optionalTag, f.Name, parent)
+			"invalid value %q for %q tag on field %v",
+			tag, _optionalTag, f.Name)
 	}
 
 	return optional, err
--- a/dig.go
+++ b/dig.go
@@ -288,10 +288,15 @@ func (cv connectionVisitor) Visit(res result) resultVisitor {
 			return nil
 		}
 
-		if _, ok := cv.c.providers[k]; ok {
+		if ps := cv.c.providers[k]; len(ps) > 0 {
+			csigs := make([]string, len(ps))
+			for i, p := range ps {
+				csigs[i] = fmt.Sprint(p.ctype)
+			}
+
 			*cv.err = fmt.Errorf(
-				"cannot provide %v from %v in constructor %v: already in the container",
-				k, path, cv.n.ctype)
+				"cannot provide %v from %v in constructor %v: already provided by %v",
+				k, path, cv.n.ctype, csigs)
 			return nil
 		}
 
--- a/dig.go
+++ b/dig.go
@@ -24,9 +24,11 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"math/rand"
 	"reflect"
 	"strconv"
 	"strings"
+	"time"
 )
 
 const (
@@ -45,8 +47,12 @@ type key struct {
 // Option configures a Container. It's included for future functionality;
 // currently, there are no concrete implementations.
 type Option interface {
-	unimplemented()
+	applyOption(*Container)
 }
+
+type optionFunc func(*Container)
+
+func (f optionFunc) applyOption(c *Container) { f(c) }
 
 // A ProvideOption modifies the default behavior of Provide. It's included for
 // future functionality; currently, there are no concrete implementations.
@@ -71,15 +77,33 @@ type Container struct {
 
 	// Values groups that have already been generated in the container.
 	groups map[key][]reflect.Value
+
+	// Source of randomness.
+	rand *rand.Rand
 }
 
 // New constructs a Container.
 func New(opts ...Option) *Container {
-	return &Container{
+	c := &Container{
 		providers: make(map[key][]*node),
 		values:    make(map[key]reflect.Value),
 		groups:    make(map[key][]reflect.Value),
+		rand:      rand.New(rand.NewSource(time.Now().UnixNano())),
 	}
+
+	for _, opt := range opts {
+		opt.applyOption(c)
+	}
+	return c
+}
+
+// Changes the source of randomness for the container.
+//
+// This will help provide determinism during tests.
+func setRand(r *rand.Rand) Option {
+	return optionFunc(func(c *Container) {
+		c.rand = r
+	})
 }
 
 // Provide teaches the container how to build values of one or more types and
--- a/dig.go
+++ b/dig.go
@@ -472,10 +472,14 @@ func shallowCheckDependencies(c *Container, p param) error {
 type stagingReceiver struct {
 	err    error
 	values map[key]reflect.Value
+	groups map[key][]reflect.Value
 }
 
 func newStagingReceiver() *stagingReceiver {
-	return &stagingReceiver{values: make(map[key]reflect.Value)}
+	return &stagingReceiver{
+		values: make(map[key]reflect.Value),
+		groups: make(map[key][]reflect.Value),
+	}
 }
 
 func (sr *stagingReceiver) SubmitError(err error) {
@@ -489,6 +493,11 @@ func (sr *stagingReceiver) SubmitValue(name string, t reflect.Type, v reflect.Va
 	sr.values[key{t: t, name: name}] = v
 }
 
+func (sr *stagingReceiver) SubmitGroupValue(group string, t reflect.Type, v reflect.Value) {
+	k := key{t: t, group: group}
+	sr.groups[k] = append(sr.groups[k], v)
+}
+
 // Commit commits the received results to the provided container.
 //
 // If the resultReceiver failed, no changes are committed to the container.
@@ -499,6 +508,10 @@ func (sr *stagingReceiver) Commit(c *Container) error {
 
 	for k, v := range sr.values {
 		c.values[k] = v
+	}
+
+	for k, vs := range sr.groups {
+		c.groups[k] = append(c.groups[k], vs...)
 	}
 
 	return nil
--- a/dig.go
+++ b/dig.go
@@ -21,7 +21,6 @@
 package dig
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"math/rand"
@@ -382,12 +381,12 @@ type errCycleDetected struct {
 }
 
 func (e errCycleDetected) Error() string {
-	b := new(bytes.Buffer)
-	for _, k := range e.Path {
-		fmt.Fprintf(b, "%v ->", k.t)
+	items := make([]string, len(e.Path)+1)
+	for i, k := range e.Path {
+		items[i] = fmt.Sprint(k)
 	}
-	fmt.Fprintf(b, "%v", e.Key.t)
-	return b.String()
+	items[len(items)-1] = fmt.Sprint(e.Key)
+	return strings.Join(items, " -> ")
 }
 
 func detectCycles(par param, graph map[key][]*node, path []key) error {
--- a/dig.go
+++ b/dig.go
@@ -39,7 +39,9 @@ const (
 
 // Unique identification of an object in the graph.
 type key struct {
-	t     reflect.Type
+	t reflect.Type
+
+	// Only one of name or group will be set.
 	name  string
 	group string
 }
--- a/dig.go
+++ b/dig.go
@@ -385,7 +385,7 @@ func (e errCycleDetected) Error() string {
 	for i, k := range e.Path {
 		items[i] = fmt.Sprint(k)
 	}
-	items[len(items)-1] = fmt.Sprint(e.Key)
+	items[len(e.Path)] = fmt.Sprint(e.Key)
 	return strings.Join(items, " -> ")
 }
 
--- a/dig.go
+++ b/dig.go
@@ -32,12 +32,14 @@ import (
 const (
 	_optionalTag = "optional"
 	_nameTag     = "name"
+	_groupTag    = "group"
 )
 
 // Unique identification of an object in the graph.
 type key struct {
-	t    reflect.Type
-	name string
+	t     reflect.Type
+	name  string
+	group string
 }
 
 // Option configures a Container. It's included for future functionality;
@@ -66,6 +68,9 @@ type Container struct {
 
 	// Values that have already been generated in the container.
 	values map[key]reflect.Value
+
+	// Values groups that have already been generated in the container.
+	groups map[key][]reflect.Value
 }
 
 // New constructs a Container.
@@ -73,6 +78,7 @@ func New(opts ...Option) *Container {
 	return &Container{
 		providers: make(map[key][]*node),
 		values:    make(map[key]reflect.Value),
+		groups:    make(map[key][]reflect.Value),
 	}
 }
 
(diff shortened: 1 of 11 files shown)


abhinav:
Your interpretation is correct. That is indeed a limitation of the current
system.

We can support flattening or we can let consumers deal with the complexity.
For example,

    type TransportSpecResult struct {
        dig.Out
        
        Specs []config.TransportSpec `group:"yarpc.transports"`
    }

    type Params struct {
        dig.In
        
        Specs [][]config.TransportSpec `group:"yarpc.transports"`
    }
    
And then the constructor consuming Params will flatten the slice into a
`[]config.TransportSpec`.

breerly:
Flattening is nice, but AFAICT we can add that later. (+1 @abhinav)

glibsm:
I wasn't suggesting that it should be added immediately. I was trying to gauge where this PR Is headed

breerly:
I think it's actually a pretty compelling addition, but considering everything we add here is basically permanent, I'd like to see the use-cases crop in the wild first :)

dig_test.go
```diff
@@ -782,6 +782,57 @@ func TestEndToEndSuccess(t *testing.T) {
 		assert.Contains(t, err.Error(), "B isn't in the container")
 	})
 
+	t.Run("value groups", func(t *testing.T) {
```
breerly: I like this test, but can we add a dumber one that more plainly illustrates the feature?

param.go
```diff
@@ -326,3 +342,52 @@ func (po paramObject) Build(c *Container) (reflect.Value, error) {
...
+	return paramGroupedSlice{Group: group, Type: t}
+}
+
+func (pt paramGroupedSlice) Build(c *Container) (reflect.Value, error) {
+	k := key{group: pt.Group, t: pt.Type.Elem()}
+
+	for _, n := range c.producers[k] {
+		if err := n.Call(c); err != nil {
+			// TODO: better message
+			return _noValue, errWrapf(err, "failed to build %v", k)
```
breerly: What were you thinking here, in terms of a better message?

param.go
```diff
@@ -326,3 +342,52 @@ func (po paramObject) Build(c *Container) (reflect.Value, error) {
...
+	result := reflect.MakeSlice(pt.Type, len(items), len(items))
+	for i, v := range items {
+		result.Index(i).Set(v)
+	}
+	return result, nil
+}
+
+func shuffledCopy(items []reflect.Value) []reflect.Value {
+	newItems := make([]reflect.Value, len(items))
+	// TODO(abg): Inject rand to get some determinism during tests.
```
breerly: good idea!

kriskowal: We can get randomness implicitly if we track the values internally with a map[T]struct{}, translating to and from slices at the edges.

breerly: Thinking out loud - should we favor explicitness in this case, since it’s all transparent to the anyways?

dig.go
```diff
@@ -66,13 +68,17 @@ type Container struct {
 
 	// Values that have already been generated in the container.
 	values map[key]reflect.Value
+
+	// Values groups that have already been generated in the container.
+	groups map[key][]reflect.Value
```
kriskowal: I recommend map[key]map[reflect.Value]struct{} here.

abhinav: I dislike the reliance that would put on the undefined hash behavior of
`reflect.Value`.

See also, https://play.golang.org/p/rXBlkDiBCq. Two `reflect.Value`s
containing the same value don't actually hash to the same thing.
(Although in this case, that's probably what we want since two separate
constructors should be able to produce the same value if they were
provided separately.)

But generally, it seems safer and more obvious to me to shuffle by hand.

dig_test.go
abhinav: This test is just a placeholder to demonstrate this feature for now.

param.go
abhinav: I think we can track more information and include it here. I haven't had a chance to fully explore how much or how little information should go here.

dig.go
```diff
@@ -246,28 +252,34 @@ func (cv connectionVisitor) Visit(res result) resultVisitor {
...
-	if _, ok := cv.c.producers[k]; ok {
-		*cv.err = fmt.Errorf(
-			"cannot provide %v from %v in constructor %v: already in the container",
-			k, path, cv.n.ctype)
-		return nil
+		cv.keyPaths[k] = path
+
+	case resultGrouped:
+		// we don't really care about the path for this since conflicts are
+		// okay for group results. We'll track it for the sake of having a
```
peter-edge: Or not? lol

param.go
```diff
@@ -288,22 +293,33 @@ func newParamObject(t reflect.Type) (paramObject, error) {
 				f.Name, f.Type, t)
 		}
 
-		p, err := newParam(f.Type)
-		if err != nil {
-			return po, errWrapf(err, "bad field %q of %v", f.Name, t)
-		}
+		var p param
+		if group := f.Tag.Get(_groupTag); group != "" {
+			// TODO(abg): Verify that a name or optional was not specified.
```
peter-edge: Do it

param.go
```diff
@@ -288,22 +293,33 @@ func newParamObject(t reflect.Type) (paramObject, error) {
...
+			if f.Type.Kind() != reflect.Slice {
+				return po, fmt.Errorf("value groups may be consumed as slices only: "+
+					"field %q (%T) of %v is not a slice", f.Name, f.Type, t)
+			}
+			p = newParamGroupedSlice(group, f.Type)
+		} else {
+			var err error
+			p, err = newParam(f.Type)
+			if err != nil {
+				return po, errWrapf(err, "bad field %q of %v", f.Name, t)
```
peter-edge: Why is errWrapf used here but fmt.Errorf used above?

param.go
```diff
@@ -326,3 +342,52 @@ func (po paramObject) Build(c *Container) (reflect.Value, error) {
...
+	for _, n := range c.producers[k] {
+		if err := n.Call(c); err != nil {
+			// TODO: better message
+			return _noValue, errWrapf(err, "failed to build %v", k)
+		}
+	}
+
+	items := c.groups[k]
+
+	// shuffle the list so users don't rely on the ordering of grouped values
```
peter-edge: ehhhhhhhhhhhh idk, does this really do anything? maybe people aren't allowed to rely on the ordering, but there's some optimization they take into account when things are "ordered" in the way they generally expect? Think golang maps, which aren't ordered, but when you iterate the iteration order is generally the insertion order. It just feels like shuffling is eh.

param.go
abhinav: errWrapf is used only in places where there's an `err` to wrap. It relies on some machinery defined in error.go to make sure we track the root cause and expose it via `dig.RootCause`.

param.go
abhinav: This is very deliberate. Groups are unordered so you don't get to rely even on the implicit ordering.

Also, Go map iteration order is not insertion order. It's intentionally randomized by the runtime.

https://play.golang.org/p/x1Covh7eIv
https://nathanleclaire.com/blog/2014/04/27/a-surprising-feature-of-golang-that-colored-me-impressed/

akshayjshah: Go map ordering is explicitly randomized. https://play.golang.org/p/tKEWlW8Fiv

param.go
akshayjshah: Is this the behavior we want? This will get difficult as modules provide e.g., middleware, that depends on types that may or may not be present. Should we default to skipping failed constructors, or should we support optional groups?

abhinav: Discussed offline with @kriskowal: Yeah, this is desirable and largely
unavoidable with a static dependency graph.

abhinav:
@prashantv @kriskowal @akshayjshah @glibsm 

This is ready. Please review thoroughly. Thanks!

param.go
kriskowal: This mitigates a common problem in C. You’ll often only test short-lived programs and use pointers as hashmap keys. The first run will tend to lay the map out sequentially as malloc pioneers the heap. It’ll work well enough on the first run. You’ll discover new behaviors in production (too late) when the heap turns to Swiss cheese.

prashantv:
I haven't done a full review, but the overall code looks fine. One question -- are we ensuring that the same name can't be used in a name and as a group?

E.g., I don't think there's a good use case for:
```go
type Result struct {
  dig.Out

  Name string `name:"yarpc"`
  Names []string `group:"yarpc"`
}
```

Thoughts?

abhinav:
@prashantv Thanks for looking. There's no reason to disallow that use case. There are no conflicts. `name` is for single named values. `group` is for named collections of values.

dig.go
```diff
@@ -21,31 +21,40 @@
...
+	// Only one of name or group will be set.
+	name  string
+	group string
 }
 
 // Option configures a Container. It's included for future functionality;
 // currently, there are no concrete implementations.
 type Option interface {
-	unimplemented()
+	applyOption(*Container)
```
akshayjshah: Just `apply`? Only curious - obviously fine as-is too.

dig.go
```diff
@@ -246,28 +277,34 @@ func (cv connectionVisitor) Visit(res result) resultVisitor {
...
 
-	if conflict, ok := cv.keyPaths[k]; ok {
-		*cv.err = fmt.Errorf(
-			"cannot provide %v from %v in constructor %v: already provided by %v",
-			k, path, cv.n.ctype, conflict)
-		return nil
-	}
+		if _, ok := cv.c.providers[k]; ok {
+			*cv.err = fmt.Errorf(
+				"cannot provide %v from %v in constructor %v: already in the container",
```
akshayjshah: Hmmm...the `%v` of the conflict wasn't super-great for debugging, but it's a bummer to not have *any* identifying information about the other provider of the type.

doc.go
```diff
@@ -267,4 +267,50 @@
...
+//     dig.Out
+//
+//     Handler Handler `group:"server"`
+//   }
+//
+//   func NewHelloHandler() HandlerResult {
+//     ..
+//   }
+//
+//   func NewEchoHanlder() HandlerResult {
```
akshayjshah: Spelling of "Handler"

doc.go
```diff
@@ -267,4 +267,50 @@
...
+// unordered collection in the container. Other constructors can request all
+// values in this collection as a slice.
+//
+// Constructors can send values into value groups by returning a dig.Out
+// struct tagged with `group:".."`.
+//
+//   type HandlerResult struct {
+//     dig.Out
+//
+//     Handler Handler `group:"server"`
```
akshayjshah: `Handler http.Handler`?

param.go
```diff
@@ -326,3 +296,126 @@ func (po paramObject) Build(c *Container) (reflect.Value, error) {
...
+
+	name := f.Tag.Get(_nameTag)
+	optional, _ := isFieldOptional(f)
+	switch {
+	case f.Type.Kind() != reflect.Slice:
+		return pg, fmt.Errorf("value groups may be consumed as slices only: "+
+			"field %q (%v) is not a slice", f.Name, f.Type)
+	case name != "":
+		return pg, fmt.Errorf(
+			"cannot use named values with value groups: name:%q provided with group:%q", name, pg.Group)
```
akshayjshah: Can we avoid using the word "provided" here, since that has special connotations in dig and fx? "Requested" might be a workable alternative.

param.go
```diff
@@ -326,3 +296,126 @@ func (po paramObject) Build(c *Container) (reflect.Value, error) {
...
+	return pg, nil
+}
+
+func (pt paramGroupedSlice) Build(c *Container) (reflect.Value, error) {
+	k := key{group: pt.Group, t: pt.Type.Elem()}
+
+	for _, n := range c.providers[k] {
+		if err := n.Call(c); err != nil {
+			// TODO: better message
+			return _noValue, errWrapf(err, "failed to build %v", k)
```
akshayjshah: Can we include the group name in the error message? (Or is it included further up the error stack?)

result.go
```diff
@@ -283,3 +251,96 @@ func (ro resultObject) Extract(rr resultReceiver, v reflect.Value) {
...
+// newResultGrouped(f) builds a new resultGrouped from the provided field.
+func newResultGrouped(f reflect.StructField) (resultGrouped, error) {
+	rg := resultGrouped{Group: f.Tag.Get(_groupTag), Type: f.Type}
+
+	name := f.Tag.Get(_nameTag)
+	optional, _ := isFieldOptional(f)
+	switch {
+	case name != "":
+		return rg, fmt.Errorf(
+			"cannot use named values with value groups: name:%q provided with group:%q", name, rg.Group)
```
akshayjshah: In this case, I think "provided" makes sense - result objects come from providers.

stringer.go
```diff
@@ -51,9 +56,16 @@ func (n *node) String() string {
 }
 
 func (k key) String() string {
+	// ~tally.Scope means optional
+	// ~tally.Scope:foo means named optional
+	// [io.Reader]:foo refers to a group of io.Readers called 'foo'
```
akshayjshah: I suppose this is okay.

Would this be clearer with different notation?

```
tally.Scope
tally.Scope[optional]
tally.Scope[optional, name=foo]
tally.Scope[group=foo]
```

If we add a new tag, for example, I'm not sure how we'd extend the current system.

dig.go
abhinav: Unexported methods on interfaces from the same package are in the same
namespace.

If you have,

    type Option interface {
        apply(*Container)
    }
    
    type InvokeOption interface {
        apply(*invokeOptions)
    }
    
You can't have a single object implement both interfaces.

This is obviously not needed at this time but may be required in the future.

doc.go
abhinav: No, that was intentionally not `http.Handler` because given `[]http.Handler`, you can't really register them on a `ServeMux` -- there's no pattern/name attached.

param.go
abhinav: The message right now will be,

    failed to build [$type]:$groupName: ...

stringer.go
abhinav: I like this. Will change in a follow up.

abhinav: #168

abhinav:
I'll merge this later today (or Monday at the latest) if nobody has objections.

I'm going to spend some time test driving this without release next week.

param.go
amckinney: Since we aren't allowed to use optional, what's the behavior for a request value group with no provided types? I'm assuming that this would be an error and that this is the point of failure. We should consider adding a note to the package-level documentation about this.

abhinav: If you request `group:"foo"` and nothing provided values to it, you'll get an empty slice.

(Is that what you're asking?)

PR #170, opened 2017-11-06
Use cleaner string representations

### LINKED ISSUE #168: Cleaner string representations
As per https://github.com/uber-go/dig/pull/164#discussion_r147284505, string
representation for keys should be,

```
tally.Scope
tally.Scope[optional]
tally.Scope[optional, name=foo]
tally.Scope[group=foo]
```

abhinav:
This changes string representations for keys in the graph to be more
self-explanatory:

    Foo
    Bar[optional]
    Baz[name="x"]
    Qux[optional, name="y"]
    Quux[group="z"]

Resolves #168


dig_test.go
```diff
@@ -1357,7 +1357,7 @@ func TestProvideCycleFails(t *testing.T) {
 		err := c.Provide(newD)
 		require.Error(t, err)
 		assert.Contains(t, err.Error(),
-			"introduces a cycle: *dig.D -> [int]:bar -> [string]:foo -> *dig.D")
+			`introduces a cycle: *dig.D -> int[group="bar"] -> string[group="foo"] -> *dig.D`)
```
glibsm: this message is not correct as it strips the fact that it was a slice

it should be `[]int[group="bar"]` or `int[][group="bar"]`

abhinav: I agree that it's a bit weird but I think `int[group="bar"]` makes more sense
than `[]int[group="bar"]`  because each entry here is referring a single node
from the group. The way I'm visualizing it is,

```
                       +-----------+                          
                       |  MyThing  |<------------------------+
                       +-----------+                         |
     Ints []int `group:"foo"`|                               |
                             |                               |
         +-------------------+-------------------+           |
         |                   |                   |           |
         |                   |                   |           |
         v                   v                   v           |
+-----------------+ +-----------------+ +-----------------+  |
|int[group="foo"] | |int[group="foo"] | |int[group="foo"] |--+
+-----------------+ +-----------------+ +-----------------+   
         |                   |                                
         |                   |                                
         |                   |                                
         v                   v                                
+-----------------+ +-----------------+                       
|Type1[name="bar"]| |      Type2      |                       
+-----------------+ +-----------------+                       
```

So, in that sense each entry in a group is its own node.

glibsm: That's true, I was thinking about relying on a group, rather than providing one component into the group. Nevermind, disregard the comment.

Also, sweet ASCII art.

PR #172, opened 2017-11-30
Improved error messages

abhinav:
This PR attempts to make the error messages returned by dig more
helpful.

One common complaint people have with the error messages is that
the function signature is useless. The following message provides no
avenue for debugging.

    already provided by func() *Foo

One of the primary changes made in this PR is that function locations
are reported where appropriate. So the above becomes,

    already provided by "path/to/package".NewFoo (path/to/file.go:42)

I recommend reviewing each commit individually as these changes
collectively modify most of the files in dig in different places.

### DIFF
--- a/cycle.go
+++ b/cycle.go
@@ -1,0 +1,107 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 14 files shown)


breerly:
Can you provide some output when there is a longer chain of errors? I

abhinav:
This change doesn't do anything for error formatting right now (except the cycle detection errors, in which case there's an example in the comments). A long chain of errors will still be that massive mess that it is today but with more useful information.

breerly:
(thanks for updating tests)

abhinav:
Will land and release this on Monday.

zmt:
🥇

PR #175, opened 2017-12-20
Provide: Add Name option

abhinav:
This adds a `Name(string)` option to `Provide()` so that the following,

    c.Provide(NewThing, Name("foo"))

Is the same as the following,

    type result struct {
        dig.Out

        Thing *Thing `name:"foo"`
    }

    c.Provide(func(...) result {
        return result{Thing: NewThing(..)}
    })

The name applies to *all* values produced by the constructor--even if it
produces multiple values. Passing this option is not allowed if the
constructor produces a Result object.

This change was motivated by an offline discussion with @bayesianmind.
We agreed that Fx needs to support some way to specify the name for a
value produced by a constructor without rewriting the constructor to
produce an fx.Out. An API similar to the following was suggested.

    fx.Provide(fx.Named("foo", NewThing))

Regardless of the user-facing API, such functionality would need to rely
on reflection pretty heavily. It would need to generate a function using
`reflect.MakeFunc` equivalent to the following,

    func(...) (out struct {
        fx.Out

        Result *Thing `name:"foo"`
    }) {
        out.Result = NewThing(...)
        return
    }

See yarpc/yarpc-go#1200 for an example of the initial attempt made by
@bayesianmind.

I don't think that this functionality should rely this heavily on some
fairly fragile reflection logic. The way to build anonymous structs with
embedded fields is not the same between Go 1.8 and 1.9 (see
https://github.com/uber-go/dig/blob/a84ae412585ef6c4b8fc2f7577ff80f701317314/dig_test.go#L628-L635).

Supporting this in Dig natively is much more straightforward and safer,
and it requires no reflection.

This change defers the decision of how we should treat names on Result Objects
by throwing an error in that case.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -61,6 +61,17 @@ type provideOptions struct {
 	Name string
 }
 
+func (o *provideOptions) Validate() error {
+	// Names must be representable inside a backquoted string. The only
+	// limitation for raw string literals as per
+	// https://golang.org/ref/spec#raw_string_lit is that they cannot contain
+	// backquotes.
+	if strings.ContainsRune(o.Name, '`') {
+		return fmt.Errorf("invalid dig.Name(%q): names cannot contain backquotes", o.Name)
+	}
+	return nil
+}
+
 // A ProvideOption modifies the default behavior of Provide.
 type ProvideOption interface {
 	applyProvideOption(*provideOptions)
@@ -174,6 +185,9 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 	var options provideOptions
 	for _, o := range opts {
 		o.applyProvideOption(&options)
+	}
+	if err := options.Validate(); err != nil {
+		return err
 	}
 
 	if err := c.provide(constructor, options); err != nil {
--- a/dig.go
+++ b/dig.go
@@ -57,10 +57,47 @@ type optionFunc func(*Container)
 
 func (f optionFunc) applyOption(c *Container) { f(c) }
 
-// A ProvideOption modifies the default behavior of Provide. It's included for
-// future functionality; currently, there are no concrete implementations.
+type provideOptions struct {
+	Name string
+}
+
+// A ProvideOption modifies the default behavior of Provide.
 type ProvideOption interface {
-	unimplemented()
+	applyProvideOption(*provideOptions)
+}
+
+type provideOptionFunc func(*provideOptions)
+
+func (f provideOptionFunc) applyProvideOption(opts *provideOptions) { f(opts) }
+
+// Name is a ProvideOption that specifies that all values produced by a
+// constructor should have the given name. See also the package documentation
+// about Named Values.
+//
+// The name also applies to the fields of a dig.Out struct returned by the
+// constructor if those fields did not have their own name:"..." tag.
+//
+// Given,
+//
+//   type ConnectionResult struct {
+//     dig.Out
+//
+//     Conn    *sql.DB
+//     ROConn  *sql.DB `name:"ro"`
+//   }
+//
+//   func NewConnection(...) (ConnectionResult, error)
+//
+// The following will provide two connections to the container: one under the
+// name "ro" and the other under the name "rw".
+//
+//   c.Provide(NewConnection, dig.Name("rw"))
+//
+// This option has no effect on Value Groups.
+func Name(name string) ProvideOption {
+	return provideOptionFunc(func(opts *provideOptions) {
+		opts.Name = name
+	})
 }
 
 // An InvokeOption modifies the default behavior of Invoke. It's included for
@@ -133,7 +170,13 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 	if ctype.Kind() != reflect.Func {
 		return fmt.Errorf("must provide constructor function, got %v (type %v)", constructor, ctype)
 	}
-	if err := c.provide(constructor); err != nil {
+
+	var options provideOptions
+	for _, o := range opts {
+		o.applyProvideOption(&options)
+	}
+
+	if err := c.provide(constructor, options); err != nil {
 		return errProvide{
 			Func:   digreflect.InspectFunc(constructor),
 			Reason: err,
@@ -188,8 +231,8 @@ func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
 	return nil
 }
 
-func (c *Container) provide(ctor interface{}) error {
-	n, err := newNode(ctor)
+func (c *Container) provide(ctor interface{}, opts provideOptions) error {
+	n, err := newNode(ctor, nodeOptions{ResultName: opts.Name})
 	if err != nil {
 		return err
 	}
@@ -347,7 +390,12 @@ type node struct {
 	Results resultList
 }
 
-func newNode(ctor interface{}) (*node, error) {
+type nodeOptions struct {
+	// If specified, all values produced by this node have the provided name.
+	ResultName string
+}
+
+func newNode(ctor interface{}, opts nodeOptions) (*node, error) {
 	ctype := reflect.TypeOf(ctor)
 
 	params, err := newParamList(ctype)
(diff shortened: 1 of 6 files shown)


dig.go
```diff
@@ -57,10 +57,48 @@ type optionFunc func(*Container)
 
 func (f optionFunc) applyOption(c *Container) { f(c) }
 
+type provideOptions struct {
+	Name string
+}
+
 // A ProvideOption modifies the default behavior of Provide. It's included for
 // future functionality; currently, there are no concrete implementations.
```
bayesianmind: nit: this comment needs to be updated now

dig.go
```diff
@@ -347,15 +391,22 @@ type node struct {
 	Results resultList
 }
 
-func newNode(ctor interface{}) (*node, error) {
+type nodeOptions struct {
+	// If specified, all values produced by this node have the provided name.
+	//
+	// This cannot be mixed wiht result objects.
```
willhug: `with`

willhug: (also outdated doc)

doc.go
```diff
@@ -221,10 +221,25 @@
...
+//   func NewReadOnlyConnection(...) (*sql.DB, error)
+//   func NewReadWriteConnection(...) (*sql.DB, error)
+//
+// You can provide *sql.DB into a Container under different names by passing
+// the dig.Name option.
+//
+//   c.Provide(NewReadOnlyConnection, dig.Name("ro"))
+//   c.Provide(NewReadWriteConnection, dig.Name("rw"))
+//
+// Alternatively, you can producce a dig.Out struct and tag its fields with
```
willhug: `produce`

result.go
```diff
@@ -296,18 +303,16 @@ func newResultObjectField(idx int, f reflect.StructField) (resultObjectField, er
...
+			// can modify in-place because options are passed-by-value.
+			opts.Name = name
+		}
+		r, err = newResult(f.Type, opts)
 		if err != nil {
 			return rof, err
 		}
 	}
 
-	if rs, ok := r.(resultSingle); ok {
```
willhug: groups don't have names right?

dig.go
```diff
@@ -57,10 +57,47 @@ type optionFunc func(*Container)
...
+//
+//   func NewConnection(...) (ConnectionResult, error)
+//
+// The following will provide two connections to the container: one under the
+// name "ro" and the other under the name "rw".
+//
+//   c.Provide(NewConnection, dig.Name("rw"))
+//
+// This option has no effect on Value Groups.
+func Name(name string) ProvideOption {
```
willhug: Not sure if we want to reuse the "Name" in other Options, would need to make an interface/struct to make that work.

prashantv:
Capturing offline discussion with @abhinav 

We should probably validate names -- previously we relied on the compiler to validate names.

dig.go
abhinav: Discussed offline but: Even if we could support a similar option for
parameters, it would have to be a new option rather than `Name` being re-used
because they would both need to be provided at `c.Provide(..)` time. That
option could be a superset of ProvideOption and InvokeOption if needed.

I don't foresee us wanting to re-use this option but if we find a really
compelling need in the future, two things worth noting:

- Cutting a 2.0 of Dig is cheap as long as the Fx API remains the same.
- Depending on where you draw the line on what constitutes as breaking at this
  level, changing `Name` (a top-level function) to return a type that is valid
  for both, its current return type and something else *should* be okay.

result.go
abhinav: groups and names are mutually exclusive.

glibsm:
> We should probably validate names

What do you mean by "validating names"? It's not entirely clear to me

dig.go
```diff
@@ -188,8 +231,8 @@ func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
 	return nil
 }
 
-func (c *Container) provide(ctor interface{}) error {
-	n, err := newNode(ctor)
+func (c *Container) provide(ctor interface{}, opts provideOptions) error {
```
glibsm: Are you sure that `opts ...ProvideOption` is not flattened too soon if it's immediately converted to a new struct and propagated down through `opts provideOptions`? Seems like you had to pipe it through quite a few structs in the new form and change tests.

At the very least you wouldn't have to change the test signatures if it was kept as varargs for a while.

prashantv:
> What do you mean by "validating names"? It's not entirely clear to me

Currently, names can only be specified in struct tags, which have limitations enforced by the compiler (E.g., no new lines).

If we allow a `Name` option that takes an arbitrary string, it might contain a name that cannot be consumed because the name contains invalid characters.

abhinav:
RE: Validation: Turns out we need very minimal validation. As per https://golang.org/pkg/reflect/#StructTag, tag values can be anything that you can represent with a double quoted string (with escapes) [including newlines](https://play.golang.org/p/pNaWieSFtSe).

Tag values can't contain a `\`` because the Go backtick string literals don't have a way of representing backticks inside them.

dig_test.go
```diff
@@ -543,6 +598,49 @@ func TestEndToEndSuccess(t *testing.T) {
 		}), "invoke should succeed, pulling out two named instances")
 	})
 
+	t.Run("name tags on result structs recurse", func(t *testing.T) {
```
akshayjshah: Errr...is "recurse" a verb? I always thought it was "recur" 🤷‍♂️

abhinav: https://en.wiktionary.org/wiki/recurse#Verb

:man_shrugging:

dig.go
abhinav: I wanted to avoid the user-facing `ProvideOption` being passed to the underlying layers. This is a good place IMO to convert to an "internal" representation of the options which is easy to consume from the other parts of the library.

The other structs are mostly in anticipation of future options that don't necessarily map to result options. Plumbing resultOptions in from provide, or plumbing provideOptions down to result, both smell to me.

abhinav:
@uber-go/fx, this is ready to merge if there are no more concerns. I'll hold
off on merging this until I've had a chance to finish the Fx-side of this
work.

dig.go
```diff
@@ -57,10 +57,58 @@ type optionFunc func(*Container)
...
+type provideOptionFunc func(*provideOptions)
+
+func (f provideOptionFunc) applyProvideOption(opts *provideOptions) { f(opts) }
+
+// Name is a ProvideOption that specifies that all values produced by a
+// constructor should have the given name. See also the package documentation
+// about Named Values.
+//
+// The name also applies to the fields of a dig.Out struct returned by the
+// constructor if those fields did not have their own name:"..." tag.
```
prashantv: I thought we were going to make this an error? I'd prefer to start with *only* supporting fields that don't have an existing `name` tag.

abhinav: Discussed offline: We will disallow `dig.Out`s for now.

Applying to fields of a `dig.Out` brings up questions we don't have good
answers to yet. For example, if we don't stomp on existing names, it's
impossible for someone to provide two copies of `somemodule.Result` with
different names.

akshayjshah: Fair enough. Changing this decision isn't backward-compatible, so we may want to call this option `ShallowName` or something similar.

abhinav: Going from error-to-no-error is backwards compatible here. If the following
stops exploding between minor versions, that's completely okay.

```go
c.Provide(NewThing, dig.Name("foo"))
```

The other aspect of the future change is how `name:".."` is interpreted on
embedded result fields. The complexity of behavior there is what lead to
deferring the decision for now by disallowing `dig.Name` on dig.Out structs.

Given,

```go
type Result struct {
  dig.Out
  
  A   string        `name:"a"`
  Foo AnotherResult `name:"foo"`
}

type AnotherResult struct {
    dig.Out
    
    B string `name:"b"`
}
```

The `name:"foo"` on `Result.Foo` has no effect currently because `name:".."`
tags on embedded result structs are ignored.

If we implement some sort of merging behavior here, that could technically
count as breaking change because previously, `Provide(New)` would result in
two values being provided: `string[name="a"]` and `string[name="b"]`. With
merging, the latter may become something like `string[name="foo/b"]`.

However, I think that this is not *actually* breaking because `name:".."` on
embedded result structs is currently unsupported/invalid usage.

abhinav:
The requested change was made; it can be reviewed in the newest commit.

dig.go
akshayjshah: You're absolutely right. I didn't read carefully enough, and thought that we were *ignoring* result structs - as long as we're erroring, this sounds great.

PR #176, opened 2017-12-27
Reduce access of internals between abstractions

abhinav:
(Each commit should be reviewed individually.)

This pull request changes some of the internals to stop relying on internal
details of each other.

Previously, everything was accessing and modifying the internals of
`*Container` and `*node`. This pull request puts a bunch of interfaces and
unexported methods in place instead.

These changes are needed so that we can move some of the code from the
top-level package into internal packages, which will allow us to re-use that
functionality from a Dig introspection package.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -126,6 +126,18 @@ type Container struct {
 	rand *rand.Rand
 }
 
+// containerWriter provides write access to the Container.
+type containerWriter interface {
+	// setValue sets the value with the given name and type in the container.
+	// If a value with the same name and type already exists, it will be
+	// overwritten.
+	setValue(name string, t reflect.Type, v reflect.Value)
+
+	// submitGroupedValue submits a value to the value group with the provided
+	// name.
+	submitGroupedValue(name string, t reflect.Type, v reflect.Value)
+}
+
 // New constructs a Container.
 func New(opts ...Option) *Container {
 	c := &Container{
@@ -148,6 +160,15 @@ func setRand(r *rand.Rand) Option {
 	return optionFunc(func(c *Container) {
 		c.rand = r
 	})
+}
+
+func (c *Container) setValue(name string, t reflect.Type, v reflect.Value) {
+	c.values[key{name: name, t: t}] = v
+}
+
+func (c *Container) submitGroupedValue(name string, t reflect.Type, v reflect.Value) {
+	k := key{group: name, t: t}
+	c.groups[k] = append(c.groups[k], v)
 }
 
 // Provide teaches the container how to build values of one or more types and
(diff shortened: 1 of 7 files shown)


result.go
```diff
@@ -196,19 +207,19 @@ func (resultList) Extract(resultReceiver, reflect.Value) {
...
-		r.Extract(rr, values[i])
-	}
-}
-
-// resultError is an error returned by a constructor.
-type resultError struct{}
+func (rl resultList) ExtractList(rr resultReceiver, values []reflect.Value) error {
+	for i, v := range values {
+		if resultIdx := rl.resultIndexes[i]; resultIdx >= 0 {
+			rl.Results[resultIdx].Extract(rr, v)
```
amckinney: clever use of indexing here

dig.go
```diff
@@ -162,15 +198,57 @@ func setRand(r *rand.Rand) Option {
...
+	typeSet := make(map[reflect.Type]struct{}, len(c.providers))
+	for k := range c.providers {
+		typeSet[k.t] = struct{}{}
+	}
+
+	types := make([]reflect.Type, 0, len(typeSet))
+	for t := range typeSet {
+		types = append(types, t)
+	}
+	sort.Sort(byTypeName(types))
```
amckinney: Is this to make the output for suggestions deterministic?

error.go
```diff
@@ -204,8 +205,8 @@ func newErrMissingType(c *Container, k key) errMissingType {
 
 	err := errMissingType{Key: k}
 	for _, t := range suggestions {
-		k.t = t
-		if len(c.providers[k]) > 0 {
+		if len(c.getValueProviders(k.name, t)) > 0 {
+			k.t = t
```
amckinney: Mutating `k.t` before appending to suggestions isn't ideal. It might be worth adding methods to `errMissingType` with a similar set signature as `getValueProviders` above so that this call is a little more explicit. This obviously doesn't need to happen here, but I think it's worth noting for a potential future change.

param.go
```diff
@@ -204,22 +203,20 @@ type paramSingle struct {
 	Type     reflect.Type
 }
 
-func (ps paramSingle) Build(c *Container) (reflect.Value, error) {
-	k := key{name: ps.Name, t: ps.Type}
-	if v, ok := c.values[k]; ok {
+func (ps paramSingle) Build(c containerStore) (reflect.Value, error) {
```
amckinney: Great cleanup in this method. Much easier to read now.

param.go
```diff
@@ -231,10 +228,13 @@ func (ps paramSingle) Build(c *Container) (reflect.Value, error) {
 			return reflect.Zero(ps.Type), nil
 		}
 
-		return _noValue, errParamSingleFailed{Key: k, Reason: err}
+		return _noValue, errParamSingleFailed{Key: key{t: ps.Type, name: ps.Name}, Reason: err}
```
amckinney: nit: consider using the same static construction in the error case above for consistency:
```
key{name: ps.Name, t: ps.Type}
```

dig.go
abhinav: Yep. Deterministic output is easier to test.

error.go
abhinav: Fair. This will change a bit in a follow-up PR and I'll see if the
`k.t` mutation can be simplified.

PR #178, opened 2018-01-02
travis: Fix lint

abhinav:
Turns out `make lint` failures were not actually breaking the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,5 +15,5 @@ cache:
 install:
   - make dependencies
 script:
-  - test "$LINT" -eq 1 && make lint || echo "Skipping lint"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -15,5 +15,5 @@ cache:
 install:
   - make dependencies
 script:
-  - test "$LINT" -eq 1 && make lint || echo "Skipping lint"
+  - test "$LINT" = 1 && make lint
```
akshayjshah: I guess we'll find out, but won't this fail the build whenever linting is disabled?

abhinav: Derp. You're right.

akshayjshah: I think we want something more like

```
test -z "$LINT" || make lint
```

PR #180, opened 2018-01-04
CHANGELOG: Standardize format

abhinav:
As discussed over email, this standardizes our changelog to the format defined
by http://keepachangelog.com/en/1.0.0/.

Preview at https://github.com/uber-go/dig/blob/standardize-changelog/CHANGELOG.md

For this repo, I went back and added the Changed/Added/etc. sections. We don't
need to do that for other repositories.


abhinav:
CC @uber-go/fx

PR #188, opened 2018-05-23
Introspection: Introduce DOT graph in Container

leyao:
Added dotgraph field in Container that represents the type dependency graph using graph edges.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -113,8 +113,10 @@ type InvokeOption interface {
 
 // A graphNode represents a node in the dependency DOT graph
 type graphNode struct {
-	Type string
-	// TODO(leyao): Add Groups, Names, and Optionals
+	Type     string
+	name     string
+	optional bool
+	group    string
 }
 
 // A graphEdge represents an edge in the dependency DOT graph
@@ -399,17 +401,6 @@ func (c *Container) addToGraph(ge []graphEdge) {
 	c.dotgraph = append(c.dotgraph, ge...)
 }
 
-// toGraphEdges returns graph edges in node n
-func (n *node) toGraphEdges() []graphEdge {
-	edges := []graphEdge{}
-	for _, pnode := range n.paramList.GraphNode() {
-		for _, rnode := range n.resultList.GraphNode() {
-			edges = append(edges, graphEdge{param: pnode, result: rnode})
-		}
-	}
-	return edges
-}
-
 // Builds a collection of all result types produced by this node.
 func (c *Container) findAndValidateResults(n *node) (map[key]struct{}, error) {
 	var err error
@@ -598,6 +589,17 @@ func (n *node) Call(c containerStore) error {
 	receiver.Commit(c)
 	n.called = true
 	return nil
+}
+
+// toGraphEdges returns graph edges in node n
+func (n *node) toGraphEdges() []graphEdge {
+	edges := []graphEdge{}
+	for _, pnode := range n.paramList.GraphNode() {
+		for _, rnode := range n.resultList.GraphNode() {
+			edges = append(edges, graphEdge{param: pnode, result: rnode})
+		}
+	}
+	return edges
 }
 
 // Checks if a field of an In struct is optional.
--- a/dig.go
+++ b/dig.go
@@ -111,7 +111,7 @@ type InvokeOption interface {
 	unimplemented()
 }
 
-// A dotNode represents a node in a Container's DOT-format graph.
+// A dotNode represents a single node in a dotGraph.
 type dotNode struct {
 	Type     string
 	name     string
@@ -119,15 +119,30 @@ type dotNode struct {
 	group    string
 }
 
-// A dotEdge represents an edge in a Container's DOT-format graph.
+// A dotEdge connects a dotNode parameter to a dotNode result.
 type dotEdge struct {
-	param  dotNode
-	result dotNode
+	param  *dotNode
+	result *dotNode
 }
 
 // A dotGraph represents the DOT-format graph in a Container.
 type dotGraph struct {
-	edges []dotEdge
+	edges []*dotEdge
+}
+
+// add adds the dotEdges in node n into dg.
+func (dg *dotGraph) add(n *node) {
+	pnodes := n.paramList.DotNodes()
+	rnodes := n.resultList.DotNodes()
+	edges := make([]*dotEdge, len(pnodes)*len(rnodes))
+
+	for i, pnode := range pnodes {
+		for j, rnode := range rnodes {
+			edges[i*len(rnodes)+j] = &dotEdge{param: pnode, result: rnode}
+		}
+	}
+
+	dg.edges = append(dg.edges, edges...)
 }
 
 // Container is a directed acyclic graph of types and their dependencies.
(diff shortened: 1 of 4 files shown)


dig.go
```diff
@@ -111,6 +111,20 @@ type InvokeOption interface {
 	unimplemented()
 }
 
+// A graphNode represents a node in the dependency DOT graph
+type graphNode struct {
```
amckinney: Seeing as "graph" is a fairly overloaded term here, let's consider naming this `dotNode` (since this is represented in DOT-format).

dig.go
```diff
@@ -111,6 +111,20 @@ type InvokeOption interface {
...
+	Type     string
+	name     string
+	optional bool
+	group    string
+}
+
+// A graphEdge represents an edge in the dependency DOT graph
+type graphEdge struct {
```
amckinney: Similar to the above comment, `dotEdge` might be better here.

dig.go
```diff
@@ -125,6 +139,9 @@ type Container struct {
 
 	// Source of randomness.
 	rand *rand.Rand
+
+	// Dot graph of dependencies
+	dotgraph []graphEdge
```
amckinney: nit: `dotgraph` -> `dotGraph`. We might also want to consider making:
```
type dotGraph struct { ... }
```
for the `[]graphEdge` representation.

This could become (or something similar):

```
type Container struct {
  ...
  dg *dotGraph
}
```

dig.go
```diff
@@ -372,10 +390,17 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
 			return err
 		}
 	}
+	ge := n.toGraphEdges()
+	c.addToGraph(ge)
```
amckinney: Since the `ge` variable is only used in the call here, we could instead have a `Container.addDotNode` method that accepts the `node` struct.

```
func (c *Container) addDotNode(n *node) {
  c.dotgraph = append(c.dotgraph, n.toGraphEdges()...)
}
```

dig.go
```diff
@@ -566,6 +591,17 @@ func (n *node) Call(c containerStore) error {
 	return nil
 }
 
+// toGraphEdges returns graph edges in node n
+func (n *node) toGraphEdges() []graphEdge {
+	edges := []graphEdge{}
```
amckinney: nit: We can actually pre-determine the size of the slice here:
```
edges := make([]graphEdge, len(n.paramList.GraphNode()) * len(n.resultList.GraphNode()))
...
```

param.go
```diff
@@ -45,6 +45,9 @@ type param interface {
 	//
 	// This MAY panic if the param does not produce a single value.
 	Build(containerStore) (reflect.Value, error)
+
+	// GraphNode returns a slice of param(s) represented in graphNodes for the dotgraph
+	GraphNode() []graphNode
```
amckinney: Let's consider making this `DotNodes()` (plural), so that it better describes that a slice is returned.

param.go
```diff
@@ -144,6 +147,14 @@ type paramList struct {
 	Params []param
 }
 
+func (pl paramList) GraphNode() []graphNode {
+	types := []graphNode{}
```
amckinney: nit: we can instantiate this as `var types []graphNode` to avoid an unnecessary allocation if `len(pl.Params) == 0`.

param.go
```diff
@@ -245,6 +264,14 @@ type paramObject struct {
 	Fields []paramObjectField
 }
 
+func (po paramObject) GraphNode() []graphNode {
+	types := []graphNode{}
```
amckinney: similar to above comment on allocation.

result.go
```diff
@@ -41,6 +41,9 @@ type result interface {
 	//
 	// This MAY panic if the result does not consume a single value.
 	Extract(containerWriter, reflect.Value)
+
+	// GraphNode returns a slice of result(s) represented in graphNodes for the dotgraph
+	GraphNode() []graphNode
```
amckinney: Similar to above - let's consider `DotNodes` here.

result.go
```diff
@@ -163,6 +166,14 @@ type resultList struct {
 	resultIndexes []int
 }
 
+func (rl resultList) GraphNode() []graphNode {
+	types := []graphNode{}
```
amckinney: alloc

result.go
```diff
@@ -235,6 +253,14 @@ type resultObject struct {
 	Fields []resultObjectField
 }
 
+func (ro resultObject) GraphNode() []graphNode {
+	types := []graphNode{}
```
amckinney: alloc

amckinney:
This looks really great! There are some comments with respect to naming and some small nits with respect to allocations, but otherwise awesome!

dig.go
```diff
@@ -111,6 +111,20 @@ type InvokeOption interface {
 	unimplemented()
 }
 
+// A graphNode represents a node in the dependency DOT graph
```
amckinney: nit: Comment should be complete sentences. (Also relevant in other places)
```
A graphNode represents a node in a Container's DOT-format graph.
```

akshayjshah:
@SyrieBianco You may want to take a look at this one too. Leah's working on making our dependency injection system capable of producing pretty visualizations of the dependency graph, which should make it easier to debug missing/otherwise broken dependencies.

The format she's using is DOT, which the [GraphViz](https://graphviz.gitlab.io/gallery/) tool understands.

dig.go
```diff
@@ -111,6 +111,25 @@ type InvokeOption interface {
 	unimplemented()
 }
 
+// A dotNode represents a node in a Container's DOT-format graph.
```
amckinney: nit: Given that `dotGraph` includes the context with respect to the `DOT-format` graph, we could make this more focused towards how it relates to the other data structures and remove the `DOT-format` note.

Something like:
```
// A dotNode represents a single node in a dotGraph.
```

dig.go
```diff
@@ -111,6 +111,25 @@ type InvokeOption interface {
...
+type dotNode struct {
+	Type     string
+	name     string
+	optional bool
+	group    string
+}
+
+// A dotEdge represents an edge in a Container's DOT-format graph.
```
amckinney: Similar to above:
```
// A dotEdge connects a dotNode parameter to a dotNode result.
```

dig.go
```diff
@@ -372,10 +395,28 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
 			return err
 		}
 	}
+	c.addNode(n)
 
 	return nil
 }
 
+// addNode adds the dotEdges in node n into the DOT-format graph in Container c.
+func (c *Container) addNode(n *node) {
```
amckinney: nit: Seeing as node is frequently referenced, we should be a _little_ more explicit with what this method is doing. We could make a method on the `dotGraph` type so that it's more clearly insulated.

```
func (dg *dotGraph) add(n *node) { ... }
```

The above call would then become `c.dg.add(n)`

dig.go
```diff
@@ -372,10 +395,28 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
...
+	pnodes := n.paramList.DotNodes()
+	rnodes := n.resultList.DotNodes()
+	edges := make([]dotEdge, len(pnodes)*len(rnodes))
+	i := 0
```
amckinney: Rather than allocating an `i` here, we can use the index from the `range` in our loop:
```
for  i, pnode := range pnodes { ... } // i == 0 -> i == 1, ...
```

amckinney: On second thought, now that we are collecting the `edges` slice and simply expanding it into the `c.dg.edges` slice (via the `c.dg.edges = append(c.dg.edges, edges...)` call below), it's probably easier to append directly into the `c.dg.edges`.

This would look similar to:
```
for _, pnode := range pnodes {
  for _, rnode:= range rnodes {
    c.dg.edges = append(c.dg.edges, dotEdge{param: pnode, result: rnode})
}
```

dig.go
```diff
@@ -111,6 +111,25 @@ type InvokeOption interface {
...
+
+// A dotEdge represents an edge in a Container's DOT-format graph.
+type dotEdge struct {
+	param  dotNode
+	result dotNode
+}
+
+// A dotGraph represents the DOT-format graph in a Container.
+type dotGraph struct {
+	edges []dotEdge
```
amckinney: Seeing as we might need to mutate the `dotEdge` contents (i.e. in the presence of `Invoke` failures), let's use pointers here.

```
type dotGraph struct {
  edges []*dotEdge
}
```

This would also apply to the above:
```
type dotEdge struct {
  param *dotNode
  result *dotNode
}
```

PR #189, opened 2018-05-25
Move dot structs (node, edge, graph) to internal package

leyao:


internal/dot/graph.go
```diff
@@ -0,0 +1,53 @@
...
+	Edges []*Edge
+}
+
+// Add adds the edges in node n into dg.
+func (dg *Graph) Add(params []*Node, results []*Node) {
+	edges := make([]*Edge, len(params)*len(results))
+
+	for i, param := range params {
+		for j, result := range results {
+			edges[i*len(results)+j] = &Edge{Param: param, Result: result}
```
abhinav: You can do this without the index calculation math if you use the
`make([]foo, length, capacity)` form of `make`.

```go
edges := make([]*Edge, 0, len(params)*len(results))
```

This creates a slice that has room for that many items but it's still empty
(so `edges[0]` will panic).

Then it's just a matter of appending in the right order, which in this case is
iteration order:

```go
for _, param := range params {
    for _, result := range results {
        edges = append(edges, &Edge{Param: param, Result: result})
    }
}
```

That said, it looks like creating this slice is unnecessary because you can
just append to `dg.Edges` directly. Is there a reason not to do that?

dig.go
```diff
@@ -229,7 +196,7 @@ func New(opts ...Option) *Container {
 		values:    make(map[key]reflect.Value),
 		groups:    make(map[key][]reflect.Value),
 		rand:      rand.New(rand.NewSource(time.Now().UnixNano())),
-		dg:        new(dotGraph),
+		dg:        new(dot.Graph),
```
abhinav: It may be safer in the long run to build `dot.Graph` on-demand from a `Container` rather than keeping a running record of it. Otherwise any changes to internal book-keeping have to be propagated to the `dot.Graph` code.

internal/dot/graph.go
```diff
@@ -0,0 +1,53 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package dot
+
+// Node represents a single node in a graph.
+type Node struct {
```
abhinav: nit: For comments, `X represents foo` is redundant. You can just say `X is foo` or `X does bar`.  (Edge is fine: Edge connects [..]).

internal/dot/graph.go
```diff
@@ -0,0 +1,53 @@
...
+	Result *Node
+}
+
+// Graph represents the DOT-format graph in a Container.
+type Graph struct {
+	Edges []*Edge
+}
+
+// Add adds the edges in node n into dg.
+func (dg *Graph) Add(params []*Node, results []*Node) {
```
abhinav: Related to our offline discussion: This is basically `Add(constructor details)`
so it can be,

```go
type Constructor struct {
    Params  []*Node
    Results []*Node
}

func (dg *Graph) Add(*Constructor) {
    // ...
}
```

abhinav:
@glibsm The plan is to generate [DOT files][1] which will then be rendered as
images into a debug-page.

  [1]: https://en.wikipedia.org/wiki/DOT_(graph_description_language)

dig.go
leyao: Good point, we can address this in a future PR

internal/dot/graph.go
leyao: Yes, the structure of the graph will be changed in the next PR after this is merged

internal/dot/graph.go
leyao: I was attempting to avoid the unnecessary size re-allocation in the `dg.Edges` slice here so I used a separate `edge` slice and appended it on after. I think that the best approach would be to take your
```
edges := make([]*Edge, 0, len(params)*len(results))
```
suggestion and use `append` afterwards. 

What are your thoughts? Do you think the `make` allocation is unnecessary/overkill?

PR #190, opened 2018-05-30
dot/Graph: Include constructor details

leyao:
In order to better visualize the type dependencies, the types should be
clustered by constructors in the graph. The concept of `Edge`s from one
type to another is replaced by constructors (`Ctor`s) that contain
details about the constructor.


internal/dot/graph.go
```diff
@@ -28,26 +50,34 @@ type Node struct {
...
+func (dg *Graph) Add(c *Ctor, params []*Node, results []*Node) {
+	c.Params = getNodes(dg, params)
+	c.Results = getNodes(dg, results)
+	dg.Ctors = append(dg.Ctors, c)
 }
 
-// Graph is the DOT-format graph in a Container.
-type Graph struct {
-	Edges []*Edge
+func (n *Node) str() string {
```
amckinney: Seeing as this is already in an `internal/` package, we can export this and implement the [`fmt.Stringer`](https://golang.org/pkg/fmt/#Stringer) interface:
```
func (dg *Graph) String() string { ... }
```

In general I agree with your inclination to un-export methods that aren't used outside of this package (as in `getNodes` below), but conforming to the `fmt.Stringer` interface is more appropriate here.

internal/dot/graph.go
```diff
@@ -28,26 +50,34 @@ type Node struct {
...
+	for i, node := range nodes {
+		k := node.str()
 
-	for _, param := range params {
-		for _, result := range results {
-			edges = append(edges, &Edge{Param: param, Result: result})
+		if n := dg.Nodes[k]; n != nil {
+			ptrs[i] = n
+		} else {
+			dg.Nodes[k] = node
```
amckinney: A more-detailed note on how this function behaves might be helpful here. 
Seeing as this is storing nodes into the `dg.Nodes` map, we might want to update the function name so that it better encapsulates this mutation. `getNodes` might lead to unexpected results (i.e. "A 'read' ended up mutating the 'Graph' unexpectedly")

internal/dot/graph.go
```diff
@@ -28,26 +50,34 @@ type Node struct {
...
-// Edge connects a node parameter to a node result.
-type Edge struct {
-	Param  *Node
-	Result *Node
+// Add updates the param and result nodes in the constructor and adds the constructor to the graph
+func (dg *Graph) Add(c *Ctor, params []*Node, results []*Node) {
+	c.Params = getNodes(dg, params)
+	c.Results = getNodes(dg, results)
+	dg.Ctors = append(dg.Ctors, c)
 }
```
amckinney: It might be useful to add a note on why you aren't re-using the `dig.node.String` method here (i.e. a note on the optional field).

internal/dot/graph.go
```diff
@@ -20,6 +20,28 @@
...
+type Ctor struct {
+	Name    string
+	Package string
+	File    string
+	Line    int
+	Params  []*Node
+	Results []*Node
+}
+
+// Graph is the DOT-format graph in a Container represented by a list of Ctor-s.
```
amckinney: `Ctor-s` -> `Ctors`

internal/dot/graph.go
```diff
@@ -20,6 +20,28 @@
 
 package dot
 
+import (
+	"fmt"
+)
+
+// Ctor encodes the edges in a graph. It includes information like the name
+// of the constructor, the package, file, and line where the constructor is provided,
```
amckinney: nit: can drop the `and` before line

internal/dot/graph.go
```diff
@@ -20,6 +20,28 @@
 
 package dot
 
+import (
+	"fmt"
+)
+
+// Ctor encodes the edges in a graph. It includes information like the name
```
amckinney: nit: A simple description of the concept of an `edge` (in this context) might help here.

dig_test.go
```diff
@@ -2455,214 +2456,266 @@ func TestFailingFunctionDoesNotCreateInvalidState(t *testing.T) {
 	}), "second invoke must fail")
 }
 
-func TestDotGraph(t *testing.T) {
-	type T1 struct{}
-	type T2 struct{}
-	type T3 struct{}
-	type T4 struct{}
+func compareCtor(t *testing.T, expected *dot.Ctor, ctor *dot.Ctor) {
```
amckinney: Nice! This, in combination with `compareCtors`, is great.

dig.go
```diff
@@ -377,7 +377,14 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
...
 	}
-	c.dg.Add(n.paramList.DotNodes(), n.resultList.DotNodes())
+
+	constructor := &dot.Ctor{
+		Name:    n.location.Name,
+		Package: n.location.Package,
+		File:    n.location.File,
+		Line:    n.location.Line,
+	}
+	c.dg.Add(constructor, n.paramList.DotNodes(), n.resultList.DotNodes())
```
amckinney: It looks like introducing the `internal/dot` package made it difficult to to interface between this package and the `dot.Graph` type (since we aren't able to pass in the `node` type to the other package , we have to pre-process these types). Let's consider alternatives here.

dig.go
```diff
@@ -377,7 +377,16 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
...
+
+	constructor := &dot.Ctor{
+		Name:    n.location.Name,
+		Package: n.location.Package,
+		File:    n.location.File,
+		Line:    n.location.Line,
+		Params:  n.paramList.DotNodes(),
+		Results: n.resultList.DotNodes(),
+	}
+	c.dg.Ctors = append(c.dg.Ctors, constructor)
```
amckinney: An alternative would be to use the `dot.Graph.Add` method like before:

```
func (dg *Graph) Add(c *Ctor) {
  dg.Ctors = append(dg.Ctors, c)
}
```

```
c.dg.Add(&dot.Ctor{
  Name:    n.location.Name,
  Package: n.location.Package,
  File:    n.location.File,
  Line:    n.location.Line,
  Params:  n.paramList.DotNodes(),
  Results: n.resultList.DotNodes(),
})
```

But seeing as this is largely the same, I'll leave it up to you!

dig.go
```diff
@@ -377,7 +377,16 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
...
 	}
-	c.dg.Add(n.paramList.DotNodes(), n.resultList.DotNodes())
+
+	constructor := &dot.Ctor{
+		Name:    n.location.Name,
+		Package: n.location.Package,
+		File:    n.location.File,
+		Line:    n.location.Line,
+		Params:  n.paramList.DotNodes(),
+		Results: n.resultList.DotNodes(),
```
abhinav: minor readability thing: I would create a `newDotCtor(*node)` function in this package so that this code can be,

```
c.dg.Ctors = append(c.dg.Ctors, newDotCtor(n))
```

dig_test.go
abhinav: nit: rename to `assertCtorEquals` or similar? "compare" suggests that this will compare them without failing the test.

dig_test.go
```diff
@@ -2455,214 +2456,266 @@ func TestFailingFunctionDoesNotCreateInvalidState(t *testing.T) {
...
+	assert.Equal(t, path.Base(expected.File), path.Base(ctor.File))
+	assert.Equal(t, expected.Params, ctor.Params)
+	assert.Equal(t, expected.Results, ctor.Results)
+	assert.NotZero(t, ctor.Line)
+}
 
-	e1, e2, e3, e4 := &dot.Node{Type: "dig.T1"}, &dot.Node{Type: "dig.T2"}, &dot.Node{Type: "dig.T3"}, &dot.Node{Type: "dig.T4"}
+func compareCtors(t *testing.T, expected []*dot.Ctor, ctors []*dot.Ctor) {
+	for i := range ctors {
+		compareCtor(t, expected[i], ctors[i])
```
abhinav: nit: You can use the two-value form of `range` here.

```
for i, c := range ctors {
  compareCtor(t, expected[i], c)
}
```

internal/dot/graph.go
```diff
@@ -20,34 +20,27 @@
 
 package dot
 
+// Ctor encodes the edges that connects results to params in a graph. It includes information
+// like the name of the constructor, the package, file, line where the constructor is
+// provided, and the params and results of the constructor.
+type Ctor struct {
```
abhinav: nit: Docs don't need to repeat obvious code. Don't need to list fields in the
docs. Maybe something along the lines of this?

```
// Ctor encodes a constructor provided to the container for the DOT graph.
```

internal/dot/graph.go
```diff
@@ -20,34 +20,27 @@
...
+	Name    string
+	Package string
+	File    string
+	Line    int
+	Params  []*Node
+	Results []*Node
+}
+
+// Graph is the DOT-format graph in a Container represented by a list of Ctors.
```
abhinav: Same as before: Remove the "list of Ctors" part since that's just listing the field name.

dig_test.go
```diff
@@ -2455,214 +2456,266 @@ func TestFailingFunctionDoesNotCreateInvalidState(t *testing.T) {
...
 
-	t.Run("single edge", func(t *testing.T) {
-		expected := []*dot.Edge{
-			{Param: e1, Result: e2},
-		}
+	pkg := "go.uber.org/dig"
+	file := "dig_test.go"
 
-		c := New()
+	e1, e2, e3, e4 := &dot.Node{Type: "dig.t1"}, &dot.Node{Type: "dig.t2"}, &dot.Node{Type: "dig.t3"}, &dot.Node{Type: "dig.t4"}
```
abhinav: nit: change e1, e2, etc. to n1, n2, etc. since they're nodes, not edges.

dig_test.go
```diff
@@ -2455,214 +2456,266 @@ func TestFailingFunctionDoesNotCreateInvalidState(t *testing.T) {
...
 
-	t.Run("more edges", func(t *testing.T) {
-		expected := []*dot.Edge{
-			{Param: e1, Result: e2},
-			{Param: e1, Result: e3},
-			{Param: e2, Result: e4},
-		}
+	t.Run("constructor with one param and one result", func(t *testing.T) {
+		expected := []*dot.Ctor{{
+			Name:    "TestDotGraph.func1.1",
```
abhinav: Using information from this file in the tests makes them a bit fragile. New
tests in the wrong place could break the tests.

It would be safer to just not check the name of the function here.  Instead,
we can test that the `newDotCtor` function copies the name over from the
`node`. That the `node` has the correct function name is already tested
elsewhere.

dig_test.go
```diff
@@ -2455,214 +2456,266 @@ func TestFailingFunctionDoesNotCreateInvalidState(t *testing.T) {
...
 
 		c := New()
+		c.Provide(func(A t1) t2 { return t2{} })
+		compareCtors(t, expected, c.dg.Ctors)
+	})
+
+	t.Run("more constructors", func(t *testing.T) {
+		expected := []*dot.Ctor{{
+			Name:    "TestDotGraph.func2.1",
+			Package: pkg,
```
abhinav: formatting nit: for complex literals like this, we prefer the `{` to be on its
own line and the `}` to be with the `,`.

```
[]*dot.Ctor{
    {
        Package: pkg,
        Params: []*dot.Node{e1},
        Results: []*dot.Node{e2},
    },
    {
        // ..
    },
}
```

PR #191, opened 2018-06-04
Visualize: Parse graph into DOT format

leyao:
Added `Visualize` method that takes in a `Container` object, parses its dependency graph into DOT format, and writes to an `io.Writer`.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -213,10 +213,7 @@ type VisualizeOption interface {
 	unimplemented()
 }
 
-// Visualize parses the graph in Container c into DOT format and writes it to
-// io.Writer w.
-func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
-	graphTmpl := template.Must(template.New("DotGraph").Parse(`digraph {
+var _graphTmpl = template.Must(template.New("DotGraph").Parse(`digraph {
 	graph [compound=true];
 	{{range $index, $ctor := .Ctors}}
 		subgraph cluster_{{$index}} {
@@ -231,7 +228,10 @@ func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
 	{{end}}
 }`))
 
-	if err := graphTmpl.Execute(w, c.dg); err != nil {
+// Visualize parses the graph in Container c into DOT format and writes it to
+// io.Writer w.
+func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
+	if err := _graphTmpl.Execute(w, c.dg); err != nil {
 		return fmt.Errorf("error executing template: %v", err)
 	}
 	return nil
--- a/dig.go
+++ b/dig.go
@@ -23,11 +23,13 @@ package dig
 import (
 	"errors"
 	"fmt"
+	"io"
 	"math/rand"
 	"reflect"
 	"sort"
 	"strconv"
 	"strings"
+	"text/template"
 	"time"
 
 	"go.uber.org/dig/internal/digreflect"
@@ -203,6 +205,45 @@ func New(opts ...Option) *Container {
 		opt.applyOption(c)
 	}
 	return c
+}
+
+// A VisualizeOption modifies the default behavior of Visualize. It's included for
+// future functionality; currently, there are no concrete implementations.
+type VisualizeOption interface {
+	unimplemented()
+}
+
+// Visualize parses the graph in Container c into DOT format and writes it to
+// io.Writer w. The nodes(types) are clustered by constructors. Clusters and
+// nodes are connected by directed edges in the graph to indicate dependency.
+//
+// For example, if constructor c1 has params p1 and p2, c1 -> t1 and c1 -> t2.
+//
+// Different type attributes are represented differently on the graph -
+//   The names and groups of types are indicated in sublabels.
+//   An optional dependency is indicated by dotted edges to the params.
+func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
+	tmpl := `digraph {
+	graph [compound=true];
+{{range $index, $ctor := .Ctors}}
+	subgraph cluster_{{$index}} {
+		"{{.Name}}" [shape=plaintext];
+{{range $res := .Results}}
+		"{{.String}}" [label=<{{.Type}}{{.Attributes}}>];{{end}}
+	}
+{{range $par := .Params}}
+	"{{$ctor.Name}}" -> "{{.String}}" [ltail=cluster_{{$index}}{{if .Optional}} style=dashed{{end}}];{{end}}
+{{end}}}`
+
+	t := template.New("DOT graph")
+	t, err := t.Parse(tmpl)
+	if err != nil {
+		return fmt.Errorf("error parsing DOT graph template: got %v", err)
+	}
+	if err := t.Execute(w, c.dg); err != nil {
+		return fmt.Errorf("error executing DOT graph template: got %v", err)
+	}
+	return nil
 }
 
 // Changes the source of randomness for the container.
--- a/dig.go
+++ b/dig.go
@@ -207,8 +207,8 @@ func New(opts ...Option) *Container {
 	return c
 }
 
-// A VisualizeOption modifies the default behavior of Visualize. It's included for
-// future functionality; currently, there are no concrete implementations.
+// A VisualizeOption modifies the default behavior of Visualize. It's included
+// for future functionality; currently, there are no concrete implementations.
 type VisualizeOption interface {
 	unimplemented()
 }
--- a/dig.go
+++ b/dig.go
@@ -214,34 +214,25 @@ type VisualizeOption interface {
 }
 
 // Visualize parses the graph in Container c into DOT format and writes it to
-// io.Writer w. The nodes(types) are clustered by constructors. Clusters and
-// nodes are connected by directed edges in the graph to indicate dependency.
-//
-// For example, if constructor c1 has params p1 and p2, c1 -> t1 and c1 -> t2.
-//
-// Different type attributes are represented differently on the graph -
-//   The names and groups of types are indicated in sublabels.
-//   An optional dependency is indicated by dotted edges to the params.
+// io.Writer w.
 func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
-	tmpl := `digraph {
+	graphTmpl := template.Must(template.New("DotGraph").Parse(`digraph {
 	graph [compound=true];
-{{range $index, $ctor := .Ctors}}
-	subgraph cluster_{{$index}} {
-		"{{.Name}}" [shape=plaintext];
-{{range $res := .Results}}
-		"{{.String}}" [label=<{{.Type}}{{.Attributes}}>];{{end}}
-	}
-{{range $par := .Params}}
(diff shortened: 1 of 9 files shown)


dig.go
```diff
@@ -205,6 +207,45 @@ func New(opts ...Option) *Container {
...
 }
 
+// A VisualizeOption modifies the default behavior of Visualize. It's included
+// for future functionality; currently, there are no concrete implementations.
+type VisualizeOption interface {
+	unimplemented()
+}
+
+// Visualize parses the graph in Container c into DOT format and writes it to
+// io.Writer w. The nodes(types) are clustered by constructors. Clusters and
```
abhinav: nit: We don't necessarily need to describe the visualization in the comments.

dig.go
```diff
@@ -205,6 +207,45 @@ func New(opts ...Option) *Container {
...
+//   An optional dependency is indicated by dotted edges to the params.
+func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
+	tmpl := `digraph {
+	graph [compound=true];
+{{range $index, $ctor := .Ctors}}
+	subgraph cluster_{{$index}} {
+		"{{.Name}}" [shape=plaintext];
+{{range $res := .Results}}
+		"{{.String}}" [label=<{{.Type}}{{.Attributes}}>];{{end}}
+	}
```
abhinav: This is okay for now but in a future change, I'd like to move the template
into its own file and use something like go-bindata to embed it into the
binary.

dig.go
```diff
@@ -205,6 +207,45 @@ func New(opts ...Option) *Container {
...
+//   The names and groups of types are indicated in sublabels.
+//   An optional dependency is indicated by dotted edges to the params.
+func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
+	tmpl := `digraph {
+	graph [compound=true];
+{{range $index, $ctor := .Ctors}}
+	subgraph cluster_{{$index}} {
+		"{{.Name}}" [shape=plaintext];
+{{range $res := .Results}}
+		"{{.String}}" [label=<{{.Type}}{{.Attributes}}>];{{end}}
```
abhinav: Rather than structuring the template to have readable output, err on the side
of having a readable template.

What I mean is, it's okay to indent things if it helps readability here.

```
digraph {
    graph [compound=true];
    {{ range $index, $ctor := .Ctors }}
        subgraph cluster_{{$index}} {
            "{{.Name}}" [shape=plaintext];
            {{ range $res := .Results }}
                "{{.String}}" [label=...];
            {{ end }}
            ...
        }
    {{ end }}
}
```

Otherwise it's not clear what starts and ends where.

dig.go
```diff
@@ -205,6 +207,45 @@ func New(opts ...Option) *Container {
...
+//
+// Different type attributes are represented differently on the graph -
+//   The names and groups of types are indicated in sublabels.
+//   An optional dependency is indicated by dotted edges to the params.
+func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
+	tmpl := `digraph {
+	graph [compound=true];
+{{range $index, $ctor := .Ctors}}
+	subgraph cluster_{{$index}} {
+		"{{.Name}}" [shape=plaintext];
```
abhinav: Although unlikely (even impossible?) for `.Name` to contain a `"`, it's safer
to use `printf` with `%q` to let Go do the quoting and escaping for you.

    {{printf "%q" .Name}}
    
Is the same as

    "{{.Name}}"

Except that it'll escape `"`s in `.Name`.

dig.go
```diff
@@ -205,6 +207,45 @@ func New(opts ...Option) *Container {
...
+{{range $res := .Results}}
+		"{{.String}}" [label=<{{.Type}}{{.Attributes}}>];{{end}}
+	}
+{{range $par := .Params}}
+	"{{$ctor.Name}}" -> "{{.String}}" [ltail=cluster_{{$index}}{{if .Optional}} style=dashed{{end}}];{{end}}
+{{end}}}`
+
+	t := template.New("DOT graph")
```
abhinav: nit: Typically template names don't contain spaces. I don't see anything
against that in the docs so this is probably fine but maybe worth changing for
consistency.

dig.go
```diff
@@ -205,6 +207,45 @@ func New(opts ...Option) *Container {
...
+	t, err := t.Parse(tmpl)
+	if err != nil {
+		return fmt.Errorf("error parsing DOT graph template: got %v", err)
+	}
+	if err := t.Execute(w, c.dg); err != nil {
+		return fmt.Errorf("error executing DOT graph template: got %v", err)
```
abhinav: when chaining errors, you can include the original error message after the `:`. The "got" is implied.

    fmt.Errorf("error executing template: %v", err)

dig.go
```diff
@@ -205,6 +207,45 @@ func New(opts ...Option) *Container {
...
+{{range $res := .Results}}
+		"{{.String}}" [label=<{{.Type}}{{.Attributes}}>];{{end}}
+	}
+{{range $par := .Params}}
+	"{{$ctor.Name}}" -> "{{.String}}" [ltail=cluster_{{$index}}{{if .Optional}} style=dashed{{end}}];{{end}}
+{{end}}}`
+
+	t := template.New("DOT graph")
+	t, err := t.Parse(tmpl)
+	if err != nil {
```
abhinav: Seeing as the template doesn't change between invocations, you can parse it once at the package level.

```
var graphTmpl = template.Must(template.New("...").Parse(`
[template goes here]
`))
```

That avoids all this complication inside the visualize function.

(Note: We're using template.Must. That panics if the template fails to parse.
This is okay because if the template failed to parse, there's a serious bug in
*our* code and we should find out about that at start up.)

internal/dot/graph.go
```diff
@@ -42,3 +44,29 @@ type Node struct {
...
+	}
+
+	return n.Type
+}
+
+// Attributes composes and returns a string to style the sublabels when
+// visualizing graph.
+func (n *Node) Attributes() string {
+	if n.Name != "" {
+		return fmt.Sprintf("<BR /><FONT POINT-SIZE=\"10\">Name: %v</FONT>", n.Name)
```
abhinav: Go provides "`" for string literals so that you don't have to escape double quotes.

    return fmt.Sprintf(`<BR /><FONT POINT-SIZE="10">Name: %v</FONT>`, n.Name)

internal/dot/graph.go
```diff
@@ -42,3 +44,29 @@ type Node struct {
...
+	} else if n.Group != "" {
+		return fmt.Sprintf("<BR /><FONT POINT-SIZE=\"10\">Group: %v</FONT>", n.Group)
```
abhinav: Is the `<BR/>` part of the "attributes" of this node?

internal/dot/graph.go
```diff
@@ -42,3 +44,29 @@ type Node struct {
...
+	} else {
+		return ""
```
abhinav: Code style nit: When there are >2 independent code paths based on single
boolean expressions, a `switch` statement without a value is more common.

```go
switch {
case n.Name != "":
    return fmt.Sprintf(...)
case n.Group != "":
    return fmt.Sprintf(...)
default:
    return ""
}
```

visualize_golden_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"bytes"
+	"flag"
+	"io/ioutil"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var generate = flag.Bool("generate", false, "generates output to testdata/ if set")
```
abhinav: Can we use something more specific than `-generate`. That's a bit generic and
has the possibility of conflicting with another flag.

visualize_golden_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+func VerifyVisualization(t *testing.T, testname string, c *Container) {
+	flag.Parse()
```
abhinav: Don't need this. `go test` calls `flag.Parse` for you.

visualize_golden_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	var b bytes.Buffer
+	require.NoError(t, Visualize(c, &b))
+
+	if *generate {
+		err := ioutil.WriteFile("testdata/"+testname+".dot", b.Bytes(), 0644)
```
abhinav: nit: Calculate the output file name once before the `if` block with `filepath.Join.

```
dotFile := filepath.Join("testdata", testname+".dot")
```

dig.go
leyao: Sounds good, we can look into that in the future.

internal/dot/graph.go
leyao: The string produced is to be appended to the labels in the graph. I'm not sure if `<BR />` should be included here, but we only want a newline if the label has a name or group sub-label. Maybe I can rename the method to something like `Sublabel`?

dig.go
```diff
@@ -205,6 +207,36 @@ func New(opts ...Option) *Container {
...
+// for future functionality; currently, there are no concrete implementations.
+type VisualizeOption interface {
+	unimplemented()
+}
+
+// Visualize parses the graph in Container c into DOT format and writes it to
+// io.Writer w.
+func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
+	graphTmpl := template.Must(template.New("DotGraph").Parse(`digraph {
+	graph [compound=true];
```
abhinav: With the `template.Must` business, I suggest putting it in a package-level variable so it happens only once.

```go
var _graphTmpl = template.Must(...)

func Visualize(c *Container, ...) error {
  if err := graphTmpl.Execute(w, c.dg); err != nil {
    ...
  }
  return nil
}
```

dig_test.go
```diff
@@ -2721,3 +2721,57 @@ func TestNewDotCtor(t *testing.T) {
 	assert.Equal(t, []*dot.Node{{Type: "dig.t1"}}, ctor.Params)
 	assert.Equal(t, []*dot.Node{{Type: "dig.t2"}}, ctor.Results)
 }
+
+func TestVisualize(t *testing.T) {
+	n1, n2, n3, n4 := &dot.Node{Type: "t1"}, &dot.Node{Type: "t2"}, &dot.Node{Type: "t3"}, &dot.Node{Type: "t4"}
```
abhinav: nit: putting these four on their own lines would be more readable. you don't gain much by putting it all on one line.

internal/dot/graph.go
```diff
@@ -42,3 +44,30 @@ type Node struct {
 	Optional bool
 	Group    string
 }
+
+// String returns the string representation of a node so the different
+// constructors can refer to the same node. We omit information on the optional
+// field since the same type can be optional to one constructor and required
+// for another.
+func (n *Node) String() string {
```
abhinav: nit: Are we doing anything different here than just implementing the
`Stringer` interface? If not, the doc for this can just be,

    // String implements fmt.Stringer for Node.
    
Or,

    // String returns a user-readable string representation of Node.

leyao: We might want to explain why we omit the `optional` field in the string representation since the `String()` method for `paramSingle` includes that.

dig.go
leyao: Done!

PR #194, opened 2018-06-14
dot Group: Add Group nodes to graph

leyao:
`Node` is now embedded in `Param` and `Result` structs. The new structs include param/result specific fields like `Optional` and `GroupIndex`. 

A new struct `Group` is introduced to represent value groups. It will be a special node in the graph. Constructors that depend on or provide to a value group will have edges entering or leaving the group node. This separates the grouped type from the regular params and reduces the number of edges in the graph.


dig.go
```diff
@@ -215,16 +215,25 @@ type VisualizeOption interface {
 
 var _graphTmpl = template.Must(template.New("DotGraph").Parse(`digraph {
 	graph [compound=true];
+	{{range $g := .Groups}}
+		{{- printf "%q" .String}} [shape=diamond label=<{{.Type}}{{.Attributes}}>];
+		{{range .Results}}
+			{{- printf "%q" $g.String}} -> {{printf "%q" .String}};
```
abhinav: This is probably fine for now but in a future change, we should maybe fix the
repetitive printfs. We can probably add methods to the underlying types.
Something like,

```go
func (n *Node) QuotedString() string {
    return strconv.Quote(n.String())
}
```

(`strconv.Quote(n.String())` is equivalent to `fmt.Sprintf("%q", n.String())`.)

Then in the template, you can do `{{.QuotedString}}` and it'll call the method.

Or you can add a `quote` function to the `template.FuncMap` (see [Template.Funcs][1])
and do `{{quote .String}}`.

[1]: https://golang.org/pkg/text/template/#Template.Funcs

internal/dot/graph.go
```diff
@@ -20,53 +20,166 @@
...
-	File    string
-	Line    int
-	Params  []*Node
-	Results []*Node
+	Name       string
+	Package    string
+	File       string
+	Line       int
+	Params     []*Param
+	GroupParam []*Group
```
abhinav: nit: For consistency, it should probably be GroupParams.

internal/dot/graph.go
```diff
@@ -20,53 +20,166 @@
...
+// Param is a parameter node in the graph.
+type Param struct {
+	*Node
+	Optional bool
+}
+
+// Result is a result node in the graph.
+type Result struct {
+	*Node
+	GroupIndex int
```
abhinav: nit: add a comment here explaining the purpose of GroupIndex because it won't
make much sense to someone who wasn't part of that discussion.

internal/dot/graph.go
```diff
@@ -20,53 +20,166 @@
...
+}
+
+// Group is a group node in the graph.
+type Group struct {
+	Type    reflect.Type
```
abhinav: nit: mention that this is the type of values in the group. Otherwise, it's easy to confuse between `foo` which was added to the group and `[]foo` which is how the values in the group are consumed.

internal/dot/graph.go
```diff
@@ -20,53 +20,166 @@
...
+	return &Group{
+		Type:  k.t,
+		Group: k.g,
+	}
+}
+
+// AddCtor adds the constructor with paramList and resultList into the graph.
+func (dg *Graph) AddCtor(c *Ctor, paramList []*Param, resultList []*Result) {
+	var params []*Param
+	var groupParam []*Group
```
abhinav: style nit: use `var` group

```
var (
  params []*Param
  groupParam []*Group
)
```

Also, it should probably be groupParams

internal/dot/graph.go
```diff
@@ -20,53 +20,166 @@
...
+	// grouped params. For grouped params, we use getGroup to find the actual
+	// group.
+	for _, param := range paramList {
+		if param.Group != "" {
+			k := key{t: param.Type.Elem(), g: param.Group}
+			group := dg.getGroup(k)
+			groupParam = append(groupParam, group)
+		} else {
+			params = append(params, param)
+		}
```
abhinav: nit: Go encourages reducing nesting of primary logic by handling "special" cases first.

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

So this would be,

```
if param.Group == "" {
  // Not a value group.
  params = append(params, param)
  continue
}

k := key{...}
...
groupParams = append(groupParams, group)
```

internal/dot/graph.go
```diff
@@ -20,53 +20,166 @@
...
 }
 
-// String implements fmt.Stringer for Node. We omit the optional field since a
-// type can be optional to one constructor and required for another.
-func (n *Node) String() string {
+// getGroup finds the group by key from the graph. If it is not available, a
+// new group is created and returned.
+func (dg *Graph) getGroup(k key) *Group {
+	if _, ok := dg.Groups[k]; !ok {
+		dg.Groups[k] = NewGroup(k)
```
abhinav: nit: I know it won't matter that much but we'll end up doing the lookup twice in this case. Would be better to do,


```
g := NewGroup(k)
dg.Groups[k] = g
return g
```

internal/dot/graph_test.go
```diff
@@ -21,27 +21,170 @@
...
+
+var p1 = &Param{Node: n1}
+var p2 = &Param{Node: n2}
+var p3 = &Param{Node: n4}
+
+var r1 = &Result{Node: n1}
+var r2 = &Result{Node: n2}
+var r3 = &Result{Node: n3, GroupIndex: 5}
+
+var g1 = &Group{Type: reflect.TypeOf(t1{}), Group: "group1"}
```
abhinav: I think it would be preferable to avoid these globals. It's okay to have duplication of test data for separate tests. We don't want a change here for one test to cause failures in multiple other tests because they all rely on the same test data.

internal/dot/graph.go
```diff
@@ -20,53 +20,166 @@
...
-	Optional bool
-	Group    string
+	Type  reflect.Type
+	Name  string
+	Group string
+}
+
+type key struct {
+	t reflect.Type
+	g string
```
amckinney: We can be more descriptive here and have `group`.

internal/dot/graph.go
```diff
@@ -20,53 +20,166 @@
...
-	Type     string
-	Name     string
-	Optional bool
-	Group    string
+	Type  reflect.Type
+	Name  string
+	Group string
+}
+
+type key struct {
```
amckinney: seeing as we only use `key` for the map of groups now, we might want to rename this `groupKey`.

internal/dot/graph.go
```diff
@@ -20,53 +20,166 @@
...
+
+	for _, result := range resultList {
+		// If the result is a grouped value, we want to update its GroupIndex
+		// and add it to the Group.
+		if result.Group != "" {
+			k := key{t: result.Type, g: result.Group}
+			group := dg.getGroup(k)
+
+			result.GroupIndex = len(group.Results)
+			group.Results = append(group.Results, result)
```
amckinney: nit: This is fine as is, but we should consider encapsulating all of this in another helper `dg.addGroup(k, result)`, `dg.addGroupNode(result)` or something similar. That way we aren't mutating the `group.Results` slice outside of the method call.

param.go
```diff
@@ -48,8 +48,8 @@ type param interface {
 	// This MAY panic if the param does not produce a single value.
 	Build(containerStore) (reflect.Value, error)
 
-	// DotNodes returns a slice of param(s) represented in dot.Nodes for the DOT-format graph.
-	DotNodes() []*dot.Node
+	// DotNodes returns a slice of dot.Param(s).
+	DotNodes() []*dot.Param
```
amckinney: Now that these methods are different for parameters and results, let's consider renaming them to `DotParams` and `DotResults`.

internal/dot/graph.go
leyao: Changed to GroupParams.

internal/dot/graph.go
leyao: sure, I realized it's not really self-explanatory. I added comments explaining GroupIndex.

internal/dot/graph.go
leyao: Added comment.

internal/dot/graph.go
leyao: thanks for including the link. I've seen that elsewhere but I didn't know it was a Go convention. Fixed in next commit!

internal/dot/graph.go
leyao: Good point, I'll change that.

internal/dot/graph_test.go
leyao: I wasn't sure of the best way to do it since there's a lot of repetitive test data needed, but what you said about failing other tests on change makes sense. I've moved them back into individual tests.

internal/dot/graph.go
leyao: sure, `groupKey` definitely is more descriptive.

internal/dot/graph.go
leyao: I added a new `dg.addToGroup(*result)` function; however we would still have to mutate the Result passed in to update the `GroupIndex`.

dig.go
```diff
@@ -213,21 +213,38 @@ type VisualizeOption interface {
 	unimplemented()
 }
 
-var _graphTmpl = template.Must(template.New("DotGraph").Parse(`digraph {
+var funcMap = template.FuncMap{
+	"quote": quote,
```
abhinav: optional: can inline this instead of declaring a new global variable.

```go
var _graphTmpl = template.Must(
  template.New("DotGraph").
    Funcs(template.FuncMap{
    }).
    Parse(`
`),
)
```

(Note that the lines are ending with `.` to indicate that more for this expression is going to follow on the next line.)

dig.go
```diff
@@ -213,21 +213,38 @@ type VisualizeOption interface {
...
+			{{- quote $ctor.Name}} -> {{quote .String}} [ltail=cluster_{{$index}}{{if .Optional}} style=dashed{{end}}];
 		{{end}}
+		{{range .GroupParams}}
+			{{- quote $ctor.Name}} -> {{quote .String}} [ltail=cluster_{{$index}}];
+		{{end -}}
 	{{end}}
 }`))
 
+func quote(s string) string {
+	return strconv.Quote(s)
```
abhinav: nit: You don't need to declare this function. You can do, `"quote": strconv.Quote`.

dig_test.go
```diff
@@ -2468,24 +2468,56 @@ func assertCtorsEqual(t *testing.T, expected []*dot.Ctor, ctors []*dot.Ctor) {
 	}
 }
 
+func testParam(t reflect.Type, n string, g string, o bool) *dot.Param {
```
abhinav: optional: since these are helpers for testing in a test file, they don't need the `test` prefix. You can use a shorter name if you prefer.

Also, it looks like these are only used in TestDotGraph. You can declare anonymous functions and assign them to variables inside TestDotGraph if you're concerned about overly generic names polluting the namespace.

```
tparam := func(....) *dot.Param {
  ...
}

tresult := func(...) *dot.Result {
  ...
}
```

internal/dot/graph.go
```diff
@@ -20,53 +20,183 @@
...
+	Line        int
+	Params      []*Param
+	GroupParams []*Group
+	Results     []*Result
+}
+
+// Param is a parameter node in the graph.
+type Param struct {
+	*Node
+	Optional bool
```
abhinav: style nit: add an empty line after embedded types

```
type Param struct {
   *Node

  Optional bool
}
```

internal/dot/graph.go
```diff
@@ -20,53 +20,183 @@
...
+}
+
+// Result is a result node in the graph.
+type Result struct {
+	*Node
+
+	// GroupIndex is added to differenciate grouped values from one another.
```
abhinav: typo: differentiate

internal/dot/graph.go
```diff
@@ -20,53 +20,183 @@
...
+	// representations are the same so we need indices to uniquely identify
+	// the values.
+	GroupIndex int
+}
+
+// Group is a group node in the graph.
+type Group struct {
+	// Type is the type of values in the group.
+	Type    reflect.Type
+	Group   string
```
abhinav: nit: This should be called Name. (`Group.Name` versus `Group.Group`.)

internal/dot/graph.go
```diff
@@ -20,53 +20,183 @@
...
+	Results []*Result
```
abhinav: nit: Add a comment explaining that these are references to the result nodes provided by *possibly different* constructors.

internal/dot/graph.go
```diff
@@ -20,53 +20,183 @@
...
+
+// Result is a result node in the graph.
+type Result struct {
+	*Node
+
+	// GroupIndex is added to differenciate grouped values from one another.
+	// Since grouped values have the same type and group, their Node / string
+	// representations are the same so we need indices to uniquely identify
+	// the values.
+	GroupIndex int
```
abhinav: nit: Add,

    // This is the index of this object in the corresponding Group's Results list.

dig.go
leyao: Combined this and next comment and made the template a single expression 😄

leyao:
A new struct `Group` is introduced to represent value groups. It will be a special node in the graph. Constructors that depend on or provide to a value group will have edges entering or leaving the group node. This separates them from regular params and reduces the number of edges in the graph.

PR #195, opened 2018-06-22
error: Detect and visualize invoke errors in graph

leyao:
On invoke errors, the graph is updated to reflect the node(s) that failed to build. The root cause(s) of failure will have red borders, while nodes that failed to build due to missing/failed dependencies will be orange.


dig.go
```diff
@@ -648,12 +696,14 @@ func shallowCheckDependencies(c containerStore, p param) error {
...
 		if ns := c.getValueProviders(ps.Name, ps.Type); len(ns) == 0 && !ps.Optional {
 			missing = append(missing, newErrMissingType(c, key{name: ps.Name, t: ps.Type}))
+			missingNodes = append(missingNodes, ps.DotParam()...)
 		}
 
 		return true
 	}))
 
 	if len(missing) > 0 {
+		c.failNodes(missingNodes, 0)
```
abhinav: What does it mean when id is 0? Use of a magic value like this suggests we
should do something else here. Maybe `c.failMissingNodes(missingNodes)`?

dig.go
```diff
@@ -168,6 +168,14 @@ type containerStore interface {
 	// Returns the providers that can produce values for the given group and
 	// type.
 	getGroupProviders(name string, t reflect.Type) []provider
+
+	// Adds the nodes to the list of failed nodes of the graph in the container
+	// and marks the corresponding constructor as failed.
+	failNodes(params []*dot.Param, id uintptr)
```
abhinav: I think it would be preferable to pass the `provider` here since the
`containerStore` deals in `provider`s. The underlying call to the graph can
deal in raw ids.

That theoretically lets you drop the `params` argument too because the
`failNodes` implementation can do `provider.ParamList().DotParam()`.

(We can add a separate `fail*` method for the case when `id` is 0.)

dig.go
```diff
@@ -189,6 +197,8 @@ type provider interface {
 	// The values produced by this provider should be submitted into the
 	// containerStore.
 	Call(containerStore) error
+
+	ID() uintptr
```
abhinav: nit: I would make this the first one in the list since it's the unique identifier.

(Also, consider adding a comment above this.)

    // ID is a unique numerical identifier for this provider.

dig.go
```diff
@@ -311,6 +329,28 @@ func (c *Container) getProviders(k key) []provider {
...
+
+	for i, p := range params {
+		failed[i] = &dot.Result{Node: p.Node}
+	}
+	c.dg.FailNodes(failed, id)
+}
+
+func (c *Container) failGroupNodes(params []*dot.Param, id uintptr) {
+	// Taking first element since there will only be one group node that fails.
+	p := params[0]
```
abhinav: This is fine for now but consider adding a TODO to receive the group name here so that this looks like,

    func (c *Container) failGroupNode(name string, t reflect.Type)

It matches the rest of the containerStore interface.

dig.go
```diff
@@ -221,16 +231,18 @@ var _graphTmpl = template.Must(
...
+		{{- quote .String}} [{{.Attributes}}];
 		{{range .Results}}
 			{{- quote $g.String}} -> {{quote .String}};
 		{{end}}
 	{{end -}}
 	{{range $index, $ctor := .Ctors}}
 		subgraph cluster_{{$index}} {
 			{{quote .Name}} [shape=plaintext];
+			{{- if eq .State 1}}color=red;{{end}}
+			{{- if eq .State 2}}color=orange;{{end}}
```
abhinav: You can add a method on `State` and do,

    {{- with .State }}color={{.Color}};{{ end }}

(`with` will execute the block with `.` set to the provided argument if it's non-zero.)

internal/dot/graph.go
```diff
@@ -25,15 +25,26 @@ import (
 	"reflect"
 )
 
+// State of a constructor or group is updated when they fail to build.
+type State int
+
+const (
+	initial State = iota
+	pointOfFailure
+	transitiveFailure
```
abhinav: nit: FailureState? initial can be `success`, pointOfFailure can be `rootCause` and transitiveFailure can be `failed`.

internal/dot/graph.go
```diff
@@ -25,15 +25,26 @@ import (
...
+)
+
 // Ctor encodes a constructor provided to the container for the DOT graph.
 type Ctor struct {
 	Name        string
 	Package     string
 	File        string
 	Line        int
+	ID          uintptr
```
abhinav: Instead of using raw `uintptr`s, can we declare a type?

    type CtorID uintptr

From the main dig package, you'll do `dot.CtorID(cptr)` to build this.

internal/dot/graph.go
```diff
@@ -60,12 +71,27 @@ type Group struct {
...
-	Ctors  []*Ctor
-	Groups map[groupKey]*Group
+	Ctors   []*Ctor
+	ctorMap map[uintptr]*Ctor
+	Groups  map[groupKey]*Group
+
+	// Pof is a list of the point of failures. They are the root causes of
+	// failed invokes and can be either missing types (not provided) or error
+	// types (error providing).
+	Pof []*Result
```
abhinav: RootCauses?

internal/dot/graph.go
```diff
@@ -60,12 +71,27 @@ type Group struct {
 	Type    reflect.Type
 	Name    string
 	Results []*Result
+	State   State
+
+	// ResultMap maps constructor ID to the grouped result it produces so we
+	// can get the result that failed given constructor ID.
+	ResultMap map[uintptr]*Result
```
abhinav: This is a bit confusing. Are these all constructors that produce values to
this value group? This is a map from those constructors to which results?
(Constructors have multiple results.)

Also, can we avoid having the 'index' here  and just have it in the `Graph`?
This can have a list of IDs of all constructors that provide to this group if
needed.

That lets us keep the book-keeping in one place.

internal/dot/graph.go
```diff
@@ -60,12 +71,27 @@ type Group struct {
...
+	Groups  map[groupKey]*Group
+
+	// Pof is a list of the point of failures. They are the root causes of
+	// failed invokes and can be either missing types (not provided) or error
+	// types (error providing).
+	Pof []*Result
+
+	// Failed is the list of nodes that failed to build due to missing/failed
+	// dependencies.
+	Failed []*Result
```
abhinav: Is there overlap with root causes or are they separate?

internal/dot/graph.go
```diff
@@ -177,27 +253,26 @@ func (g *Group) String() string {
...
 }
 
-// Attributes composes and returns a string to style the Group's sublabel.
+// Attributes composes and returns a string of the Group node's attributes.
 func (g *Group) Attributes() string {
-	return fmt.Sprintf(`<BR /><FONT POINT-SIZE="10">Group: %v</FONT>`, g.Name)
+	attr := fmt.Sprintf(`shape=diamond label=<%v<BR /><FONT POINT-SIZE="10">Group: %v</FONT>>`, g.Type, g.Name)
+	switch g.State {
+	case pointOfFailure:
+		attr += " color=red"
```
abhinav: this appears to be repeating some of the logic from `if state == 1`. The `Color` method would help here too.

dig.go
leyao: I've added a `missingNodes` function that deals with nodes with no constructors.

dig.go
leyao: We need to pass the specific `param` in since we want to colour only the node that was required. If `provider.ParamList().DotParam()` was used instead, every node in the constructor will be coloured. That is technically correct since every node in the constructor failed to build, but I think is less desirable since we want to be able to tell which nodes are causing the failure.

What do you think is the desired behaviour?

abhinav: Ah, that makes sense. This is fine as-is, then.

dig.go
leyao: Similar to above, we need the constructor id to determine the node that failed, so we don't end up colouring the entire group. I changed the function to 
```go
func (c *Container) failGroupNode(name string, t reflect.Type, id uintptr)
```

dig.go
leyao: Great, this looks a lot nicer 👍

internal/dot/graph.go
leyao: Renamed the failure states to `success`, `rootCause`, and `failed` everywhere.

internal/dot/graph.go
leyao: No, they are disjoint sets.

dig.go
```diff
@@ -168,10 +168,23 @@ type containerStore interface {
 	// Returns the providers that can produce values for the given group and
 	// type.
 	getGroupProviders(name string, t reflect.Type) []provider
+
+	// Finds and adds the node to the list of failed nodes of the graph in the
+	// container and mark the corresponding constructor and group as failed.
```
amckinney: nit: mark -> marks

dig.go
```diff
@@ -168,10 +168,23 @@ type containerStore interface {
...
+	failGroupNodes(name string, t reflect.Type, id dot.CtorID)
```
amckinney: nit: consider updating the signature here to
```
failGroupNodes(t reflect.Type, group string, id dot.CtorID)
```

That way it is consistent with the `failNodes` signature (types first, then group / ids after).

dig.go
```diff
@@ -311,6 +331,28 @@ func (c *Container) getProviders(k key) []provider {
 	return providers
 }
 
+func (c *Container) missingNodes(params []*dot.Param) {
+	missing := make([]*dot.Result, len(params))
```
amckinney: Why do we need to map the `dot.Param` to a `dot.Result` here?

dig.go
```diff
@@ -552,6 +594,10 @@ type node struct {
 	// Location where this function was defined.
 	location *digreflect.Func
 
+	// id is the address of memory location for the constructor. It is used to
```
amckinney: nit: We can remove the "the address ..." note and just have something similar to:
```
id uniquely identifies the constructor that produces a node.
```

dig.go
```diff
@@ -648,12 +698,14 @@ func shallowCheckDependencies(c containerStore, p param) error {
...
 		if ns := c.getValueProviders(ps.Name, ps.Type); len(ns) == 0 && !ps.Optional {
 			missing = append(missing, newErrMissingType(c, key{name: ps.Name, t: ps.Type}))
+			missingNodes = append(missingNodes, ps.DotParam()...)
 		}
 
 		return true
 	}))
 
 	if len(missing) > 0 {
+		c.missingNodes(missingNodes)
```
amckinney: It might be better to update the name to `addMissingNodes` so that it's more clear that there is some mutation happening here.

internal/dot/graph.go
```diff
@@ -25,15 +25,36 @@ import (
 	"reflect"
 )
 
+// FailureState of a constructor or group is updated when they fail to build.
+type FailureState int
```
amckinney: This conflicts with @abhinav's comment, but seeing as `success` is a valid `FailureState`, we should consider renaming to just `State`. The different states could then be:

```
const (
  success State = iota
  transitiveFailure
  rootFailure
)
```

@abhinav + @leyao What do you think? I'm fine with either way.

internal/dot/graph.go
```diff
@@ -60,19 +81,29 @@ type Group struct {
...
+
+	Failed *failedNodes
 }
 
-// Node is a single node in a graph and is embedded into Params and Results.
-type Node struct {
-	Type  reflect.Type
-	Name  string
-	Group string
+type failedNodes struct {
```
amckinney: It looks like all of the failures are of type `Result` - should they be of type `Node`? If `Result` is more desirable, could you explain? I'm fine with `Result`, but want to make sure it's necessary.

internal/dot/graph.go
```diff
@@ -130,21 +163,84 @@ func (dg *Graph) AddCtor(c *Ctor, paramList []*Param, resultList []*Result) {
...
+func (dg *Graph) failNode(r *Result, pof bool) {
+	if pof {
+		dg.addRootCause(r)
+	} else {
+		dg.addTransitiveFailure(r)
+	}
+}
+
+// MissingNodes adds missing nodes to the list of failed Results in the graph.
+func (dg *Graph) MissingNodes(results []*Result) {
```
amckinney: Similar to the above comment, if this is "adding" nodes, we should write `AddMissingNodes`.

internal/dot/graph.go
```diff
@@ -130,21 +163,84 @@ func (dg *Graph) AddCtor(c *Ctor, paramList []*Param, resultList []*Result) {
...
+
+	for _, r := range results {
+		dg.failNode(r, pof)
+	}
+}
+
+// FailNodes adds results to the list of failed Results in the graph, and
+// updates the state of the constructor with the given id accordingly.
+func (dg *Graph) FailNodes(results []*Result, id CtorID) {
+	pof := len(dg.Failed.RootCauses) == 0
```
amckinney: When is `len(dg.Failed.RootCauses) != 0`? It seems like once a failure is encountered, there will always be at least one root (that therefore create transitive failures). If this is the case, after the initial root failure is added, this will always be > 0, and everything would be colored as red since `pof` is true. 

I'm assuming that transitive failures are appropriately colored, however, so what piece am I misunderstanding?

dig.go
leyao: In order to pinpoint which grouped node failed in the graph, we need the resulting failed nodes to be `dot.Result`s so the `GroupIndex` is included.

internal/dot/graph.go
leyao: As above, `GroupIndex` in `Result` is needed to identify which grouped node failed.

internal/dot/graph.go
leyao: `pof` is true when there are no existing root causes. Once the first failed node(s) are recorded, they are added to `dg.Failed.RootCauses`, and all other errors will have `pof = false` and be added as transitive failures.

dig.go
leyao: Updated to `addMissingNodes`.

internal/dot/graph.go
leyao: Changed to `AddMissingNodes` too.

internal/dot/graph.go
abhinav: Fair point, although there is precedence for "zero value is different" when it
comes to iota. For example, in yarpcerrors, [CodeOK] means "no error" even
though `Code` represents the error code.

We could rename this to `ErrorType` and `success` to `noError`, which might
make more sense as a zero value then, but I don't feel strongly about it.
I'm open to other names too. I wanted to avoid `State` because it's pretty
generic on its own. `Ctor.State` doesn't tell you what kind of state we're
talking about. The `Ctor` struct is the state of the Ctor already.

internal/dot/graph.go
abhinav: Can we add a quick comment about this in the places where we're doing `pof := `

    // This failure is the root cause if there are no other failures.
    pof := len(...) == 0

internal/dot/graph.go
```diff
@@ -60,19 +81,29 @@ type Group struct {
...
-// Node is a single node in a graph and is embedded into Params and Results.
-type Node struct {
-	Type  reflect.Type
-	Name  string
-	Group string
+type failedNodes struct {
+	// RootCauses is a list of the point of failures. They are the root causes
+	// of failed invokes and can be either missing types (not provided) or
+	// error types (error providing).
+	RootCauses []*Result
```
abhinav: Will this ever have a length more than one? It'll either be zero or one, right?

This is fine for now but in a future change we probably can simplify the internal representation to be,

    // Results that failed to build, in-order. The first result is the root
    // cause, the second result failed to build because the first was
    // unavailable, and so on.
    FailureChain []*Result

internal/dot/graph.go
amckinney: Yea that's a good point. I'd lean towards your `ErrorType` suggestion in this case.

internal/dot/graph.go
amckinney: Cool, makes sense.

internal/dot/graph.go
amckinney: +1 makes sense. The comment would be helpful.

internal/dot/graph.go
amckinney: +1 here - it was fairly confusing at first glance to see a slice of roots.

internal/dot/graph.go
leyao: I added comments, also renamed `pof` to `isRootCause`.

dig.go
```diff
@@ -569,6 +615,7 @@ type nodeOptions struct {
 
 func newNode(ctor interface{}, opts nodeOptions) (*node, error) {
 	ctype := reflect.TypeOf(ctor)
+	cptr := reflect.ValueOf(ctor).Pointer()
```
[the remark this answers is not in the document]
leyao: That's cool, changed it!

dig_test.go
```diff
@@ -2856,4 +2859,149 @@ func TestVisualize(t *testing.T) {
 
 		VerifyVisualization(t, "grouped", c)
 	})
+
+	t.Run("error types", func(t *testing.T) {
+		c := New()
+
+		type in struct {
+			In
+			C []t1 `group:"g1"`
```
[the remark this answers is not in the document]
leyao: Fixed!

dig_test.go
```diff
@@ -2856,4 +2859,149 @@ func TestVisualize(t *testing.T) {
...
+		}
+
+		type out struct {
+			Out
+			B t3 `name:"n3"`
+			C t2 `group:"g2"`
+		}
+
+		type in1 struct {
```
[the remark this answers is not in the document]
leyao: Fixed!

internal/dot/graph.go
leyao: `ErrorType` sounds good, I'll go with that then 👍 Also updated `failed` to `transitiveFailure` for clarity and to match `ErrorType`.

internal/dot/graph.go
```diff
@@ -60,19 +81,29 @@ type Group struct {
...
 type Graph struct {
-	Ctors  []*Ctor
-	Groups map[groupKey]*Group
+	Ctors   []*Ctor
+	ctorMap map[CtorID]*Ctor
+
+	Groups   []*Group
+	groupMap map[groupKey]*Group
+
+	Failed *failedNodes
```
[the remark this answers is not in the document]
leyao: Exported.

internal/dot/graph.go
leyao: There will be multiple root causes in the case of multiple missing types (`errMissingManyTypes`). IMO it would only make sense if all of them are marked red.

PR #196, opened 2018-06-29
 VisualizeError: Move updates to graph into errors

leyao:
Made `Visualize` and `VisualizeError` separate methods. 
```go
Visualize(*Container, io.Writer, ...VisualizeOption) error
```
`Visualize` takes a container and parses the dependency graph in the container into a DOT format graph with no error detection/representation, and writes it to an `io.Writer`.
```go
VisualizeError(error, io.Writer, ...VisualizeOption) error
```
 `VisualizeError` takes in an error that is generated on `Invoke` failure. It parses the dependency graph included in the error and updates the graph to reflect failing nodes, and writes it to an `io.Writer`.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -168,10 +168,22 @@ type containerStore interface {
 	// Returns the providers that can produce values for the given group and
 	// type.
 	getGroupProviders(name string, t reflect.Type) []provider
+
+	// Marks the given value group as failed because of the given constructor.
+	failGroupNodes(name string, t reflect.Type, id dot.CtorID)
+
+	// Marks the given parameters as failed because of the given constructor.
+	failNodes(params []*dot.Param, id dot.CtorID)
+
+	// Marks the given parameters as missing.
+	addMissingNodes(params []*dot.Param)
 }
 
 // provider encapsulates a user-provided constructor.
 type provider interface {
+	// ID is a unique numerical identifier for this provider.
+	ID() dot.CtorID
+
 	// Location returns where this constructor was defined.
 	Location() *digreflect.Func
 
@@ -221,24 +233,31 @@ var _graphTmpl = template.Must(
 		Parse(`digraph {
 	graph [compound=true];
 	{{range $g := .Groups}}
-		{{- quote .String}} [shape=diamond label=<{{.Type}}{{.Attributes}}>];
+		{{- quote .String}} [{{.Attributes}}];
 		{{range .Results}}
 			{{- quote $g.String}} -> {{quote .String}};
 		{{end}}
 	{{end -}}
 	{{range $index, $ctor := .Ctors}}
 		subgraph cluster_{{$index}} {
-			{{quote .Name}} [shape=plaintext];
+			constructor_{{$index}} [shape=plaintext label={{quote .Name}}];
+			{{- with .ErrorType}}color={{.Color}};{{end}}
 			{{range .Results}}
-				{{- quote .String}} [label=<{{.Type}}{{.Attributes}}>];
+				{{- quote .String}} [{{.Attributes}}];
 			{{end}}
 		}
 		{{range .Params}}
-			{{- quote $ctor.Name}} -> {{quote .String}} [ltail=cluster_{{$index}}{{if .Optional}} style=dashed{{end}}];
+			constructor_{{$index}} -> {{quote .String}} [ltail=cluster_{{$index}}{{if .Optional}} style=dashed{{end}}];
 		{{end}}
 		{{range .GroupParams}}
-			{{- quote $ctor.Name}} -> {{quote .String}} [ltail=cluster_{{$index}}];
+			constructor_{{$index}} -> {{quote .String}} [ltail=cluster_{{$index}}];
 		{{end -}}
+	{{end}}
+	{{range .Failed.TransitiveFailures}}
+		{{- quote .String}} [color=orange];
+	{{end -}}
+	{{range .Failed.RootCauses}}
+		{{- quote .String}} [color=red];
 	{{end}}
 }`))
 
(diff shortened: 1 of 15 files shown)


visualize_golden_test.go
```diff
@@ -53,3 +53,24 @@ func VerifyVisualization(t *testing.T, testname string, c *Container) {
...
+	}
+
+	wantBytes, err := ioutil.ReadFile(dotFile)
+	require.NoError(t, err)
+
+	got := b.String()
+	want := string(wantBytes)
+	assert.Equal(t, want, got,
+		"Output did not match. Make sure you updated the testdata by running 'go test -generate'")
+}
```
amckinney: This looks like it creates a file with the bytes written into the `bytes.Buffer` and asserts that the result of reading that file are the same bytes. Will this ever fail?

Did you instead mean to compare this against some fixture?

dig.go
```diff
@@ -265,12 +256,61 @@ var _graphTmpl = template.Must(
 // Visualize parses the graph in Container c into DOT format and writes it to
 // io.Writer w.
 func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
-	if err := _graphTmpl.Execute(w, c.dg); err != nil {
+	dg := c.createGraph()
+
+	if err := _graphTmpl.Execute(w, dg); err != nil {
```
amckinney: nit: If there is enough context in the error message (without the additional formatting), we could write: 
```
func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
  return _graphTmpl.Execute(w, c.createGraph())
}
```

What do you think?

dig.go
```diff
@@ -265,12 +256,61 @@ var _graphTmpl = template.Must(
...
 	}
 	return nil
 }
 
+// VisualizeError parses the container and errors included in err into a DOT
+// format graph and writes it to io.Writer w.
+func VisualizeError(err error, w io.Writer, opts ...VisualizeOption) error {
+	ep, ok := err.(errEntryPoint)
+	if !ok {
+		return fmt.Errorf("No graph included in error: %v", err)
```
amckinney: nit: errors shouldn't begin with capitalization.

dig.go
```diff
@@ -265,12 +256,61 @@ var _graphTmpl = template.Must(
...
+		}
+		err = e.cause()
+	}
+
+	if len(errors) == 0 {
+		return fmt.Errorf("cannot visualize error: %v", err)
+	}
+
+	dg := ep.container().createGraph()
+	for i := len(errors) - 1; i >= 0; i-- {
```
amckinney: Any particular reason for the reverse iteration?

dig.go
```diff
@@ -265,12 +256,61 @@ var _graphTmpl = template.Must(
...
+		errors[i].updateGraph(dg)
+	}
+
+	if err := _graphTmpl.Execute(w, dg); err != nil {
+		return fmt.Errorf("error executing template: %v", err)
+	}
+
+	return nil
```
amckinney: Similar to the above comment, we could write: `return _graphTmpl.Execute(w, dg)` if the additional context isn't required.

It might be helpful / necessary though, so feel free to ignore if so.

error.go
```diff
@@ -112,25 +113,31 @@ func (e errConstructorFailed) Error() string {
 // errArgumentsFailed is returned when a function could not be run because one
 // of its dependencies failed to build for any reason.
 type errArgumentsFailed struct {
+	c      *Container
```
amckinney: We can export these fields and make them more descriptive:
```
type errArgumentsFailed struct {
  Container *Container
  ...
}
```

visualize_golden_test.go
leyao: The bytes will only be written to the file if the `generate` flag is on when running `go test`, otherwise it will be compared against existing files in the `testdata/` directory.

dig.go
leyao: That's neat. The error itself is indeed pretty self-explanatory and I removed error formatting.

dig.go
leyao: Yes. As we unwrap the errors the outermost error becomes the first element in `errors`, and the innermost error becomes the last element. We want to call `updateGraph` on the innermost error first since it is the root cause, so we iterate in reverse.

dig.go
leyao: As above, removed error formatting.

error.go
leyao: Renamed and made exported!

dig.go
```diff
@@ -265,10 +256,50 @@ var _graphTmpl = template.Must(
...
-		return fmt.Errorf("error executing template: %v", err)
+	return _graphTmpl.Execute(w, c.createGraph())
+}
+
+// VisualizeError parses the container and errors included in err into a DOT
+// format graph and writes it to io.Writer w.
+func VisualizeError(err error, w io.Writer, opts ...VisualizeOption) error {
+	ep, ok := err.(errEntryPoint)
+	if !ok {
+		return fmt.Errorf("no graph included in error: %v", err)
```
abhinav: Nice. I think we may be able to drop this if we make `VisualizeError` an
option as discussed offline earlier. So,

    Visualize(c, w)  // visualizes the container
    Visualize(c, w, VisualizeError(err))  // visualizes the error in the container visualization
    
That way, the container is always available.

Thoughts? @amckinney @leyao

dig.go
```diff
@@ -265,10 +256,50 @@ var _graphTmpl = template.Must(
...
+		}
+		e, ok := err.(causer)
+		if !ok {
+			break
+		}
+		err = e.cause()
+	}
+
+	if len(errors) == 0 {
+		return fmt.Errorf("cannot visualize error: %v", err)
```
abhinav: With VisualizeError being an option, "nothing to visualize" can be a no-op instead of error.

dig.go
```diff
@@ -265,10 +256,50 @@ var _graphTmpl = template.Must(
...
+// format graph and writes it to io.Writer w.
+func VisualizeError(err error, w io.Writer, opts ...VisualizeOption) error {
+	ep, ok := err.(errEntryPoint)
+	if !ok {
+		return fmt.Errorf("no graph included in error: %v", err)
 	}
-	return nil
+
+	var errors []errVisualizer
+
```
abhinav: nit: delete empty line. The contents of the declared  `errors` are being built
in the block so they should be together.

dig.go
abhinav: We should add a comment here about this.

visualize_golden_test.go
amckinney: *facepalm* Ah yea, my mistake. Thanks for clarifying!

dig.go
amckinney: Cool. Might be worth adding a simple in-line comment here so that this behavior isn't accidentally changed in the future.

dig_test.go
```diff
@@ -2865,28 +2889,33 @@ func TestVisualize(t *testing.T) {
...
 		type in1 struct {
 			In
+
 			A []t2 `group:"g2"`
 			B t3   `name:"n3"`
 		}
 
 		type out1 struct {
 			Out
+
```
abhinav: 👍

dig.go
amckinney: I was actually just making the same comment :) - let's go with the option.

amckinney: But as @abhinav suggests, let's handle this in a follow-up.

leyao: Great suggestion, that way we can avoid passing around the container in the errors. I'll work on this next!

PR #202, opened 2018-07-16
CanVisualizeError: Determine if error is visualizable

leyao:
`dig.CanVisualizeError(err)` returns true if err is visualizable. This is used in fx to determine if we want to output the graph given an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Added `Visualize` function to parse the graph in a container into DOT format
+  to allow visualization of errors types dependency relationship of types in
+  container.
+- Added `CanVisualizeError` function to determine if an error can be visualized
+  in the graph.
 - Added `Name` option for `Provide` to add named values to the container
   without rewriting constructors. See package documentation for more
   information.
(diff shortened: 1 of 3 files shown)


dig.go
```diff
@@ -320,6 +321,12 @@ func Visualize(c *Container, w io.Writer, opts ...VisualizeOption) error {
 	return _graphTmpl.Execute(w, dg)
 }
 
+// CanVisualizeError returns true if the error is an errVisualizer.
+func CanVisualizeError(err error) bool {
+	_, ok := err.(errVisualizer)
+	return ok
```
abhinav: I think we also want to handle the case where the outer error isn't
visualizable but the inner one  is. So the logic should be more like "is this
thing or anything wrapped by this an errVisualizer"

```
for {
    if errVisualizer { return true }
    if !errCauser { return false }
    err = err.Cause()
}
```

leyao: Good catch, updated to include all layers of errors.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Added `Visualize` function to parse the graph in a container into DOT format
```
breerly: What is DOT format exactly?

abhinav: We can clarify that in the changelog. How about,

    Added `Visualize` function to visualize the state of the container in the
    GraphViz DOT format.

leyao: I've updated the changelog to provide more information!

PR #204, opened 2018-07-24
test: Ordering of tags is irrelevant

abhinav:
This adds an explicit test for the bug reported in #635 to verify that
the ordering of struct tags is irrelevant.


abhinav:
Good call on the added test. Updated to test both, absent and present cases
for the named value.

PR #205, opened 2018-08-16
dot: graph representation of dependency relations in container

leyao:
`Visualize` function visualizes the state of the container in the GraphViz DOT format. This is useful for debugging application startup failures as it allows the visualization of error types and the dependency relations of types in the container.


internal/dot/graph.go
```diff
@@ -0,0 +1,319 @@
...
+
+	Failed *FailedNodes
+}
+
+// FailedNodes is the nodes that failed in the graph.
+type FailedNodes struct {
+	// RootCauses is a list of the point of failures. They are the root causes
+	// of failed invokes and can be either missing types (not provided) or
+	// error types (error providing).
+	RootCauses []*Result
```
amckinney: I could be mistaken, but it appears that there will only ever be a single `RootCause` - appending to the internal slice is guarded by a `isRootCause := len(dg.Failed.RootCauses) == 0`. I don't think we need a slice in this case.

This is not at all actionable - things will work exactly the same regardless :)

PR #207, opened 2018-09-12
Prevent wasteful recursion by counting visits to params in verifyAcyclic

dnathe4th:
In a private test suite I instantiate a FX application in each test to execute blackbox functionality assessment. Logging the `detectCycles` call yielded the following data:

```
[dn-fx] dnathe4th@~/gopath/src/xxx: grep "detectCycles" dig.log | wc -l
 23991368
[dn-fx] dnathe4th@~/gopath/src/xxx: grep "detectCycles" dig.log | sort | uniq -c | sort -nr | head -n 5
9229660 detectCycles with (dot.CtorID=86815168)
6579496 detectCycles with (dot.CtorID=95051312)
3842944 detectCycles with (dot.CtorID=98789856)
1543708 detectCycles with (dot.CtorID=98919152)
```

Nearly 24M calls to detectCycles were made, with the top parameter experiencing over 9M calls. This `dig` logic accounted for ~60% of my test suite execution time.

With the changes from this PR:
```
[dn-fx] dnathe4th@~/gopath/src/xxx: grep "detectCycles" dig.log | wc -l
    9232
```
Only OVER9000 calls are made once the redundant recursions are avoided.

### DIFF
--- a/param.go
+++ b/param.go
@@ -98,9 +98,32 @@ type paramVisitor interface {
 // recursed into.
 type paramVisitorFunc func(param) (recurse bool)
 
-func (f paramVisitorFunc) Visit(p param) paramVisitor {
-	if f(p) {
-		return f
+type paramVisitorAtMostTwice struct {
+	f       paramVisitorFunc
+	visited map[string]int
+}
+
+// NewParamVisitorAtMostTwice constructs a paramVisitor with statekeeping to avoid
+// redundant recursions.
+func NewParamVisitorAtMostTwice(visitorCounts map[string]int, f paramVisitorFunc) paramVisitor {
+	return paramVisitorAtMostTwice{
+		f:       f,
+		visited: visitorCounts,
+	}
+}
+
+// Visit operates pv.f on params while accumulating a count of operations run.
+// Because pv.f is doing cycle detection (on a list of entries not in scope here)
+// it must be allowed to operate twice on the same parameter during a recurse.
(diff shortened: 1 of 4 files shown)


dnathe4th:
Actually friends the longer I think about this the more I am convinced this is an overly optimistic approach. Let me take another stab actually memorizing `detectCycles` itself, returning the path, in the am.

glibsm:
I knew this day would come :( I'm actually surprised `detectCycles` has survived this long

Caching the subtrees should be a pretty good optimization (assuming they are correctly invalidated on insert).

param.go
```diff
@@ -98,9 +98,32 @@ type paramVisitor interface {
...
+	return paramVisitorAtMostTwice{
+		f:       f,
+		visited: visitorCounts,
+	}
+}
+
+// Visit operates pv.f on params while accumulating a count of operations run.
+// Because pv.f is doing cycle detection (on a list of entries not in scope here)
+// it must be allowed to operate twice on the same parameter during a recurse.
+// There is no usecase presently requiring no executions of pv.f on params beyond
```
zmt: nit: "no usecase presently requiring no executions" - double neg. - which is it?

dnathe4th:
Reopening with some revised logic. The counting was a red herring for the fact that I was tracking visits to non-edge nodes, and their `String()` method was not unique (specifically, a `paramObject` with a single field `String()`'s the same as ` paramSingle` of that parameter).

PR #208, opened 2018-09-13
Prevent wasteful recursion by tracking param visits in detectCycles

dnathe4th:
This is an iteration on #207

### DIFF
--- a/cycle.go
+++ b/cycle.go
@@ -73,16 +73,6 @@ func detectCycles(n provider, c containerStore, path []cycleEntry, visited map[k
 			return false
 		}
 
-		// Don't check for cycles if we've already verified that this param
-		// doesn't introduce cycles.
-		if pk, ok := param.(keyer); ok {
-			k := pk.Key()
-			if _, ok := visited[k]; ok {
-				return false
-			}
-			visited[k] = struct{}{}
-		}
-
 		var (
 			k         key
 			providers []provider
@@ -90,10 +80,18 @@ func detectCycles(n provider, c containerStore, path []cycleEntry, visited map[k
 		switch p := param.(type) {
 		case paramSingle:
 			k = key{name: p.Name, t: p.Type}
+			if _, ok := visited[k]; ok {
+				// We've already checked the dependencies for this type.
+				return false
+			}
 			providers = c.getValueProviders(p.Name, p.Type)
 		case paramGroupedSlice:
 			// NOTE: The key uses the element type, not the slice type.
 			k = key{group: p.Group, t: p.Type.Elem()}
+			if _, ok := visited[k]; ok {
+				// We've already checked the dependencies for this type.
+				return false
+			}
 			providers = c.getGroupProviders(p.Group, p.Type.Elem())
 		default:
 			// Recurse for non-edge params.
@@ -111,6 +109,7 @@ func detectCycles(n provider, c containerStore, path []cycleEntry, visited map[k
 			return false
 		}
 
+		visited[k] = struct{}{}
 		for _, n := range providers {
 			if e := detectCycles(n, c, append(path, entry), visited); e != nil {
 				err = e
--- a/cycle.go
+++ b/cycle.go
@@ -56,21 +56,31 @@ func (e errCycleDetected) Error() string {
 }
 
 func verifyAcyclic(c containerStore, n provider, k key) error {
-	visitor := newParamVisitOnce()
+	visited := make(map[key]struct{})
 	err := detectCycles(n, c, []cycleEntry{
 		{Key: k, Func: n.Location()},
-	}, visitor)
+	}, visited)
 	if err != nil {
 		err = errWrapf(err, "this function introduces a cycle")
 	}
 	return err
 }
 
-func detectCycles(n provider, c containerStore, path []cycleEntry, visitor paramVisitorSetter) error {
+func detectCycles(n provider, c containerStore, path []cycleEntry, visited map[key]struct{}) error {
 	var err error
-	walkParam(n.ParamList(), visitor.With(func(param param) bool {
+	walkParam(n.ParamList(), paramVisitorFunc(func(param param) bool {
 		if err != nil {
 			return false
+		}
+
+		// Don't check for cycles if we've already verified that this param
+		// doesn't introduce cycles.
+		if pk, ok := param.(keyer); ok {
+			k := pk.Key()
+			if _, ok := visited[k]; ok {
+				return false
+			}
+			visited[k] = struct{}{}
 		}
 
 		var (
@@ -102,7 +112,7 @@ func detectCycles(n provider, c containerStore, path []cycleEntry, visitor param
 		}
 
 		for _, n := range providers {
-			if e := detectCycles(n, c, append(path, entry), visitor); e != nil {
+			if e := detectCycles(n, c, append(path, entry), visited); e != nil {
 				err = e
 				return false
 			}
--- a/cycle.go
+++ b/cycle.go
@@ -56,7 +56,7 @@ func (e errCycleDetected) Error() string {
 }
 
 func verifyAcyclic(c containerStore, n provider, k key) error {
-	visited := make(map[string]bool)
+	visited := make(map[string]struct{})
 	err := detectCycles(n, c, []cycleEntry{
 		{Key: k, Func: n.Location()},
 	}, visited)
@@ -66,7 +66,7 @@ func verifyAcyclic(c containerStore, n provider, k key) error {
 	return err
 }
 
-func detectCycles(n provider, c containerStore, path []cycleEntry, visited map[string]bool) error {
+func detectCycles(n provider, c containerStore, path []cycleEntry, visited map[string]struct{}) error {
 	var err error
 	walkParam(n.ParamList(), newParamVisitOnce(visited, func(param param) bool {
 		if err != nil {
--- a/cycle.go
+++ b/cycle.go
@@ -68,7 +68,7 @@ func verifyAcyclic(c containerStore, n provider, k key) error {
 
 func detectCycles(n provider, c containerStore, path []cycleEntry, visited map[string]bool) error {
 	var err error
-	walkParam(n.ParamList(), NewParamVisitOnce(visited, func(param param) bool {
+	walkParam(n.ParamList(), newParamVisitOnce(visited, func(param param) bool {
 		if err != nil {
 			return false
 		}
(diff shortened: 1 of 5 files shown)


dnathe4th:
I will do some benchmarking after lunch

dnathe4th:
In a somewhat contrived benchmark I am about to push onto this branch, I got the following numbers:

```

**Comment by dnathe4th** [comment]
I updated the benchmark test to be much more punishing on the unoptimized implementation. The results are now as follows:

Branch:
```
[dn-memoize-cycle-detection] dnathe4th@~/gopath/src/go.uber.org/dig: go test -bench=. -run BenchmarkProvideCycleDetection
goos: darwin
goarch: amd64
pkg: go.uber.org/dig
BenchmarkProvideCycleDetection-8   	    2000	    815564 ns/op
PASS
ok  	go.uber.org/dig	1.736s
```

Master:
```
[master] dnathe4th@~/gopath/src/go.uber.org/dig: go test -bench=. -run BenchmarkProvideCycleDetection
goos: darwin
goarch: amd64
pkg: go.uber.org/dig
BenchmarkProvideCycleDetection-8   	     100	  20065544 ns/op
PASS
ok  	go.uber.org/dig	2.041s
```

File: dig_test.go [hunk 1]
```diff
@@ -3049,3 +3049,107 @@ func TestCanVisualizeError(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkProvideCycleDetection(b *testing.B) {
```
zmt: Yes, thank you. We should have this anyway. Perhaps we should land the benchmark first as a separate PR so we have them on master/dev and can more easily check against various optimization implementations?

param.go
```diff
@@ -98,9 +98,36 @@ type paramVisitor interface {
...
+// Visit operates pv.f on param if:
+//		- param is not an edge node (i.e. param is a paramList), OR
+//		- param is an edge node that has not yet been visited
+func (pv paramVisitOnce) Visit(p param) paramVisitor {
+	switch p.(type) {
+	case paramSingle, paramGroupedSlice:
+		if pv.visited[p.String()] {
+			return pv
+		}
+		defer func() {
```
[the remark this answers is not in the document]
dnathe4th: This is an interesting though. Instead of the defer it appears to be just as valid to set the `visited` map before returning `pv` a few lines below. There is no real value is setting `visited` if we have already found a cycle (and therefore `pv.f()` is returning `false` to indicate no further recursion necessary.

As for the Annotation pattern from that comment, while I did not initially read that comment I still don't feel like this is the right approach. The utility of `paramVisitor` interface is in its usage by `walkParam`. The only argument or adding to this interface is if `walkParam` needs a new way to interact with the `paramVisitor` implementation. I would argue the encapsulation of the caching functionality inside the `paramVisitOnce` implementation while satisfying the `paramVisitor` interface is cleaner than providing explicit caching hooks. Maybe I am wrong though!

cycle.go
```diff
@@ -91,15 +92,16 @@ func detectCycles(n provider, c containerStore, path []cycleEntry) error {
...
 
-		for _, p := range path {
-			if p.Key == k {
-				err = errCycleDetected{Path: append(path, entry)}
-				return false
-			}
+		// The first element of path is the new addition to the graph, therefore
+		// it must be in any cycle that exists, assuming verifyAcyclic has been
+		// run for every previous Provide.
+		if path[0].Key == k {
```
abhinav: nice catch.

I would add a comment reminding readers that path will always have at least one item.

dig_test.go
dnathe4th: Resolving this because I added this Benchmark in #210, although our difference in results is yet unexplained.

dig.go
```diff
@@ -753,7 +753,7 @@ func isFieldOptional(f reflect.StructField) (bool, error) {
 func shallowCheckDependencies(c containerStore, p param) error {
 	var missing errMissingManyTypes
 	var addMissingNodes []*dot.Param
-	walkParam(p, paramVisitorFunc(func(p param) bool {
+	walkParam(p, newParamVisitOnce(make(map[string]struct{}), func(p param) bool {
```
abhinav: I don't think you need this here because there's no recursion on new
dependencies here. (The only time we have recursion here is when we have non-
leaf nodes: paramList, paramGroup.)

This checks the direct descendants of a parameter so checking for repeated
visits of nodes will be beneficial only if you had `func(io.Reader,
io.Reader)` or similar.

param.go
```diff
@@ -98,9 +98,36 @@ type paramVisitor interface {
...
+//		- param is not an edge node (i.e. param is a paramList), OR
+//		- param is an edge node that has not yet been visited
+func (pv paramVisitOnce) Visit(p param) paramVisitor {
+	switch p.(type) {
+	case paramSingle, paramGroupedSlice:
+		if _, ok := pv.visited[p.String()]; ok {
+			return pv
+		}
+		defer func() {
+			pv.visited[p.String()] = struct{}{}
```
abhinav: Why the defer? Wouldn't marking that param as visited right here before you
call `f` work?

param.go
```diff
@@ -98,9 +98,36 @@ type paramVisitor interface {
 // recursed into.
 type paramVisitorFunc func(param) (recurse bool)
 
-func (f paramVisitorFunc) Visit(p param) paramVisitor {
-	if f(p) {
-		return f
+// paramVisitOnce encapsulates a paramVisitorFunc and a map to track visited nodes.
+type paramVisitOnce struct {
+	f       paramVisitorFunc
+	visited map[string]struct{}
```
abhinav: It would be safer to use the `key` representation here instead of `String()`.
I don't like the idea of making the human-readable representation part of core
business logic.

```
map[key]struct{}
```

And,

```
case paramSingle:
  k = key{name: p.Name, t: p.Type}
case paramGroup:
  k = key{group: p.Group, t: p.Type.Elem()}  // []Foo -> Foo
```

param.go
```diff
@@ -98,9 +98,36 @@ type paramVisitor interface {
...
+}
+
+// NewParamVisitOnce constructs a paramVisitor avoiding redundant visits.
+func newParamVisitOnce(visited map[string]struct{}, f paramVisitorFunc) paramVisitor {
```
abhinav: NBD since this is all internal but strictly speaking, newParamVisitOnce
doesn't need to operate on paramVisitorFunc. It can be on any paramVisitor,
and it can track the list of visited nodes internally.

```
type paramVisitorOnce struct {
    visited map[key]struct{}
    visitor paramVisitor
}

func newParamVisitorOnce(paramVisitor) paramVisitor

func (v paramVisitorOnce) Visit(p param) paramVisitor {
    // If visited, return nil
    // else,
    
    if newV := v.visitor.Visit(p); newV != nil {
        v.visitor = newV
        return v
    }
    
    return nil
}
```

This would mean that detectCycles has to use a regular `paramVisitor` instead
of paramVisitorFunc. Or we can change `paramVisitorFunc` to the following and
it'll work with newParamVisitorOnce seamlessly.

    type paramVisitorFunc func(param) paramVisitor

---

That said, this is largely optional since this is all super internal.

param.go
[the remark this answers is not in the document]
zmt: I am convinced of the struct wrapper approach. I also tinkered with the defer and convinced myself that is the right choice, too.

param.go
dnathe4th: I actually went down this route originally! However, because `detectCycles` calls `paramVisitorFunc`/`newParamVisitOnce` itself we would lose some of our efficiency due to a new `visited` map getting created each time it recurses. My solution therefore evolved to have the map get passed in. That said, maybe an equally valid solution is to construct the `paramVisitOnce` a single time and pass *that* into `detectCycles`. It would still require the extra parameter to `detectCycles` like this PR adds, but it might clarify some of the internal semantics. Do you think that is preferable?

param.go
dnathe4th: This is a good idea. I didn't love `.String()`, but I was unable to use `p` directly as I don't believe it's hashable. Let me construct a `key`. Actually, give how many times `key`s are constructed now, does it make sense to make that a method on the various `param` types?

param.go
abhinav: I may be misunderstanding this but under the snippet above, it wouldn't be a
new `visited` map.

In the following code,

```
if newV := v.visitor.Visit(p); newV != nil {
  v.visitor = newV
  return v
}
```

We create a new paramVisitor out of the paramVisitorOnce (that's one alloc to
convert the value into an interface) but re-uses the existing `visited` map.

param.go
dnathe4th: Hm maybe you are right, not sure why I thought it was important to only do it after `f`. Might be a vestige from before I was only returning early for edge parameters.

param.go
abhinav: It does, but I would skip it. I had been working on removing the `key` type
completely when this code was being refactored. The semantics with `key` get
weird because only leaf nodes have `key`s.

cycle.go
dnathe4th: I will add that @abhinav . @zmt do you mean the first addition to the graph overall? `path[0]` here will only ever be a **result** from the function `Provide`d. Not sure if that answers your question.

param.go
```diff
@@ -105,6 +115,50 @@ func (f paramVisitorFunc) Visit(p param) paramVisitor {
...
+
+// NewParamVisitOnce constructs a paramVisitor avoiding redundant visits.
+func newParamVisitOnce() paramVisitorSetter {
+	return paramVisitOnce{
+		visited: make(map[key]struct{}),
+	}
+}
+
+func (pv paramVisitOnce) With(f paramVisitorFunc) paramVisitor {
+	return paramVisitOnce{
```
abhinav: Why make this a two-part constructor instead of
`newParamVisitOnce(paramVisitor) paramVisitor`?

param.go
```diff
@@ -113,22 +113,28 @@ func (f paramVisitorFunc) Visit(p param) paramVisitor {
...
 //
 // This is very similar to how go/ast.Walk works.
-func walkParam(p param, v paramVisitor) {
+//
+// paramVisitor can end the walk early by returning nil.
+func walkParam(p param, v paramVisitor) (cont bool) {
 	v = v.Visit(p)
 	if v == nil {
-		return
+		return false
```
abhinav: Actually, this doesn't seem right. The idea behind `v.Visit` returning nil is
that it doesn't want to recurse further into that node. So if your visitor
doesn't care about *this* paramObject, it can return nil to indicate that we
shouldn't recurse into it, but we should still check the other objects. This
will end the whole iteration early.

If we want visitors to be able to abort the iteration, we can theoretically
make them return an `error` in addition to the "next" visitor.

PR #209, opened 2018-09-13
Add SkipAcyclicVerification Option, and container.VerifyAcyclic

dnathe4th:
It is not unreasonable for libraries using `dig` under the hood for dependency graph management to `c.Provide` constructors in a tight loop. Rather than paying the cost of of verifying the graph is acyclic for every `Provide`, this option and method combination can be used to verify the graph once, after the loop has completed.


dig.go
```diff
@@ -108,6 +109,22 @@ func Name(name string) ProvideOption {
...
 
+// SkipAcyclicVerification is a ProvideOption to override the default behavior
+// of container.Provide to ensure the dependency graph remains acyclic for each
+// new provider.
+//
+// Applications adding providers to a container in a tight loop may experience
+// performance improvements by using this option.
+//
+// NOTE: A dig container has undefined behavior if the dependency graph is not
+// acyclic. It is highly recommended to call container.VerifyAcyclic before
```
bayesianmind: Any downsides to programatically enforcing this? Something like a hidden verified flag on the container. Then when an invoke is run, if the flag is false we verify the container first.

dig.go
```diff
@@ -211,6 +217,20 @@ func New(opts ...Option) *Container {
...
 
+// SkipAcyclicVerification is an Option to override the default behavior
+// of container.Provide, disabling the validation the dependency graph remains
+// acyclic for each new provider.
+//
+// Applications adding providers to a container in a tight loop may experience
+// performance improvements by initializing the container with this option.
+//
+// NOTE: The container will verify the graph on first `Invoke`.
+func SkipAcyclicVerification() Option {
```
bayesianmind: Consider naming DeferAcyclicVerification instead

dig.go
```diff
@@ -211,6 +217,20 @@ func New(opts ...Option) *Container {
 	return c
 }
 
+// DeferAcyclicVerification is an Option to override the default behavior
+// of container.Provide, disabling the validation the dependency graph remains
```
bayesianmind: nit: this comment needs updating. Otherwise looks good to me. I didn't thoroughly verify the correctness of your code, I'll leave that to the folks who actually own dig and know the internals better.

dnathe4th: I actually did rewrite this a bit. Is your objection to the word "disabling"? I think my sentence is confusing. This one combined with the last sentence in the comment block is intended to indicate the container will no longer validate on each Provide, but it will before the first Invoke.

dig_test.go
```diff
@@ -1598,6 +1598,29 @@ func TestProvideCycleFails(t *testing.T) {
...
+
+		c := New(DeferAcyclicVerification())
+		assert.NoError(t, c.Provide(newA))
+		assert.NoError(t, c.Provide(newB))
+		assert.NoError(t, c.Provide(newC))
+
+		err := c.Invoke(func(*A) {})
+		require.Error(t, err, "expected error when introducing cycle")
+		assertErrorMatches(t, err,
+			`this function introduces a cycle:`,
```
glibsm: I think my concern with this approach is the error messaging. Since this test uses a regex, I'm not sure what the actual content of the message is.

But in essence, the reason why cycle verification is done on each node insertion is so you can pinpoint which element has caused the cycle. Are we still able to report the errors with the same accuracy? I think if the verification is deferred and the graph is complete (which could actually contain several cycles), are we able to properly point to the function (file, line number) that caused it?

dnathe4th: I have updated for clarity

dig.go
```diff
@@ -131,6 +131,12 @@ type Container struct {
 
 	// Source of randomness.
 	rand *rand.Rand
+
+	// Ensure verified acyclic before Invoke
+	verifiedAcyclic bool
+
+	// Defer acyclic check on provide until Invoke
```
abhinav: nit: Docs should be full sentences ending in a period.

dig.go
```diff
@@ -131,6 +131,12 @@ type Container struct {
 
 	// Source of randomness.
 	rand *rand.Rand
+
+	// Ensure verified acyclic before Invoke
+	verifiedAcyclic bool
```
abhinav: nit: Can you rename this to `isAcyclic` or `isVerifiedAcyclic`?  I misread
this as `verifyCyclic`, which made it confusing next to the
`deferAcyclicVerification` option.

dig.go
```diff
@@ -498,6 +524,31 @@ func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
 	return nil
 }
 
+func (c *Container) verifyAcyclic() error {
+	// invert the c.providers map to satisfy verifyAcyclic's arguments
+	nodesToKeys := make(map[dot.CtorID][]key)
```
abhinav: If possible, I would like to avoid making `dot.CtorID` part of the business
logic like this. It exists to identify constructors for the GraphViz
visualization only.

If you change `detectCycles` to handle the `len(path) == 0` case (which it
doesn't with #208 because of the `path[0]`), you should be able to use that
directly here.

    visited := make(map[key]struct{})
    for _, n := range c.nodes {
        if err := detectCycles(n, c, nil /* path */, visited); err != nil {
            // ...
        }
    }

dnathe4th: This makes sense. I am going to wait for #208 to land to update this PR appropriately.

dig.go
```diff
@@ -498,6 +523,18 @@ func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
 	return nil
 }
 
+func (c *Container) verifyAcyclic() error {
+	visited := make(map[key]struct{})
+	for _, n := range c.nodes {
+		if err := detectCycles(n, c, nil /* path */, visited); err != nil {
+			return errWrapf(err, "this function introduces a cycle")
```
abhinav: The original error message is for a Provide, referencing the function that was
provided. In this case, the error can probably be "found a cycle" or something
similarly generic.

cycle.go
```diff
@@ -55,6 +55,13 @@ func (e errCycleDetected) Error() string {
 	return b.String()
 }
 
+// IsCycleDetected returns a boolean as to whether the provided error indicates
+// a cycle was detected in the container graph.
+func IsCycleDetected(err error) bool {
+	_, ok := err.(errCycleDetected)
```
abhinav: You should probably be looking at the RootCause of the error. This won't return true for errors returned by invoke right now (because of the errWrapf).

cycle_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsCycleDetected(t *testing.T) {
+	assert.False(t, IsCycleDetected(errors.New("new error")))
+	assert.True(t, IsCycleDetected(errCycleDetected{}))
```
abhinav: might be safer to test this against errors returned by Invoke/Provide.

dig.go
```diff
@@ -131,6 +131,12 @@ type Container struct {
 
 	// Source of randomness.
 	rand *rand.Rand
+
+	// Ensure verified acyclic before Invoke.
```
abhinav: comment outdated?

dig.go
```diff
@@ -498,6 +523,18 @@ func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
...
 
+func (c *Container) verifyAcyclic() error {
+	visited := make(map[key]struct{})
+	for _, n := range c.nodes {
+		if err := detectCycles(n, c, nil /* path */, visited); err != nil {
+			return errWrapf(err, "this function introduces a cycle")
+		}
+	}
+
+	c.isVerifiedAcyclic = true
```
abhinav: nit: I would set this in the `if` clause above where you're checking it so
that it reads like this:

```
if !isVerified {
  verify()
  isVerified = true
}
```

dig.go
```diff
@@ -515,12 +552,18 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
...
+		c.isVerifiedAcyclic = false
+		if c.deferAcyclicVerification {
+			c.providers[k] = append(c.providers[k], n)
+			continue
+		}
+		oldProviders := c.providers[k]
+		c.providers[k] = append(oldProviders, n)
 		if err := verifyAcyclic(c, n, k); err != nil {
-			c.providers[k] = oldProducers
+			c.providers[k] = oldProviders
```
abhinav: nit: I think we can avoid appending to providers twice since we're going to do
the first append unconditionally

```
for k := range keys {
    oldProviders := c.providers[k]
    c.providiers[k] = append(oldProviders, n)

    if c.deferAcyclicVerification {
        c.isVerifiedAcyclic = false
        continue
    }

    if err := verifyAcylic(c, n, k); err != nil {
        c.providers[k] = oldProviders
        return err
    }

    c.isVerifiedAcyclic = true
}
```

dig_test.go
```diff
@@ -1598,6 +1598,41 @@ func TestProvideCycleFails(t *testing.T) {
...
+		newD := func(*C) *D { return &D{} }
+
+		c := New(DeferAcyclicVerification())
+		assert.NoError(t, c.Provide(newA))
+		assert.NoError(t, c.Provide(newB))
+		assert.NoError(t, c.Provide(newC))
+		assert.NoError(t, c.Provide(newD))
+
+		err := c.Invoke(func(*A) {})
+		require.Error(t, err, "expected error when introducing cycle")
```
abhinav: This should also have `assert.True(t, IsCycleDetected(err))`

dig.go
dnathe4th: I'm not sure I agree. The `if` clause above is only testing starting from a single node. We don't really know that the container is verifiably acyclic until after all nodes have been tested.

abhinav: Oh, my comment may have been confusing. I was talking about the one in Invoke.

```
if !c.isVerifiedAcyclic {
    if err := c.verifyAcyclic(); err != nil {
        return err
    }
}
```

Right now, we check `isVerifiedAcyclic` in Invoke, then verifyAcylic sets it.
I'm suggesting,

```
if !c.isVerifiedAcyclic {
    if err := c.verifyAcyclic(); err != nil {
        return err
    }
    c.isVerifiedAcyclic = true
}
```

PR #215, opened 2018-10-03
WIP provide params to invoke 

rubensayshi:
Would it be an idea to let `Invoke` take arguments that aren't coming from the container?

It seems rather straight forward to add this, which leads me to belief it's not in there for a (good) reason...?

I understand this PR would break BC and it would be better to have a new method, but opening the pR atm purely for discussion as it's also missing any tests atm so it's not ready to be merged anyway.  
If there's interest in merging it then I'll add `InvokeWithParam` or smt and add tests.

would be used something like;

```go

// my super complicated service
type MyService struct {
	Booyaa bool
}

// provide for my super complicated service
func NewMyService() *MyService {
	return &MyService{}
}

func CanWeBooyaa(ctx *gin.Context, service *MyService) {
	ctx.String(200, fmt.Sprintf("booyaa? %v! \n", service.Booyaa))
}

container := dig.New()
container.Provide(NewMyService)

ctx := &gin.Context{}
container.Invoke(CanWeBooyaa, ctx)

```

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -476,7 +476,7 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 //
 // The function may return an error to indicate failure. The error will be
 // returned to the caller as-is.
-func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
+func (c *Container) Invoke(function interface{}, providedParams ...interface{}) error {
 	ftype := reflect.TypeOf(function)
 	if ftype == nil {
 		return errors.New("can't invoke an untyped nil")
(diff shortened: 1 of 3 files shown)


rubensayshi:
```
BAD FOR: Exposing to user-land code as a Service Locator.
```

I noticed this in the README, though I don't really understand why dig would be bad for this nor is there anything I can find in the go space that would be better for it.

rubensayshi:
you're saying significantly slower, but benchmarking dig shows its ~1500ns for a provide, that's not that bad ...

glibsm:
> you're saying significantly slower, but benchmarking dig shows its ~1500ns for a provide, that's not that bad ...

`Provide` is not the only thing that needs to be benched, it's also the `Invoke`'s. Dig is implemented using reflection and the general rule of thumb is "no reflection in the hot path".

PR #218, opened 2018-10-26
Add the ability to prune non-error entries out of the graph

rhang-uber:


internal/dot/graph.go
```diff
@@ -252,6 +270,50 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
 	group.Results = append(group.Results, r)
 }
 
+// Prune removes elements from the graph that do not have failed results.
+// Removing elements that do not have failing results makes the graph easier to debug, since non-failing
+// nodes and edges can clutter the graph and don't help the user debug.
+func (dg *Graph) Prune() {
+	failedResults := make(map[*Result]bool)
```
abhinav: If possible, I would avoid using `*Result` as map keys. Using pointers as map
keys can lead to confusing/unexpected behavior.

For example, https://play.golang.org/p/x69eTtwuMLz. (This is probably the
behavior we want here but it's easier to mess it up by using it as a map key.)

You can define a groupKey-like struct instead.

internal/dot/graph.go
```diff
@@ -252,6 +270,50 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+	for _, v := range dg.Failed.RootCauses {
+		failedResults[v] = true
+	}
+	for _, v := range dg.Failed.TransitiveFailures {
+		failedResults[v] = true
```
abhinav: nit: For sets, we generally use `struct{}` instead of `bool`. It doesn't
matter much here but `struct{}` is zero-cost in terms of space, IIRC.

```
make(map[foo]struct[})
```

internal/dot/graph.go
```diff
@@ -252,6 +270,50 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+	}
+
+	dg.pruneCtors(failedResults)
+	dg.pruneGroups(failedResults)
```
abhinav: I think if you do the book keeping in advance, this gets simplified a bit.
You can probably just maintain a list of failed groupKeys and CtorIDs on the
FailedNodes object as FailNodes and FailGroupNodes methods are called. Then in
Prune, it's just a matter of iterating through those lists once rather than
traversing down to look for failed Results.

internal/dot/graph.go
rhang-uber: Resolved.

internal/dot/graph.go
```diff
@@ -105,6 +109,12 @@ type FailedNodes struct {
 	// TransitiveFailures is the list of nodes that failed to build due to
 	// missing/failed dependencies.
 	TransitiveFailures []*Result
+
+	//FailedCtors stores a map of failed Ctor Ids  that is populated as the graph is traversed for errors.
+	FailedCtors map[CtorID]struct{}
+
+	// FailedGroups stores a map of failed groupKeys that is populated as the graph is traversed for errors.
+	FailedGroups map[groupKey]struct{}
```
abhinav: Do both these fields need to be private. Red flag here is that groupKey is private so FailedGroups can never be used outside this package anyway.

internal/dot/graph.go
```diff
@@ -190,6 +203,7 @@ func (dg *Graph) AddMissingNodes(results []*Result) {
 func (dg *Graph) FailNodes(results []*Result, id CtorID) {
 	// This failure is the root cause if there are no other failures.
 	isRootCause := len(dg.Failed.RootCauses) == 0
+	dg.Failed.FailedCtors[id] = struct{}{}
```
abhinav: notice the stuttering here: `dg.Failed.FailedCtors`. We're already inside FailedNodes so "failed" is already in context. The names of the fields inside FailedNodes don't need to include the word "failed".

internal/dot/graph.go
```diff
@@ -252,6 +274,59 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
 	group.Results = append(group.Results, r)
 }
 
+// Prune removes elements from the graph that do not have failed results.
+// Removing elements that do not have failing results makes the graph easier to debug, since non-failing
+// nodes and edges can clutter the graph and don't help the user debug.
+func (dg *Graph) Prune() {
```
abhinav: let's rename this to something that makes the behavior slightly more obvious without reading the docs. Something like `PruneSuccess` or `FilterFailed`? (There are likely better names than that too.)

internal/dot/graph.go
```diff
@@ -252,6 +274,59 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+	dg.pruneCtors()
+	dg.pruneGroups()
+}
+
+// Remove constructors from the graph that do not have failing Results.
+func (dg *Graph) pruneCtors() {
+	var prunedCtors []*Ctor
```
abhinav: nit: given the number of variables in context, you can use shorter variable
names here. Go prefers short variable names.

https://github.com/golang/go/wiki/CodeReviewComments#variable-names

internal/dot/graph.go
```diff
@@ -252,6 +274,59 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+	}
+
+	dg.Ctors = prunedCtors
+}
+
+// Remove groups from the graph that do not have failing Results.
+func (dg *Graph) pruneGroups() {
+	var prunedGroups []*Group
+	for _, g := range dg.Groups {
+		if _, ok := dg.Failed.FailedGroups[groupKey{t: g.Type, group: g.Name}]; ok {
```
abhinav: Maybe groups should have a `.groupKey()` helper too?

internal/dot/graph.go
```diff
@@ -252,6 +274,59 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+			prunedGroups = append(prunedGroups, g)
+		} else {
+			delete(dg.groupMap, groupKey{t: g.Type, group: g.Name})
```
abhinav: groupKey constructed twice here. Should use variable.

internal/dot/graph.go
```diff
@@ -252,6 +274,59 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+
+	dg.Groups = prunedGroups
+}
+
+// PruneGroupResults removes constructor results that are still referenced in
+// their respective Group object.
+func (dg *Graph) pruneGroupResults(c *Ctor) {
+	for _, r := range c.Results {
+		var prunedResults []*Result
+		if resultGroup, ok := dg.groupMap[r.groupKey()]; ok {
```
abhinav: Result.groupKey will return `groupKey{Name: "", Type: ..}` when this isn't a group-result. You should probably check if `Group != ""` first.

abhinav: Also, code-style-wise, reduce nesting here. 

```
rg, ok := ...
if !ok {
    continue
}

for _, gr := rg.Results {
    ...
}

rg.Results = prunedResults
```

That can be simplified further by exploiting the fact that `foo[k]` returns
zero-value in the single-value form. So you can just do,

```
rg := dg.groupMap[..]
if rg == nil { continue }
```

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

internal/dot/graph.go
```diff
@@ -252,6 +274,59 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+			for _, gr := range resultGroup.Results {
+				if r.GroupIndex != gr.GroupIndex {
+					prunedResults = append(prunedResults, gr)
+				}
```
abhinav: So previously, GroupIndex existed solely to help GraphViz. With this change,
it transitions into an almost-unique identifier for the Result.

In a follow-up PR, consider refactoring this a bit so that we have a ResultID
type (which is probably just an integer).

```
type ResultID int
```

The uniqueness of the ResultID can be scoped to the Graph so you would track
the "next" group ID on the Graph object rather than based on the length of the
Group.Results.

```
type Graph struct {
    ...
    
    resultIDCounter int
}

func (g *Graph) nextResultID() int {
    c := g.resultIDCounter
    g.resultIDCounter++
    return c
}
```

This will let you change the Group itself to hold a `map[ResultID]*Result`,
which makes this particular deletion a lot more efficient.

---

To clarify, this is fine as-is for now. The refactoring should be done in a
future change. For now, you can just add a TODO to do this more efficiently.

```
// TODO: Avoid linear search
```

internal/dot/graph_test.go
```diff
@@ -36,7 +36,10 @@ func TestNewGraph(t *testing.T) {
 
 	assert.Equal(t, make(map[groupKey]*Group), g.groupMap)
 	assert.Equal(t, make(map[CtorID]*Ctor), g.ctorMap)
-	assert.Equal(t, &FailedNodes{}, g.Failed)
+	assert.Equal(t, &FailedNodes{
+		FailedCtors:  make(map[CtorID]struct{}),
+		FailedGroups: make(map[groupKey]struct{}),
+	}, g.Failed)
```
abhinav: not necessary. internal state as far as this code is concerned.

abhinav: in fact, this whole test is unnecessary. this is equivalent to writing the same code again except from the other side.

internal/dot/graph.go
rhang-uber: Resolved

internal/dot/graph.go
rhang-uber: Resolved

internal/dot/graph.go
rhang-uber: I believe in this comment you are implying that both fields should be private? I believe it should be an am changing that now.

internal/dot/graph_test.go
rhang-uber: It make sense that testing constructors are not as valuable. Removed.

internal/dot/graph.go
rhang-uber: I changed the code to integrate continue statements in. I don't believe there is much of a concision improvement of 

rg := dg.groupMap[...]
over
		rg, ok := dg.groupMap[k]
		if !ok {
			continue
		}

internal/dot/graph.go
```diff
@@ -47,24 +47,39 @@ type Ctor struct {
...
 	Results     []*Result
+	ResultCtor  map[nodeKey][]*Ctor
 	ErrorType   ErrorType
 }
 
+type nodeKey struct {
+	t     reflect.Type
+	name  string
+	group string
+}
```
abhinav: nodeKey is a superset of groupKey. You can drop groupKey if you're using nodeKey.

internal/dot/graph.go
```diff
@@ -252,6 +289,116 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+		}
+	}
+
+	dg.Ctors = pruned
+}
+
+// indexParams builds a map of params to constructors that consume those params so that constructor results
+// can be associated with constructors that consume those results.
+func (dg *Graph) indexParams() {
+	pIndex := make(map[nodeKey][]*Ctor)
```
abhinav: Don't need to build this at the end. You can build this as constructors are added. In fact, this is replicating what the Container does:

https://github.com/uber-go/dig/blob/007ab720a796b1027b6c289fb4e836c8fc077357/dig.go#L121

So you can definitely track this as you go.

internal/dot/graph.go
rhang-uber: nodeKey is used to for comparison purposes, groupKey is used to quickly generate at a key to lookup the group object in dg.groupMap and still has a purpose outside of nodeKey. If i removed groupKey I would have to manually construct the groupKey more than once.

internal/dot/graph.go
rhang-uber: Makes sense! Done.

internal/dot/graph.go
rhang-uber: I think PruneSuccess is a great name!

internal/dot/graph.go
rhang-uber: I changed all variables of a similar purpose to pruned.

internal/dot/graph.go
abhinav: Discussed offline: It'll still work with just nodeKey. We can delete groupKey completely.

internal/dot/graph.go
```diff
@@ -50,21 +50,45 @@ type Ctor struct {
...
 // Node is a single node in a graph and is embedded into Params and Results.
 type Node struct {
 	Type  reflect.Type
 	Name  string
 	Group string
 }
 
-// Param is a parameter node in the graph.
+func (n *Node) nodeKey() nodeKey {
+	return nodeKey{n.Type, n.Name, n.Group}
```
abhinav: nit: prefer keyed struct literals

    nodeKey{t: .., name: .., group: ..}

Some linters will complain if you don't.

internal/dot/graph.go
```diff
@@ -92,6 +131,8 @@ type Graph struct {
 	Groups   []*Group
 	groupMap map[groupKey]*Group
 
+	paramCtorMap map[nodeKey][]*Ctor
```
abhinav: nit: Can call this "consumers" to replicate what dig's Container does.

I know this is already done in the existing code, but we should avoid naming
things after their type since they're strongly typed and we already know the
type.

You can name it after its purpose too: It's a reverse index to find
constructors given the parameter type, so you can name it "ctorParamIndex" as
well. "consumers" is equally informative because it says you're looking for
things that consume values of this type.

internal/dot/graph.go
```diff
@@ -105,6 +146,12 @@ type FailedNodes struct {
 	// TransitiveFailures is the list of nodes that failed to build due to
 	// missing/failed dependencies.
 	TransitiveFailures []*Result
+
+	// ctors is a collection of failed constructors IDs that are populated as the graph is traversed for errors.
```
abhinav: nit: let's try to wrap comments at 80 characters. There are no hard rules in Go about line length but for comments, we should aim for short and readable.

internal/dot/graph.go
```diff
@@ -252,6 +318,87 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+// Remove constructors from the graph that do not have failing Results.
+func (dg *Graph) pruneCtors() {
+	var pruned []*Ctor
+	for _, c := range dg.Ctors {
+		if _, ok := dg.Failed.ctors[c.ID]; ok {
+			pruned = append(pruned, c)
+		} else {
+			dg.pruneCtorParams(c)
+			dg.pruneGroupResults(c)
+			delete(dg.ctorMap, c.ID)
```
abhinav: Let's add a comment to explain what's going on here.

```
If we deleted the constructor, we need to delete the stale references to the
constructor's Results from Groups and other Ctors' params.
```

internal/dot/graph.go
```diff
@@ -252,6 +318,87 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+	dg.pruneGroups()
+}
+
+// Remove constructors from the graph that do not have failing Results.
+func (dg *Graph) pruneCtors() {
+	var pruned []*Ctor
+	for _, c := range dg.Ctors {
+		if _, ok := dg.Failed.ctors[c.ID]; ok {
+			pruned = append(pruned, c)
+		} else {
```
abhinav: readability nit: reduce nesting

```
if _, ok := ...; ok {
    pruned = ...
    continue
}

// ...
```

internal/dot/graph.go
```diff
@@ -252,6 +318,87 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+}
+
+// pruneGroups removes groups from the graph that do not have failing results.
+func (dg *Graph) pruneGroups() {
+	var pruned []*Group
+	for _, g := range dg.Groups {
+		k := g.groupKey()
+		if _, ok := dg.Failed.groups[k]; ok {
+			pruned = append(pruned, g)
+		} else {
```
abhinav:
```suggestion
			continue
		}

```

internal/dot/graph.go
```diff
@@ -252,6 +318,87 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+			delete(dg.groupMap, k)
+		}
+	}
+	dg.Groups = pruned
+
+	dg.pruneCtorGroupParams()
+}
+
+// pruneCtorParams removes constructor results that are still referenced in the Params of constructors that consume those results.
+// If the results in the constructor are found in the params of a consuming constructor that result should be removed.
```
abhinav: nit: comment wrapping

internal/dot/graph.go
```diff
@@ -252,6 +318,87 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+func (dg *Graph) pruneCtorParams(c *Ctor) {
+	for _, r := range c.Results {
+		if ctors, ok := dg.paramCtorMap[r.nodeKey()]; ok {
+			for _, ctor := range ctors {
```
abhinav: Don't need `if ok` check here. accessing an item that doesn't exist gives you
a zero-value, and a zero value slice is range-able. So you can do,

```
for _, r := range c.Results {
    for _, ctor := range dg.consumers[r.nodeKey()] {
        ...
    }
}
```

internal/dot/graph.go
```diff
@@ -252,6 +318,87 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+}
+
+// pruneCtorGroupParams removes constructor results that are still referenced in the GroupParams of constructors that consume those results.
+func (dg *Graph) pruneCtorGroupParams() {
+	for _, c := range dg.Ctors {
+		var pruned []*Group
+		for _, gp := range c.GroupParams {
+			k := gp.groupKey()
+			if _, ok := dg.groupMap[k]; ok {
+				pruned = append(pruned, gp)
```
abhinav: You're using modifications made to groupMap elsewhere to communicate what
should be deleted here. That works but it's a bit fragile. Information is
being communicated here implicitly through side-effects. For example, if
someone changes the ordering of which prune operation is called when, this
will break and it'll be pretty difficult to debug.

I suggest passing in the list/set of groups to delete/keep in as a parameter.

```
dg.pruneCtorGroupParams(deleted []*Group)
```

internal/dot/graph.go
```diff
@@ -252,6 +318,87 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
 	group.Results = append(group.Results, r)
 }
 
+// Prune removes elements from the graph that do not have failed results.
+// Removing elements that do not have failing results makes the graph easier to debug, since non-failing
+// nodes and edges can clutter the graph and don't help the user debug.
+func (dg *Graph) PruneSuccess() {
+	dg.pruneCtors()
+	dg.pruneGroups()
```
abhinav: as per offline discussion: let's pass in `dg.Failed.ctors` and `dg.Failed.groups` as parameters here rather than reaching into `dg` constantly.

internal/dot/graph.go
```diff
@@ -252,6 +318,87 @@ func (dg *Graph) addToGroup(r *Result, id CtorID) {
...
+			if _, ok := dg.groupMap[k]; ok {
+				pruned = append(pruned, gp)
+			}
+		}
+		c.GroupParams = pruned
+	}
+}
+
+// pruneGroupResults removes constructor results that are still referenced in the Group object that contains that result.
+func (dg *Graph) pruneGroupResults(c *Ctor) {
```
abhinav: we should probably document which constructor `c` is here and in the other function.

internal/dot/graph_test.go
```diff
@@ -132,17 +123,43 @@ func TestAddCtor(t *testing.T) {
...
+	t.Run("test prune removes all graph entries without failing results", func(t *testing.T) {
+		dg := NewGraph()
+		c0 := &Ctor{ID: 1234}
+		c1 := &Ctor{ID: 5678}
+		node0 := &Node{Type: type3, Group: "foo"}
+		node1 := &Node{Type: type3, Group: "foo"}
+		res0 := &Result{Node: node0, GroupIndex: 0}
+		res1 := &Result{Node: node1, GroupIndex: 1}
+
+		assert.Equal(t, map[groupKey]*Group{}, dg.groupMap)
```
abhinav:
```suggestion
```

(We don't need to test internal state changes.)

internal/dot/graph_test.go
```diff
@@ -132,17 +123,43 @@ func TestAddCtor(t *testing.T) {
...
+		dg.AddCtor(c0, []*Param{}, []*Result{res0})
+		dg.AddCtor(c1, []*Param{}, []*Result{res1})
+
+		dg.PruneSuccess()
+
+		assert.Equal(t, 0, len(dg.Ctors))
+		assert.Equal(t, 0, len(dg.ctorMap))
```
abhinav:
```suggestion
```

internal/dot/graph_test.go
```diff
@@ -132,17 +123,43 @@ func TestAddCtor(t *testing.T) {
...
+		assert.Equal(t, 0, len(dg.Groups))
+		assert.Equal(t, 0, len(dg.groupMap))
```
abhinav:
```suggestion
```

internal/dot/graph_test.go
```diff
@@ -181,6 +198,41 @@ func TestFailNodes(t *testing.T) {
...
+	t.Run("no constructors should be removed because they all have failing results", func(t *testing.T) {
+		dg := NewGraph()
+		c0 := &Ctor{ID: 123}
+		c1 := &Ctor{ID: 456}
+
+		dg.AddCtor(c0, []*Param{}, []*Result{r1})
+		dg.AddCtor(c1, []*Param{}, []*Result{r2})
+
+		dg.FailNodes([]*Result{r1}, 123)
+		dg.FailNodes([]*Result{r2}, 456)
```
abhinav: nit: to prevent failure from typos here, you can do,

    dg.FailNodes(.., c0.ID)

internal/dot/graph_test.go
```diff
@@ -181,6 +198,41 @@ func TestFailNodes(t *testing.T) {
...
+		dg.PruneSuccess()
+
+		assert.Equal(t, 2, len(dg.Ctors))
+		assert.Equal(t, 2, len(dg.ctorMap))
```
abhinav: nit: can use assert.Len instead

```suggestion
		assert.Len(t, dg.ctorMap, 2)
```

internal/dot/graph_test.go
```diff
@@ -181,6 +198,41 @@ func TestFailNodes(t *testing.T) {
...
+		c1 := &Ctor{ID: 456}
+
+		dg.AddCtor(c0, []*Param{}, []*Result{r1})
+		dg.AddCtor(c1, []*Param{}, []*Result{r2})
+
+		dg.FailNodes([]*Result{r1}, 123)
+		dg.FailNodes([]*Result{r2}, 456)
+		dg.PruneSuccess()
+
+		assert.Equal(t, 2, len(dg.Ctors))
```
abhinav:
```suggestion
		assert.Len(t, dg.Ctors, 2)
```

internal/dot/graph_test.go
```diff
@@ -181,6 +198,41 @@ func TestFailNodes(t *testing.T) {
...
+
+		dg.FailNodes([]*Result{r2}, 456)
+		dg.PruneSuccess()
+
+		assert.Equal(t, 1, len(dg.Ctors))
+		assert.Equal(t, 1, len(dg.ctorMap))
+		_, ok := dg.ctorMap[c0.ID]
+		assert.False(t, ok)
+		_, ok = dg.ctorMap[c1.ID]
+		assert.True(t, ok)
```
abhinav: don't need to test internal state here

internal/dot/graph_test.go
```diff
@@ -132,17 +123,43 @@ func TestAddCtor(t *testing.T) {
 
 		assert.Equal(t, []*Ctor{c0, c1}, dg.Ctors)
 	})
+
+	t.Run("test prune removes all graph entries without failing results", func(t *testing.T) {
+		dg := NewGraph()
+		c0 := &Ctor{ID: 1234}
+		c1 := &Ctor{ID: 5678}
```
abhinav: All these tests should probably be in a TestPrune test case.

internal/dot/graph.go
rhang-uber: I removed it and casted nodeKey into groupKey because it was required to lookup a node's equivalent value in the groupMap.

internal/dot/graph.go
rhang-uber: I don't think I can just add continue since I still need to delete the entry from the groupMap

internal/dot/graph.go
rhang-uber: I don't believe I can just continue because I still need to remove the entry from the groupMap.

internal/dot/graph.go
rhang-uber: I really like this!

internal/dot/graph.go
rhang-uber: I added documentation in the comments re: what the c argument is. Is there a better pattern for parameter documentation? I tried to look online, but there didn't seem to be an idiomatic way.

internal/dot/graph.go
rhang-uber: Offline discussion: I understand the comment, will change.

internal/dot/graph.go
rhang-uber: Offline discussion: I understand the comment, will change.

internal/dot/graph.go
rhang-uber: That make sense, I think consumer is a good name!

internal/dot/graph.go
rhang-uber: Will add these changes to the grafx: The next generation ticket!

internal/dot/README.md
```diff
@@ -0,0 +1,64 @@
...
+is consumed by node_a and that node_b is a parameter of node_a.
+
+**Nodes:**
+
+*Constructors* [Rectangles] - Takes parameters and produces results.
+
+*Results* [Ovals] - Results inside a constructor are produced by that constructor. Results are consumed
+directly by other constructors and/or part of a group of results.
+
+*Groups* [Diamonds] - Represent value groups in [fx](https://godoc.org/go.uber.org/fx). Multiple results can form a group. Any
```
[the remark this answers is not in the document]
rhang-uber: Will fix.

internal/dot/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+The dot module generates a DOT file representation of a dependency graph.
+
+## Interpreting the graph
+
+The graph should be read from left to right. The leftmost node in the graph (the root node) depends
+on its dependency tree to the right. An arrow from node_a to node_b in the graph means that node_b
+is consumed by node_a and that node_b is a parameter of node_a.
+
+**Nodes:**
```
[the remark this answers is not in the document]
rhang-uber: I actually like my current description because understanding the direction and flow of the graph was one of the more difficult tasks for me over the course of this project.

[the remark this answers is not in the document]
abhinav: Fair enough. I think I was unsure about this. (Nit: No need for ":" in headers.)

PR #220, opened 2018-11-01
Change travis go versions

glibsm:
Travis is config is quite out of date. Change to using a more recent set of go versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ go_import_path: go.uber.org/dig
 
 matrix:
   include:
-    - go: 1.7
-    - go: 1.8
     - go: 1.9
+    - go: 1.10
+    - go: 1.11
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,9 +4,9 @@ go_import_path: go.uber.org/dig
 
 matrix:
   include:
-    - go: 1.7
-    - go: 1.8
     - go: 1.9
```
abhinav:
```suggestion
```

.travis.yml
```diff
@@ -4,9 +4,9 @@ go_import_path: go.uber.org/dig
...
+    - go: 1.10
```
abhinav:
```suggestion
    - go: "1.10"
```

need quotes because "1.10" as float is "1.1"

PR #225, opened 2018-11-14
Expose "Group" Provide option

vyshah:
The commits are mostly individually reviewable.

This is a step towards a solution for [#610 (fx)](https://github.com/uber-go/fx/issues/610) for both named values and value groups. This will allow fx to expose its own Group option, which will allow users to form value groups without needing fx.Out.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -564,7 +564,7 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
 	n, err := newNode(
 		ctor,
 		nodeOptions{
-			ResultName: opts.Name,
+			ResultName:  opts.Name,
 			ResultGroup: opts.Group,
 		},
 	)
@@ -741,7 +741,7 @@ type node struct {
 type nodeOptions struct {
 	// If specified, all values produced by this node have the provided name
 	// or belong to the specified value group
-	ResultName string
+	ResultName  string
 	ResultGroup string
 }
 
@@ -758,7 +758,7 @@ func newNode(ctor interface{}, opts nodeOptions) (*node, error) {
 	results, err := newResultList(
 		ctype,
 		resultOptions{
-			Name: opts.ResultName,
+			Name:  opts.ResultName,
 			Group: opts.ResultGroup,
 		},
 	)
--- a/dig.go
+++ b/dig.go
@@ -118,7 +118,7 @@ func Name(name string) ProvideOption {
 }
 
 // Group is a ProvideOption that specifies that all values produced by a
-// constructor should be members of the same group. See also the package
+// constructor should be added to the specified group. See also the package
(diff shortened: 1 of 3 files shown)


dig.go
```diff
@@ -108,6 +117,20 @@ func Name(name string) ProvideOption {
 	})
 }
 
+// Group is a ProvideOption that specifies that all values produced by a
+// constructor should be members of the same group. Since all members of a
+// group must be of same type, this option cannot be provided for constructors
```
abhinav: This is incorrect. You can totally do,

```
type Result struct {
  fx.Out

  Reader io.Reader `group:"foo"`
  Writer io.Writer `group:"foo"`
}
```

Consumption,

```
type Param struct {
  fx.In

  Readers []io.Reader `group:"foo"`
  Writers []io.Writer `group:"foo"`
}
```

The primary-key is the triple `(type, value name, group name)` with the
additional constraint that only one of "value name" and "group name" is set.

vyshah:
@glibsm tests have been added :)

dig_test.go
```diff
@@ -919,6 +919,48 @@ func TestGroups(t *testing.T) {
...
+		provide(3)
+
+		type in struct {
+			In
+
+			Values []int `group:"val"`
+		}
+
+		require.NoError(t, c.Invoke(func(i in) {
+			assert.Equal(t, []int{2, 3, 1}, i.Values)
```
abhinav: can avoid need for setting the rand seed by using [`assert.ElementsMatch`](https://godoc.org/github.com/stretchr/testify/assert#ElementsMatch)

dig_test.go
```diff
@@ -919,6 +919,48 @@ func TestGroups(t *testing.T) {
...
+
+			Value int `group:"val"`
+		}
+
+		provide := func(i int) {
+			require.Error(t, c.Provide(func() out {
+				return out{Value: i}
+			}, Group("val")), "This Provide should fail")
+		}
+
```
abhinav: function unnecessary here since this is called exactly once. consider inlining it.

```
require.Error(t, c.Provide(...))
```

The contents of the function can also mark the test as failed to guarantee that it's never called.

```
func() out {
  t.Fatal("this function should never be called")
  return out{}
}
```

dig_test.go
```diff
@@ -919,6 +919,48 @@ func TestGroups(t *testing.T) {
 		}), "invoke failed")
 	})
 
+	t.Run("groups are provided via option", func(t *testing.T) {
```
abhinav: Can we add a test case for the same function producing multiple values to the
same group?

```
c.Provide(func() (int, int, string) { ... }, Group("foo"))
```

The consumer will expect a `[]int` and `[]string`, both with group "foo".

PR #233, opened 2019-03-07
Implementing a dig.As ProvideOption

### LINKED ISSUE #197: Allow to specify an interface when providing a constructor
Update (by @glibsm)

The addition of `dig.As` `ProvideOption` has been agreed on.

---

Currently `dig` only uses a constructors that returns exactly what the function expects.

If the function accepts an interface, and we don't inject the constructor that returns that interface we get this error

```go
panic: missing dependencies for function "main".main.func1 (..../main.go:29): type main.Fooer is not in the container, did you mean to use *main.Foo?
```

Take the following example
```go
type Foo struct{}

func NewFoo() *Foo {
	return &Foo{}
}

func (b *Foo) Hello() {
	fmt.Println("hello")
}

type Fooer interface {
	Hello()
}

func main() {
	di := dig.New()
	err := di.Provide(NewFoo)
	if err != nil {
		panic(err)
	}
	err = di.Invoke(func(f Fooer) {
		fmt.Printf("f: %+v", f)
	})
	if err != nil {
		panic(err)
	}
}
```

So to make it work it needs to have a constructor function such as:
```go
func NewFoo() Fooer {
	return &Foo{}
}
```

~which is basically making us declare that `Foo implements Fooer`~ 
~Would it be feasible to change the library so that it can workout itself if the struct implements the interface?~ _[edit] as clarified in the comments this suggestion won't work_

Or alternatively, add a new `ProvideOption` which allow us to specify which concrete implementation to inject for a given interface?

alessandrozucca:
This adds a dig.As option which allows the caller to specify a
constructor and a list of interfaces that the constructed struct
implements.

Closes #197

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -151,9 +151,6 @@ func Group(group string) ProvideOption {
 	})
 }
 
-// As is a ProvideOption that specifies that the struct produced by the constructor
-// implements the given interface
-
 // As is a ProvideOption that specifies that the value produced by the
 // constructor implements one or more other interfaces.  The value will be made
 // available as that interface in the container.
@@ -659,7 +656,7 @@ type node struct {
 
 type nodeOptions struct {
 	// If specified, all values produced by this node have the provided name
-	// belong to the specified value group or implement any of the interfaces
+	// belong to the specified value group or implement any of the interfaces.
 	ResultName  string
 	ResultGroup string
 	ResultAs    []interface{}
--- a/dig.go
+++ b/dig.go
@@ -152,15 +152,45 @@ func Group(group string) ProvideOption {
 }
 
 // As is a ProvideOption that specifies that the value produced by the
-// constructor implements one or more other interfaces.  The value will be made
-// available as that interface in the container.
+// constructor implements one or more other interfaces.
 //
-// As expects one or more pointers to the implemented interfaces.
+// As expects one or more pointers to the implemented interfaces. Values
+// produced by constructors will be made available in the container as
+// implementations of all of those interfaces.
 //
 // For example, the following will make the buffer available in the container
-// as an io.Reader.
+// as io.Reader and io.Writer.
 //
-//   c.Provide(newBuffer, dig.As(new(io.Reader)))
+//   c.Provide(newBuffer, dig.As(new(io.Reader), new(io.Writer)))
+//
+// That is, the above is equivalent to the following.
+//
+//   c.Provide(func(...) (*bytes.Buffer, io.Reader, io.Writer) {
+//     b := newBuffer(...)
+//     return b, b, b
+//   })
+//
+// If used with dig.Name, the type produced by the constructor and the types
+// specified with dig.As will all use the same name. For example,
+//
+//   c.Provide(newFile, dig.As(new(io.Reader)), dig.Name("temp"))
+//
+// The above is equivalent to the following.
+//
+//   type Result struct {
+//     dig.Out
+//
+//     File   *os.File  `name:"temp"`
+//     Reader io.Reader `name:"temp"`
+//   }
+//
+//   c.Provide(func(...) Result {
+//     f := newFile(...)
+//     return Result{
+//       File: f,
+//       Reader: f,
+//     }
+//   })
 //
 // This option cannot be provided for constructors which produce result
 // objects.
@@ -587,7 +617,6 @@ func (cv connectionVisitor) Visit(res result) resultVisitor {
 			return nil
 		}
 		cv.keyPaths[k] = path
-
 		for _, asType := range r.As {
 			k := key{name: r.Name, t: asType}
 			if err := cv.checkKey(k, path); err != nil {
@@ -603,7 +632,6 @@ func (cv connectionVisitor) Visit(res result) resultVisitor {
 		// value there.
 		k := key{group: r.Group, t: r.Type}
 		cv.keyPaths[k] = path
-
 	}
 
 	return cv
--- a/dig.go
+++ b/dig.go
@@ -66,9 +66,15 @@ type provideOptions struct {
 }
 
 func (o *provideOptions) Validate() error {
-	if len(o.Group) > 0 && len(o.Name) > 0 {
-		return fmt.Errorf(
-			"cannot use named values with value groups: name:%q provided with group:%q", o.Name, o.Group)
+	if len(o.Group) > 0 {
+		if len(o.Name) > 0 {
+			return fmt.Errorf(
+				"cannot use named values with value groups: name:%q provided with group:%q", o.Name, o.Group)
+		}
+		if len(o.As) > 0 {
+			return fmt.Errorf(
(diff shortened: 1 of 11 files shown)


alessandrozucca:
still need to add a test for the `DotResult()`..

dig.go
```diff
@@ -565,11 +593,61 @@ func (cv connectionVisitor) Visit(res result) resultVisitor {
...
+			if err := checkPathsForKey(cv.keyPaths, k, path); err != nil {
+				*cv.err = err
+				return nil
+			}
+			if err := checkProvidersForKey(cv.c.providers, k, path); err != nil {
+				*cv.err = err
+				return nil
+			}
+			cv.keyPaths[k] = path
+		}
```
amckinney: I think we can dry this up a little. These helpers closely resemble the helpers used in the `resultSingle` case. We can consolidate the helper functions into a method on the connectionVisitor:
```
func (cv connectionVisitor) checkKey(k key, path string) error {
	if err := checkPathsForKey(cv.keyPaths, k, path); err != nil {
		return err
	}
	if err := checkProvidersForKey(cv.c.providers, k, path); err != nil {
		return err
	}
	return nil
}
```
Since we are only calling these helpers from this function, we can flatten their implementations here, too. This would reduce the layers of indirection slightly.
```
func (cv connectionVisitor) checkKey(k key, path string) error {
	if conflict, ok := cv.keyPaths[k]; ok {
		return fmt.Errorf(
			"cannot provide %v from %v: already provided by %v",
			k, path, conflict)
	}
	if ps := cv.c.providers[k]; len(ps) > 0 {
		cons := make([]string, len(ps))
		for i, p := range ps {
			cons[i] = fmt.Sprint(p.Location())
		}

 		return fmt.Errorf(
			"cannot provide %v from %v: already provided by %v",
			k, path, strings.Join(cons, "; "))
	}
	return nil
}
```
This way, we can call this as:
```
	case resultAs:
		k := key{name: r.Name, t: r.Type}
		if err := cv.checkKey(k, path); err != nil {
			*cv.err = err
			return nil
		}
		cv.keyPaths[k] = path

 		for _, as := range r.As {
			asIndirectValue := reflect.Indirect(reflect.ValueOf(as))
			k := key{name: r.Name, t: asIndirectValue.Type()}

 			if err := cv.checKey(k, path); err != nil {
				*cv.err = err
				return nil
			}
			cv.keyPaths[k] = path
		}
	}

	return cv
```

Now we could use this function to `resultSingle`, too.

dig.go
```diff
@@ -127,6 +145,14 @@ func Group(group string) ProvideOption {
 	})
 }
 
+// As is a ProvideOption that specifies that the struct produced by the constructor
+// implements the given interface
+func As(i interface{}) ProvideOption {
```
amckinney: We can also update this to `As(...interface{})`, too. That would simplify some of your test cases below.

glibsm: It would simplify the test cases, but introduce two ways of doing the same thing:

`dig.Provide(myfunc, dig.As(foo, bar, baz)`
`dig.Provide(myfunc, dig.As(foo), dig.As(bar), dig.As(baz)`

I think the Go way is to usually give you one way to do things (no while loops, etc) but I really don't have a strong preference here.

The only thing to consider is wether the `dig.As` *itself* needs options.

`dig.Provide(myfunc, dig.As(new(bytes.Buffer), VerySpecificAsOption(7))`

dig.go
alessandrozucca: Sound good to me :-)

When I was working on it I thought to group those those two calls in a single function but wasn't sure how to call it - `checkKey` sounds a good name.

alessandrozucca:
> We can arguably mimic this behavior in the current design by explicitly providing constructors for the interfaces you want to provide

Yeah that was the suggestion on the issue too. The problem is that (at iZettle) we opted for declaring function providers for constructors which implement interfaces.

This means that some of our fx modules where we do the DI wiring are quite large. Having the As(..) makes the code much nicer and decreases the amount of functions we need to declare.

alessandrozucca:
@amckinney, @glibsm, @prashantv 

I have applied the changes requested and is ready for a re-review

abhinav:
Thanks for the change @alessandrozucca!  Besides minor nits around the error
messages, an implementation level concern I had here was that resultAs didn't
need to be its own result type. The duplication between resultSingle and
resultAs is indicative of this. To me, dig.As and the corresponding behavior
is another option for resultSingle.

I went ahead and made the required changes to avoid a bunch of back-and-forth.
Feel free to take a look.

CC @amckinney @glibsm @prashantv

alessandrozucca:
@abhinav thank you, looks better now

It did cross my mind, didn't do it for two reasons:
- I wanted to add code leaving existing code as it is without refactoring
- the name resultSingle made me think that had to provide only 1 results, adding the As() option it will provide multiple results: one as the concrete struct and one for each interface

Having said that, reusing resultSingle now makes sense to me :+1:

dig.go
```diff
@@ -127,6 +151,28 @@ func Group(group string) ProvideOption {
 	})
 }
 
+// As is a ProvideOption that specifies that the struct produced by the constructor
+// implements the given interface
```
amckinney: Duplicate doc.

dig.go
```diff
@@ -598,9 +659,10 @@ type node struct {
 
 type nodeOptions struct {
 	// If specified, all values produced by this node have the provided name
-	// or belong to the specified value group
+	// belong to the specified value group or implement any of the interfaces
```
amckinney: nit: Missing period.

abhinav:
Added a visualization test. The output looks like this:

out

dig.go
abhinav: Derp. Forgot to delet.

dig.go
```diff
@@ -535,41 +579,55 @@ func (cv connectionVisitor) Visit(res result) resultVisitor {
...
 
-		cv.keyPaths[k] = path
-
 	case resultGrouped:
 		// we don't really care about the path for this since conflicts are
 		// okay for group results. We'll track it for the sake of having a
 		// value there.
 		k := key{group: r.Group, t: r.Type}
 		cv.keyPaths[k] = path
+
```
prashantv: nit: can remove blank

dig_test.go
```diff
@@ -1905,6 +2024,46 @@ func TestProvideFailures(t *testing.T) {
 			`dig.out embeds \*dig.Out`,
 		)
 	})
+
```
prashantv: couple of additional tests:
 - use `Named` and `As` together (and verify the behaviour -- ensure both the function's return type and the `As` type are added to the container with the name).
 - provide a function that returns an interface, and then use `As`. If the `As` is for the same type, it should fail, but it should work if it's a different interface type

abhinav:
Thanks for the contribution @alessandrozucca! We'll try to release this soon.

PR #235, opened 2019-03-27
Preparing release v1.8.0

abhinav:
This sets up release v1.8.0 with the dig.As feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+### Added
+- Added `As` option for `Provide` to specify that the value produced by the
+  constructor should be made available as one or more interfaces that it
+  implements
 
 ## [1.7.0] - 2019-01-04
 ### Added
-- Added `Group` option for `Provide` to add value groups to the container without
-rewriting constructors. See package doucmentation for more information.
+- Added `Group` option for `Provide` to add value groups to the container
+  without rewriting constructors. See package doucmentation for more
+  information.
 
 ## [1.6.0] - 2018-11-06
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [1.8.0] - 2019-03-27
 ### Added
 - Added `As` option for `Provide` to specify that the value produced by the
   constructor should be made available as one or more interfaces that it
@@ -150,7 +150,7 @@ First release candidate.
 
 Initial release.
 
-[Unreleased]: https://github.com/uber-go/dig/compare/v1.7.0...HEAD
+[1.8.0]: https://github.com/uber-go/dig/compare/v1.7.0...v1.8.0
 [1.7.0]: https://github.com/uber-go/dig/compare/v1.6.0...v1.7.0
 [1.6.0]: https://github.com/uber-go/dig/compare/v1.5.1...v1.6.0
 [1.5.1]: https://github.com/uber-go/dig/compare/v1.5.0...v1.5.1


CHANGELOG.md
```diff
@@ -4,13 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
-- No changes yet.
+## [1.8.0] - 2019-03-27
+### Added
+- Added `As` option for `Provide` to specify that the value produced by the
```
glibsm:
```
dig.Provide(func () Foo { ... }, dig.As(new(bar), new(baz))`
```

Just inserts `bar` and `baz` into the graph, but not `Foo`, correct?

glibsm: Confusing sentence. Not sure if I can make it more clear, though. 

Perhaps something like?

> Added `As` option for `Provide` to allow constructor functions explicit control over which types they are being inserted into the graph for

abhinav: Foo as well. Docs clarify this with example of what the code is equivalent to,

https://github.com/uber-go/dig/blob/master/dig.go#L164-L171

glibsm: I think that may be enough for me to lobby for a train halt here. I don't think `Foo` should be included since that may not be the intention at all.

`dig.As` seems to imply _only as_, rather than `dig.AlsoAs` which is the current semantic.

abhinav:
@glibsm raised a good point here and a couple offline in chat. I'm going to hold off until I've had a chance to address those.

bithavoc:
First, great job with this library 🌮

In case you aren't aware, [godocs](go.uber.org/dig) currently contains two statements:

* > STABLE. No breaking changes will be made in this major version.
* [`func As(i ...interface{}) ProvideOption`](https://godoc.org/go.uber.org/dig#As) As is a ProvideOption that specifies that the value produced by the constructor implements one or more other interfaces.

If I weren't using Go modules I'd be quite disappointed to find out that a feature merged to master 20 days ago changed semantics somehow.

CHANGELOG.md
bithavoc: I agree, the name can be confusing but the semantic is correct given that before this function existed I used to return multiple types to satisfy implemented interfaces by the original type and the name should state that intention pretty clear.

To me the following names fit that intention:

* `AlsoAs`: `c.Provide(newBuffer, dig.AlsoAs(new(io.Reader), new(io.Writer)))`
* `TypeAlias`: `c.Provide(newBuffer, dig.TypeAlias(new(io.Reader), new(io.Writer)))`
* `Satisfy`: `c.Provide(newBuffer, dig.Satisfy(new(io.Reader), new(io.Writer)))`
* `Implement`: `c.Provide(newBuffer, dig.Implement(new(io.Reader), new(io.Writer)))`

bithavoc:
@glibsm 
> semantics of dig.Provide(func New() (Foo, io.Reader, error), dig.As(new(io.Writer). It currently fails due to inability to case reader to writer, but we'd want some extra validation here. Perpahs dig.As is only supported for a single type,error tuple?

I think it should only work for single type and error tuples. A constructor should only have one responsibility, to return it's type and all the implemented interfaces I want to satisfy.
It's too late to add that restriction to `Provide` but it makes sense to add it to `As`(see [my comment](https://github.com/uber-go/dig/pull/235#discussion_r275605151) on the naming).

glibsm:
> STABLE. No breaking changes will be made in this major version.

The stable API for us means the latest release, not the tip of master. 

I suppose it's a bit of a problematic state since godoc tracks master, rather than the latest cut release. @bithavoc Is there any way to affect godoc towards a particular release?

> If I weren't using Go modules I'd be quite disappointed to find out that a feature merged to master 20 days ago changed semantics somehow.

The readme recommends `glide get 'go.uber.org/dig#^1'` which would mean the latest one dot release (same with dep). Yes, I know we're about two dependency managers behind in the documentation. That is a different story :)

As already mentioned, the feature appears to have been merged into master too soon. There isn't much precedence for something like this in fx/dig repos as we usually land something we fully intend to release and don't use dev/master branch combo.

> I think it should only work for single type and error tuples. A constructor should only have one responsibility, to return it's type and all the implemented interfaces I want to satisfy.
It's too late to add that restriction to Provide but it makes sense to add it to As(see my comment on the naming).

Yes, it's way too late to add constructor return params validation like these. Just anecdotally, it seems our own internal usage seems to have mostly modules with constructors that return tuples, but there are notable exceptions for constructors that need to return more than one thing. For example, our logging module provides both sugared, raw, and extra logging related information.

```go
type Result struct {
	fx.Out

	Level         zap.AtomicLevel
	Config        zap.Config
	Logger        *zap.Logger
	SugaredLogger *zap.SugaredLogger
}
```

In other words, adding the tuple restriction seems artificially limiting purely for some edge cases of API ergonomics.

bithavoc:
@glibsm what's the next step

abhinav:
> @glibsm what's the next step

Hey. We still need to address issues raised in https://github.com/uber-go/dig/pull/235#pullrequestreview-219719941.

As @glibsm pointed out, the library strictly follows semver, but for
tagged releases only. Unreleased changes on master are subject to
change so making the planned changes should be okay.

We currently don't have bandwidth to address these issues. We'll try to
get to it in the future.

abhinav:
Closing PR until we can get to addressing concerns so that this is not accidentally released.

Issue #230, opened 2019-02-20
Overriding registered implementations


amirov-m:
Hello!
I would like to override already registered class instance with another one. For example in functional tests it is very useful to override registered instances. It seems like it is not possible to do that in current version. 

Are you planning to implement this feature in the future?

glibsm:
There are a few discussions on this topic in the fx repo: https://github.com/uber-go/fx/issues/653 https://github.com/uber-go/fx/issues/653

sywhang:
It's been a while since this issue was opened, but we just released v1.17.0 that contains the changes to modify the registered values of the dependency graph using `fx.Decorate` and `fx.Replace`. Will close this issue now - let me know if you have questions regarding the usage of these APIs.

seanlafferty-ibm:
Would there still be any interest in adding this to dig? I don't need the full power of fx, but would still like to be able to replace services in dig when initializing tests.

I know fx is built on dig, maybe someone could share how `replace` is implemented in fx leveraging the existing dig constructs? That might be a fine workaround.

seanlafferty-ibm:
Oh, dig can actually do this. The trick is making sure the function you pass to `decorate` is parameterless

https://github.com/uber-go/dig/blob/f02120a92598de12d0a93127beb9f60e3cf57bd3/decorate_test.go#L429-L449

PR #240, opened 2019-04-30
[WIP] Sub graphs and Decorate

srikrsna:
uber-go/fx#653

Also address #230 

Added support for creating child containers.

- [x] Containers have access to all types provided using their parents (Invoke works anywhere)
- [x] Containers have access to all types provided using their children (Invoke works anywhere)
- [x] Provide checks the entire graph (all containers both up and down) for duplicates
- [x] All success tests repeated for children
- [x] Tests for multiple and multi level children
- [x] Test for duplicate check in one of the children
- [x] Stringer updated

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -526,10 +526,6 @@ func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
 // The name of the child is for observability purposes only. As such, it
 // does not have to be unique across different children of the container.
 func (c *Container) Child(name string) *Container {
-	if name == "" {
-		panic("dig: name should not be empty")
-	}
-
 	child := &Container{
 		providers: make(map[key][]*node),
 		values:    make(map[key]reflect.Value),
--- a/dig.go
+++ b/dig.go
@@ -704,17 +704,6 @@ func (cv connectionVisitor) checkKey(k key, path string) error {
 			"cannot provide %v from %v: already provided by %v",
 			k, path, conflict)
 	}
-	if ps := cv.c.providers[k]; len(ps) > 0 {
-		cons := make([]string, len(ps))
-		for i, p := range ps {
-			cons[i] = fmt.Sprint(p.Location())
-		}
-
-		return fmt.Errorf(
-			"cannot provide %v from %v: already provided by %v",
-			k, path, strings.Join(cons, "; "))
-	}
-
 	if ps := cv.c.getRoot().getValueProviders(k.name, k.t); len(ps) > 0 {
 		cons := make([]string, len(ps))
 		for i, p := range ps {
--- a/dig.go
+++ b/dig.go
@@ -343,23 +343,18 @@ func setRand(r *rand.Rand) Option {
 // knownTypes returns the types known to this container, including types known
 // by its descendants.
 func (c *Container) knownTypes() []reflect.Type {
-	getKnowTypes := func(c *Container) []reflect.Type {
-		typeSet := make(map[reflect.Type]struct{}, len(c.providers))
-		for k := range c.providers {
-			typeSet[k.t] = struct{}{}
-		}
-
-		types := make([]reflect.Type, 0, len(typeSet))
-		for t := range typeSet {
-			types = append(types, t)
-		}
-
-		return types
+	typeSet := make(map[reflect.Type]struct{}, len(c.providers))
+	for k := range c.providers {
+		typeSet[k.t] = struct{}{}
 	}
 
-	types := make([]reflect.Type, 0)
-	for _, c := range append(c.children, c) {
-		types = append(types, getKnowTypes(c)...)
+	types := make([]reflect.Type, 0, len(typeSet))
+	for t := range typeSet {
+		types = append(types, t)
+	}
+
+	for _, c := range append(c.children) {
+		types = append(types, c.knownTypes()...)
 	}
 
 	sort.Sort(byTypeName(types))
@@ -585,7 +580,7 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
 		if c.deferAcyclicVerification {
 			continue
 		}
-		if err := verifyAcyclic(c, n, k); err != nil {
+		if err := verifyAcyclic(c.getRoot(), n, k); err != nil {
 			c.providers[k] = oldProviders
 			return err
 		}
@@ -602,7 +597,7 @@ func (c *Container) findAndValidateResults(n *node) (map[key]struct{}, error) {
 	var err error
 	keyPaths := make(map[key]string)
 	walkResult(n.ResultList(), connectionVisitor{
-		c:        c,
+		c:        c.getRoot(),
 		n:        n,
 		err:      &err,
 		keyPaths: keyPaths,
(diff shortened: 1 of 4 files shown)


srikrsna:
@abhinav 

I've added supported for sub graphs. Please review and provide feedback. Once this is approved I'll proceed to implementing `Decorate`

dig.go
```diff
@@ -332,15 +341,25 @@ func setRand(r *rand.Rand) Option {
...
+
+		return types
 	}
 
-	types := make([]reflect.Type, 0, len(typeSet))
-	for t := range typeSet {
-		types = append(types, t)
+	types := make([]reflect.Type, 0)
+	for _, c := range append(c.children, c) {
+		types = append(types, getKnowTypes(c)...)
```
abhinav: Not a major issue but I don't think this level of change was necessary. You
can let recursion handle the work here like with the other cases.

```
func (c *Container) knownTypes() []reflect.Type {
  [previous code]

  for _, c := range c.children {
    types = append(types, c.knownTypes()...)
  }

  sort.Sort(byTypeName(types))
  return types
}
```

The result is more readable in exchange for wasted effort in sorting the
children's slices, but that should be fine since this is only used on the
error path.

dig.go
```diff
@@ -479,6 +519,27 @@ func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
...
+// will be made available to the parent.
+//
+// The name of the child is for observability purposes only. As such, it
+// does not have to be unique across different children of the container.
+func (c *Container) Child(name string) *Container {
+	child := &Container{
+		providers: make(map[key][]*node),
+		values:    make(map[key]reflect.Value),
+		groups:    make(map[key][]reflect.Value),
+		rand:      rand.New(rand.NewSource(time.Now().UnixNano())),
```
abhinav: We shouldn't instantiate a new source of randomness. We should use the
parent's `rand` so that the behavior of the child is deterministic in tests.

```suggestion
		rand:      c.rand,
```

dig.go
```diff
@@ -643,7 +704,7 @@ func (cv connectionVisitor) checkKey(k key, path string) error {
 			"cannot provide %v from %v: already provided by %v",
 			k, path, conflict)
 	}
-	if ps := cv.c.providers[k]; len(ps) > 0 {
+	if ps := cv.c.getRoot().getValueProviders(k.name, k.t); len(ps) > 0 {
```
abhinav: You can likely switch out the Container provided to `connectionVisitor` when
it's instantiated to the root container and leave this as-is.

dig_test.go
```diff
@@ -941,6 +942,1029 @@ func TestEndToEndSuccess(t *testing.T) {
 	})
 }
 
+func TestChildren(t *testing.T) {
+	t.Parallel()
+
+	t.Run("pointer constructor", func(t *testing.T) {
+		c := New()
+
```
abhinav: I don't think we need to copy all the end-to-end tests. We can parameterize
them on how they do provides and invokes instead.

So we move the existing tests into a `testEndToEnd` helper function.

```
func testEndToEnd(
    t *testing.T,
    provide func(interface{}, ...ProvideOption) error,
    invoke func(interface{}, ...InvokeOption) error,
) {
    ...
}
```

Replacing all `c.Provide` and `c.Invoke` calls in those tests with calls to
the `provide` and `invoke` arguments. (The `c := New()` can all be dropped.)

Then in the actual `TestEndToEnd` test, we can do,

```
func TestEndToEnd(t *testing.T) {
    t.Run("root", func(t *testing.T) {
        c := New()
        testEndToEnd(t, c.Provide, c.Invoke)
    })

    t.Run("provide in parent, invoke in child", func(t *testing.T) {
        parent := New()
        child := parent.Child("child")
        testEndToEnd(t, parent.Provide, child.Invoke)
    })

    t.Run("provide in child, invoke in parent", func(t *testing.T) {
        parent := New()
        child := parent.Child("child")
        testEndToEnd(t, child.Provide, parent.Invoke)
    })
}
```

dig_test.go
```diff
@@ -951,13 +1975,361 @@ func TestGroups(t *testing.T) {
...
+			`could not build arguments for function "go.uber.org/dig".TestGroups`,
+			`could not build value group string\[group="x"\]:`,
+			`function "go.uber.org/dig".TestGroups\S+ \(\S+:\d+\) returned a non-nil error:`,
+			"great sadness",
+		)
+		assert.Equal(t, gaveErr, RootCause(err))
+	})
+}
+
+func TestChildrenGroups(t *testing.T) {
```
abhinav: Same as above. We can parameterize rather than duplicating.

dig.go
srikrsna: I could, but even then it has to call `getValueProviders` instead of accessing providers directly

abhinav: Oh, yeah, true you'd still have to use getValueProviders to have access to the whole graph.

srikrsna:
@abhinav I've updated the tests. One more thing I needed direction on is how should the `Stringer` method be updated. Should it also print all the nodes of the parents? or all the nodes of children? or both?

I would go for both since it would be consistent with the behaviour of `Invoke` and `Provide`. But I'm not sure.

abhinav:
Hey, sorry for the delay. That's a fair point. It might be a bit surprising if
`String()` prints only 5 types but the container has access to 50. That said,
the expectation is still for `String()` to print a representation of the
object under inspection, even if has knowledge of more.

So I think it's best if String only print its own nodes and recurses into
children. It could also print the address of the parent (if any) so we know
that part of the graph is hidden.

abhinav:
I made a minor simplification to the tests. Please pull that before you make
more changes.

srikrsna:
I'll update the `String()` method to print all its children and also the parent if any. Also, can I proceed to implement `Decorate`?

abhinav:
> I'll update the `String()` method to print all its children and also the parent if any.

Just the address of the parent instead of the contents might be less noisy and would help differentiate between `parent` and `child` in the output.

> Also, can I proceed to implement `Decorate`?

Yes, definitely! Although I would make that a separate PR on top of this one so that this can be landed when ready and that can be iterated on independently.

srikrsna:
@abhinav Apologies for the delay. I believe this is complete and ready to be merged.

abhinav:
Hey @srikrsna! Thanks! We haven't forgotten about this but we're a bit swamped
right now. I'll try to get into this as soon as possible.

ryus08:
I know this is pretty darn old at this point, but FWIW, is a type of feature I personally look for in an IoC container, and seemed pretty close to good to go last year.

jstroem:
This is what I awesome! Have been waiting for this feature..
@srikrsna seems that this is almost done, will you be willing to sign the CLA?

srikrsna:
Done

jstroem:
@abhinav any way to can push on this?

yiminc:
@srikrsna @abhinav why not merge this PR?

abhinav:
Hey folks, apologies for the radio silence here.

I know it's been a while but we're trying to pick up support for decorate/replace again,
so we'll try to update the relevant tasks as soon as we have something.

That said, the reason this PR was not merged was this:
we started reconsidering whether subgraphs were the best way to implement this feature.

Unfortunately, we didn't write down those conversations, but from what I can recall:
We didn't like that the semantics of subgraphs were not super clear.
I'm the one who suggested those semantics, and I still couldn't tell you off the top of my head
whether a subgraph has access to everything provided by a parent and everything provided by a sibling.
That it does, based on the name subgraph, is surprising behavior.

So we weren't ready to commit to subgraphs at the time,
and then the project just got buried under other priorities.

All's not lost, though. As I said, we're trying to pick up the project again.
We'll try to update the relevant tasks (#230 and https://github.com/uber-go/fx/issues/653) as soon as we have something.

Issue #242, opened 2019-06-13
A hammer is a bad tool to dig with.


n3wscott:
You should not dig with a hammer. Fix to follow.

PR #243, opened 2019-06-13
A pick is a better tool to dig with.

n3wscott:
fixes: https://github.com/uber-go/dig/issues/242


glibsm:
I... dig it.

🥁

glibsm:
Why closed?

n3wscott:
I can't sign the CLA.

glibsm:
Oh. Can't as in it's not working or can't as in you won't?

One we should be able to easily fix, another is a totally fine stance as well :)

n3wscott:
I am sure the button works. :(

glibsm:
I proposed an alternative approach... #244

PR #244, opened 2019-06-13
Now with more picks in the emoji

### LINKED ISSUE #242: A hammer is a bad tool to dig with.

glibsm:
Fixes #242

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# :hammer: dig [![GoDoc][doc-img]][doc] [![GitHub release][release-img]][release] [![Build Status][ci-img]][ci] [![Coverage Status][cov-img]][cov] [![Go Report Card][report-card-img]][report-card]
+# :hammer_and_pick: dig [![GoDoc][doc-img]][doc] [![GitHub release][release-img]][release] [![Build Status][ci-img]][ci] [![Coverage Status][cov-img]][cov] [![Go Report Card][report-card-img]][report-card]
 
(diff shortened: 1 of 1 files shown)


n3wscott:
lololol

LGTM

blampe:
Nice fix!

kriskowal:
How much effort would it be to apply for SHOVEL and BUCKET or MATOCK, or AXE MATTOX emoji? Maybe EXCAVATOR? 🚜  Perhaps we should use SPOON as a stopgap? 🥄

mh-park:
Proposal: 🥄

robbertvanginkel:
Guess I'm not very picky with my reviews.

glibsm:
After after some more <strike>beers</strike> significant thought taking into account all the thoughtful responses in this thread, it only stands to reason that we need to fully agree on the direction of the emoji.

For that purpose I propose we collaborate and agree on [RFC: dig emoji](https://docs.google.com/document/d/1d3T9qEZWQCj8RcHrLfyj2vCmEf-sEVoLVxsTq4Zh12E/edit?usp=sharing) prior to the merging of this pull request.

kriskowal:
@glibsm, for the RPC, should leave a historical note: English used to more clearly separate the meanings of "dig" and "delve" based on which tool you were using. Digging was reserved for picks and mattocks, whereas delve was reserved for shovels and spades.

# ⛏  (dig)
# ♠ 🥄 (delve)

PR #250, opened 2019-10-30
Migrate to Go modules

abhinav:
This change migrates Dig to Go modules. This has the effect of
simplifying the Makefile, and allows us to delete the glide.lock. We
will keep the glide.yaml around to provide constraints for our legacy
ecosystem.

Note to reviewers: Each commit is meant to be reviewed individually.


Makefile
```diff
@@ -1,52 +1,47 @@
...
+	@./check_license.sh | tee -a lint.log
 	@[ ! -s lint.log ]
 
+$(GOLINT):
+	go install golang.org/x/lint/golint
+
 .PHONY: test
 test:
-	@.build/test.sh
+	go test -race ./...
```
rhang-uber: Nice!

abhinav:
Reverted the change we didn't want released. This is ready to review/merge now.

PR #252, opened 2019-11-14
Implement a dig.As ProvideOption

### LINKED ISSUE #197: Allow to specify an interface when providing a constructor

abhinav:
This brings back #233.

Per https://github.com/uber-go/dig/pull/235#pullrequestreview-219719941,
the issues we need to resolve are,

1. `dig.As` seems to indicate that it's a total override of the provided
   type. The way it currently works is it appends other interfaces on
   top of whatever the constructor already returns
2. semantics of `dig.Provide(func New() (Foo, io.Reader, error), dig.As(new(io.Writer)`.
   It currently fails due to inability to case reader to writer, but
   we'd want some extra validation here. Perhaps `dig.As` is only
   supported for a single type, error tuple?

Closes #197

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -62,12 +62,19 @@ func (f optionFunc) applyOption(c *Container) { f(c) }
 type provideOptions struct {
 	Name  string
 	Group string
+	As    []interface{}
 }
 
 func (o *provideOptions) Validate() error {
-	if len(o.Group) > 0 && len(o.Name) > 0 {
-		return fmt.Errorf(
-			"cannot use named values with value groups: name:%q provided with group:%q", o.Name, o.Group)
+	if len(o.Group) > 0 {
+		if len(o.Name) > 0 {
+			return fmt.Errorf(
+				"cannot use named values with value groups: name:%q provided with group:%q", o.Name, o.Group)
+		}
+		if len(o.As) > 0 {
+			return fmt.Errorf(
+				"cannot use dig.As with value groups: dig.As provided with group:%q", o.Group)
+		}
 	}
 
 	// Names must be representable inside a backquoted string. The only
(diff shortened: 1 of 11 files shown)


glibsm:
> semantics of dig.Provide(func New() (Foo, io.Reader, error), dig.As(new(io.Writer).
It currently fails due to inability to case reader to writer, but
we'd want some extra validation here. Perhaps dig.As is only
supported for a single type, error tuple?

I see the following:
1. As you said, only restrict to one type and error (has edge cases)

2. `dig.As` can take a vararg which gets matched to the positional arguments (minus error)

abhinav:
Superseded by #293

PR #255, opened 2019-12-14
Add dig.Optional for marking optional params

orishu:
Realized that fx can be improved so certain parameters provided, while in their absence, a default parameter value would be used.

Add a new `dig.Optional` type that can be embedded in structs to mark
that the dependency on them is optional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+- Added `dig.Optional` embedded type to mark parameter structs as optional.
 
 ## [1.8.0] - 2019-11-14
 ### Changed
--- a/dig.go
+++ b/dig.go
@@ -673,6 +673,10 @@ func (n *node) Call(c containerStore) error {
 
 // Checks if a field of an In struct is optional.
 func isFieldOptional(f reflect.StructField) (bool, error) {
+	if IsOptional(f.Type) || (f.Type.Kind() == reflect.Ptr && IsOptional(f.Type.Elem())) {
+		return true, nil
+	}
+
 	tag := f.Tag.Get(_optionalTag)
 	if tag == "" {
 		return false, nil
--- a/dig_test.go
+++ b/dig_test.go
@@ -320,6 +320,13 @@ func TestEndToEndSuccess(t *testing.T) {
 		type type3 struct{}
 		type type4 struct{}
 		type type5 struct{}
+		type type6 struct {
+			Optional
+		}
+		type type7 struct {
+			Optional
+		}
+
 		constructor := func() (*type1, *type3, *type4) {
 			return &type1{}, &type3{}, &type4{}
 		}
@@ -333,6 +340,8 @@ func TestEndToEndSuccess(t *testing.T) {
 			T3 *type3 `unrelated:"foo=42, optional"`        // type in the graph with unrelated tag
 			T4 *type4 `optional:"true"`                     // optional type present in the graph
 			T5 *type5 `optional:"t"`                        // optional type NOT in the graph with "yes"
+			T6 *type6 // pointer to optional struct
+			T7 type7  // optional struct
 		}
 		require.NoError(t, c.Provide(constructor))
 		require.NoError(t, c.Invoke(func(p param) {
@@ -341,6 +350,7 @@ func TestEndToEndSuccess(t *testing.T) {
 			assert.NotNil(t, p.T3, "required type with unrelated tag not in the graph")
 			assert.NotNil(t, p.T4, "optional type in the graph should not return nil")
 			assert.Nil(t, p.T5, "optional type not in the graph should return nil")
+			assert.Nil(t, p.T6, "optional struct not in the graph should return nil")
 		}))
 	})
 
@@ -2600,6 +2610,28 @@ func TestInvokeFailures(t *testing.T) {
 			"type dig.A is not in the container, did you mean to Provide it?",
 		)
 	})
+
+	t.Run("optional struct dependency", func(t *testing.T) {
+		type missing struct{}
+
+		type params struct {
+			Optional
+
+			String string
+		}
+
+		c := New()
+
+		// Should still be able to invoke a function that takes params, since it's
+		// optional.
+		var count int
+		err := c.Invoke(func(p params) {
+			count++
+		})
+		assert.NoError(t, err, "unexpected invoke error")
+		assert.Equal(t, 1, count, "expected invoke function to be called")
+	})
+
 }
 
 func TestNodeAlreadyCalled(t *testing.T) {
--- a/param.go
+++ b/param.go
@@ -67,6 +67,8 @@ func newParam(t reflect.Type) (param, error) {
 		return nil, fmt.Errorf("cannot depend on result objects: %v embeds a dig.Out", t)
 	case IsIn(t):
 		return newParamObject(t)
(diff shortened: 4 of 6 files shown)


abhinav:
Hey, from what I can understand, this is intended to be a replacement for
`dig.In` where all fields of the struct are optional instead of individual
fields. I'm not sold on the complexity that would get added by this, or the
duplication of dig.In vs dig.Optional. Is there a strong reason that you
cannot add `optional:"true"` tags on all fields of your `dig.In` struct?

orishu:
Let's look at the potential benefit of this feature through the following example.

An fx service ServiceA, depends on another service, ServiceB. ServiceB can use a some integer and string parameters to initialize itself.

Basic setup:
```
type ServiceA interface{}
type ServiceB interface{}

type ServiceBParams struct {
        Param1 int
        Param2 string
}

func NewServiceB(svcbParams ServiceBParams) ServiceB {
        return &serviceB{...}
}

func NewServiceA(svcb ServiceB) ServiceA {
        return &serviceA{B: svcb}
}
```

Now, for Target X, we may want to use the default parameters:
```
fx.Provide(NewServiceA, NewServiceB)
```
While for Target Y (e.g. unit tests), we may want to tweak it a little:
```
fx.Provide(NewServiceA, NewServiceB, func() ServiceBParams { return ServiceBParams{Param1:123} })
```
Now, it's true that we *could* instantiate Target X with an empty instance of `ServiceBParams`:
```
fx.Provide(NewServiceA, NewServiceB, func() ServiceBParams { return ServiceBParams{} })
```

But that idiom can easily clutter the provide option list. If we depend on 20 services and each one of them has some default parameter struct that it may need to use, our provide option list would have 40 items, 20 of which would be lambdas or defined functions, which is quite ugly for a main.go or options.go file.

abhinav:
Right, but it's also possible to do the following to get the same effect.

```
type ServiceBParams struct {
  dig.In
    
  Param1 int    `optional:"true"`
  Param2 string `optional:"true"`
}
```

And if writing that is too much work, it's also possible, given the existing
ServiceBParams struct, to dynamically generate an anonymous equivalent with
`optional:"true"` set to true on all fields.

Both versions are possible without introducing new features to Dig and Fx,
which might conflict with how we handle dig.In.

orishu:
Regarding your suggestion, if you try to provide a value of type `ServiceBParams`, you will hit an exception getting raised here:
https://github.com/uber-go/dig/blob/cfca627871180f3a5d7dab4453f7598e0c758013/result.go#L70

abhinav:
Yes, you wouldn't provide it, only consume it. If you want to "provide"
ServiceBParams, then instead of making it the `dig.In`, you'd put it inside
another struct:

```
type ServiceBParams struct { ... }

type ServiceBArgs struct {
  dig.In
  
  P ServiceBParams `optional:"true"`
}

func NewServiceB(ServiceBArgs) ServiceB
```

And as before, it's possible to take the existing `func
NewServiceB(ServiceBParams) ServiceB` function and generate an anonymous
function which accepts an anonymous generated dig.In struct with
ServiceBParams field and `optional:"true"` set. This generated function may be
provided to Dig/Fx.

orishu:
I see your points. Let's move the discussion forward by being explicit about the tradeoff:

On one hand, introducing an additional embedded type to the dig framework (`dig.Optional`) may create complexity. On the other hand, fx users have been struggling and polluting their code with workarounds for providing optional compile-time arguments to their services.

You brought up two arguments which I'd like to challenge:
1. _Introducing `dig.Optional` might conflict with how we handle `dig.In`_. Would you mind coming up with a code example that illustrates such possible conflict?

2. _The complexity of wrapping the params struct as `optional:"true"` in a `dig.In` struct can be abstracted by a utility function (arguably provided by a library?) that transforms the normal `ServiceB` constructor to one that accepts the wrapper struct and passes the arguments_. I suspect that such utility function is either not simple enough to expect user to implement it, or not generic enough to work with all types. I also expect the use of such function to impact code readability.

glibsm:
In very early prototypes of fx/dig there was a concept of embedding a special type in the structs directly (very similar to this PR) to signal optionality, very similar to this PR.

While there is an argument that can be made about it being useful (generally, there is always an argument to be made of a new feature being useful for one reason or another), that approach was rejected due to framework bleeding into the "business logic" models.

You can also make the case that using dig.In/Out "bleeds", but it doesn't need to (or in fact, can be) embedded directly into your models and objects (there is a runtime check for that). A new intermediate struct needs to be created to act as a bridge.

In the event that a codebase needs to be transitioned to a hypothetically incompatible future dig.v2, or indeed any other framework, the core logic shouldn't be affected. Only the aforementioned bridges.

At least, that's the intention.

orishu:
Thanks for your response @glibsm. Not "bleeding" embedded `fx.Optional` structs into business logic makes a valid argument.
So I'll "zoom out" to a higher-level discussion. The first stated goal of Fx is "Make dependency injection easy". As a (senior) software engineer, I don't find Fx easy enough to work with, somewhat due to the excess of struct indirection (a.k.a. "bridges"), and other workarounds such as the usage of different "base.yaml" and "testing.yaml" configuration files due to strict dependency topology.
One metric for success of an open-source project is the extent in which it is adopted by companies of various sizes and areas. It is my opinion that Fx can be more successful than it is today if we'd make it easier to use by relaxing some of those strict ("anti-bleeding") limitations.
While it is the project maintainers authority and responsibility to do what is best for the project, my point here that Fx experts such as you guys have some blindspots with regards to how junior (and senior) engineers might be struggling.

abhinav:
That's valuable feedback, @orishu. We've been looking for ways to make certain
things in Fx easier to do, or if they have to be complex, easier to
understand. We'll keep your feedback in mind in our future plans for Fx.

That said, we strongly prefer not to bleed Dig or Fx into application logic.
That has been a very deliberate design choice and we will continue to stick to
it.

As for the change:

- We're not sure if this is how we want people to use optionality. Whether a
  value is needed to do some work is usually a consumer's decision, not the
  producer's. This makes optionality a producer-level decision, leaving the
  consumer with no choice but to validate that the value is non-zero.
- This would need to somehow support named values. Whether that would take the
  form of a new embed or a tag on the embed is unclear.
- This sets a precedent for more embedded types to customize behavior, further
  increasing the surface area of Fx and Dig that leaks into business logic.
- The code changes required are more involved than what has been proposed. For
  example, an obvious case that is missing: we'd need to address such a type
  being accepted directly as an argument versus being embedded inside a
  dig.In, which this change doesn't yet do.

Overall, we're of the opinion that the added complexity of this feature is not
justifiable, especially because alternatives exist and the feature is
implementable outside Dig/Fx without any changes to either library.

Hope this clarifies things from our end.

orishu:
Understood.

I don't wish to further impose a discussion here, but I will mention that if we were to continue the discussion, I would have disagreed with two of the points:

> This would need to somehow support named values. Whether that would take the
> form of a new embed or a tag on the embed is unclear.

I don't see how this is needed.

> The code changes required are more involved than what has been proposed. For
> example, an obvious case that is missing: we'd need to address such a type
> being accepted directly as an argument versus being embedded inside a
> dig.In, which this change doesn't yet do.

The code change that I submitted should support that case, it's covered in the added unit tests as well.


I may be wrong about one of the above issues, or both, I didn't explore them too diligently because of the status of this PR. Just wanted to post my minor objections to those points, for reference.

abhinav:
> > The code changes required are more involved than what has been proposed. For
> > example, an obvious case that is missing: we'd need to address such a type
> > being accepted directly as an argument versus being embedded inside a
> > dig.In, which this change doesn't yet do.
> 
> The code change that I submitted should support that case, it's covered in the added unit tests as well.

It is indeed. My bad.

Thanks for the discussion, @orishu. It's always useful to understand use cases
that weren't covered out-of-the-box.

PR #256, opened 2019-12-17
errors: Add errf to support multi-line errors

abhinav:
*Commits are individually reviewable.*

Errors printed by dig are currently a massive wall of text. We've
received consistent feedback that these errors are difficult to read.
To help make these error messages more readable, we're going to split
them across multiple lines when printed with `%+v` instead of `%v`.

As a first step to doing this, we need to inform the system where
individual sections of error messages end so that newlines can be
introduced. A natual point for most errors in the form `foo: bar` is
after the `:`. So, `something went wrong: great sadness` could be
printed as the following with `%+v`.

    something went wrong:
    great sadness

To inform the system where error messages end, we're going to resort to
chaining error messages with our existing `wrappedError` typed. The
error above will become,

    wrappedError{
      msg: "something went wrong",
      cause: errors.New("great sadness"),
    }

Doing this by hand is cumbersome, so to build these error chains
conveniently, this change introduces an `errf(string, ...interface{})`
function. This will replace use of `fmt.Errorf` and our previous
`errWrapf` functions. It supports building chains of multiple formatted
errors by adding another printf-style string and its arguments when the
arguments for the previous string end.

    errf(
      "could not process %q (ID: %v)", thing, thingID,
      "request %v is invalid", requestID,
    )

The above will build an error chain where the second string is the
cause of the first. So the above is roughly equivalent to the following.

    wrappedError{
      msg: `could not process "foo" (ID: 12)`,
      err: errors.New("request 42 is invalid"),
    }

In addition to that, the function supports a standalone `error`
argument in place of a string for its last argument. This helps use
replace the `errWrapf` function we previously used.

    errf("could not parse %q", s, err)
    // equivalent to
    errWrapf(err, "could not parse %q", s)

    // Output:
    // could not parse "foo": not a number

You can interpret each string and its formatting arguments as one node
in the chain of errors, where the final node can be another such string
and its arguments, or a different `error`.

---

This change holds a series of individually reviewable changes that
introduce this functionality and transition existing usage of
`fmt.Errorf` and `errWrapf` over to it. Additionally, this also breaks
apart `fmt.Errorf`-based error messages on sentence boundaries so that
they are chained together, with the intent of having them broken across
multiple lines in the future.

    fmt.Errorf("cannot do x: thing y is invalid: %v", err)
    // becomes
    errf("cannot do x", "thing y is invalid", err)

This has no user-facing changes at this time, but the idea is that
we'll be able to print multi-line error messages, inserting newlines at
these boundaries.

Refs T2587659

### DIFF
--- a/error.go
+++ b/error.go
@@ -22,6 +22,7 @@ package dig
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"reflect"
 	"sort"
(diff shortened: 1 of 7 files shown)


error.go
```diff
@@ -54,6 +55,113 @@ func RootCause(err error) error {
...
+					"Please file an issue at https://github.com/uber-go/dig/issues/ "+
+					"and provide the following message: "+
+					"unexpected errf-argument type %T", arg))
+		}
+	}
+
+	// Prepend msg to the args list so that we can re-use the same
+	// args processing logic. The msg is a string just for type-safety of
+	// the first error.
+	return buildErrf(append([]interface{}{msg}, args...))
```
twilly: Make the array to avoid excessive allocations.

abhinav: Done

error.go
```diff
@@ -54,6 +55,113 @@ func RootCause(err error) error {
...
+//
+//   errf(
+//     "could not process %v", thing,
+//     "date %q could not be parsed", thing.Date,
+//     parseError,
+//  )
+func errf(msg string, args ...interface{}) error {
+	// To ensure that errf controls all calls to buildErrf, we make it a
+	// closure instead of its own function. This allows buildErrf to assume
+	// that len(args) > 0 for all calls.
```
r-hang: To me this sentences implies that making buildErrf being in the closure allows for len(arg) > 0 to be assumed. I don't think that is quite true, what I do think is true is that in ensures that buildErrf can only be used through errF, which limits API surface.

mh-park:
lgtm too. Can you also add a little more context and/or link to an issue to the pull request message on why a multi-line error is helpful and a desired change?

error.go
```diff
@@ -54,6 +55,113 @@ func RootCause(err error) error {
 	}
 }
 
+// errf is a version of fmt.Errorf with support for a chain of multiple
```
vyshah: From the documentation at least, this seems like `errors.Wrapf` with a slightly tweaked API: 

- You can bundle multiple errors at 1 callsite: why would we want to do this? I can't easily imagine a scenario where multiple errors need to be nested at once.

- The `error` argument is optional unlike Wrapf. This seems more ergonomic if we're actually expecting _large_ error chains but still, easy enough to work around.

error.go
```diff
@@ -54,6 +55,113 @@ func RootCause(err error) error {
...
+//   errf(
+//     "could not process %v", thing,
+//     "date %q could not be parsed", thing.Date,
+//     parseError,
+//  )
+func errf(msg string, args ...interface{}) error {
+	// To ensure that errf controls all calls to buildErrf, we make it a
+	// closure instead of its own function. This allows buildErrf to assume
+	// that len(args) > 0 for all calls.
+	var buildErrf func([]interface{}) error
```
vyshah: I can appreciate the utility of recursion for building a nested structure but I also feel like it would be possible to make this pretty readable if you just used iteration instead of a closure + recursion.

error.go
vyshah: It's true because errf has at least 1 argument due to it's signature and the way the recursion is written, the base case is buildErrf is called with 1 argument. 

Still, this comment doesn't really clarify much for someone reading this code - better to just explain that buildErrf is a closure being called recursively.

error.go
```diff
@@ -54,6 +55,113 @@ func RootCause(err error) error {
...
+				msg: msg,
+				err: buildErrf(args),
+			}
+		case error:
+			if len(args) > 0 {
+				panic(fmt.Sprintf(
+					"It looks like you have found a bug in dig. "+
+						"Please file an issue at https://github.com/uber-go/dig/issues/ "+
+						"and provide the following message: "+
+						"error must be the last element but got %v", args))
```
vyshah: Pull these out into a shared `bugPanic(string specificCause)` function.

error.go
```diff
@@ -54,6 +55,113 @@ func RootCause(err error) error {
...
+	}
+
+	// Prepend msg to the args list so that we can re-use the same
+	// args processing logic. The msg is a string just for type-safety of
+	// the first error.
+	return buildErrf(append([]interface{}{msg}, args...))
+}
+
+// Returns the number of formatting arguments in the provided string. Does not
+// include %-escaped %s.
```
vyshah: "i.e. the substring `%%` specifically"

error.go
abhinav: Reworded

error.go
abhinav: Ah, It looks like I didn't mention in the PR why we want to make error
chains like this. I'll add it to the PR message but:

We want to chain errors where one portion of a long error message
reasonably ends, so that it chains into the next one. Using the chains
as markers for separate sentences, we'll be able to split the error
messages across multiple lines so that dig stops printing massive walls
of text.

error.go
abhinav: I specifically wrote an iterative version of this because I felt that
the extra function needed for the recursive approach was a bit icky.
I went back to this because the iterative approach becomes difficult to
follow in terms of exactly how the information flows, since you have to
do it in two stages:

- `[]interface{}` to `[]string` based on formatting arguments, and a
  `rootCause error`.
- Build up the chain in reverse from `[]string` with `rootCause` as the
  cause of the final error in the chain.

I can switch to it if you'd like, but I think for someone reading this
code, the flow of information there is less clear.

error.go
abhinav: @mh-park @vyshah added more to the PR description

error.go
abhinav: There are other such panics in the dig code base so I would rather not
add that as a rider on this change. Created #259 to track this.

PR #257, opened 2019-12-18
Print multi-line error messages with %+v

abhinav:
*Commits are individually reviewable.*

This adds a series of commits that add support for printing multi-line error
messages when `%+v` is used.

To do so, we implement [`fmt.Formatter`][1] on all relevant error types,
interpreting `%v` the old way and `%+v` as a signal to be prettier.

  [1]: https://golang.org/pkg/fmt/#Formatter

Implementation `fmt.Formatter` as-is for the error-chain types (those
implementing the `causer` interface) causes a fair amount of duplication so
this change also adds some convenience around it. Specifically, these types
now implement a means of printing just their error message to the output,
deferring to a shared implementation to print the chain of causes.

Finally, as tests that verify error messages all use the same helper
(`assertErrorMatches`), the helper was split into two subtests. One of the
subtests verifies the old behavior and the latter verifies the new `%+v`
behavior.

Refs T2587659

### DIFF
--- a/error.go
+++ b/error.go
@@ -24,6 +24,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"reflect"
 	"sort"
 
@@ -34,11 +35,44 @@ import (
 // Errors which know their underlying cause should implement this interface to
 // be compatible with RootCause.
 //
-// We use an unexported "cause" method instead of "Cause" because we don't
-// want dig-internal causes to be confused with the cause of the user-provided
-// errors. (For example, if the users are using github.com/pkg/errors.)
+// We use unexported methods because we don't want dig-internal causes to be
+// confused with the cause of the user-provided errors. For example, if we
+// used Unwrap(), then user-provided methods would also be unwrapped by
+// RootCause. We want RootCause to eliminate the dig error chain only.
 type causer interface {
+	fmt.Formatter
+
+	// Returns the next error in the chain.
 	cause() error
+
+	// Writes the message or context for this error in the chain.
+	//
+	// verb is either %v or %+v.
+	writeMessage(w io.Writer, verb string)
+}
+
+// Implements fmt.Formatter for errors that implement causer.
+//
+// This Format method supports %v and %+v. In the %v form, the error is
+// printed on one line. In the %+v form, the error is split across multiple
+// lines on each error in the error chain.
+func formatCauser(c causer, w fmt.State, v rune) {
+	multiline := w.Flag('+') && v == 'v'
+	verb := "%v"
+	if multiline {
+		verb = "%+v"
+	}
+
+	// "context: " or "context:\n"
+	c.writeMessage(w, verb)
+	io.WriteString(w, ":")
+	if multiline {
+		io.WriteString(w, "\n")
+	} else {
+		io.WriteString(w, " ")
+	}
+
+	fmt.Fprintf(w, verb, c.cause())
 }
 
 // RootCause returns the original error that caused the provided dig failure.
@@ -174,10 +208,19 @@ type wrappedError struct {
 	msg string
 }
 
-func (e wrappedError) cause() error { return e.err }
+var _ causer = wrappedError{}
 
-func (e wrappedError) Error() string {
-	return fmt.Sprintf("%v: %v", e.msg, e.err)
+func (e wrappedError) cause() error {
+	return e.err
+}
+
+func (e wrappedError) writeMessage(w io.Writer, _ string) {
+	io.WriteString(w, e.msg)
+}
+
+func (e wrappedError) Error() string { return fmt.Sprint(e) }
+func (e wrappedError) Format(w fmt.State, c rune) {
+	formatCauser(e, w, c)
 }
 
 // errProvide is returned when a constructor could not be Provided into the
@@ -187,10 +230,19 @@ type errProvide struct {
 	Reason error
 }
 
-func (e errProvide) cause() error { return e.Reason }
+var _ causer = errProvide{}
 
-func (e errProvide) Error() string {
-	return fmt.Sprintf("function %v cannot be provided: %v", e.Func, e.Reason)
+func (e errProvide) cause() error {
+	return e.Reason
+}
+
+func (e errProvide) writeMessage(w io.Writer, verb string) {
+	fmt.Fprintf(w, "cannot provide function "+verb, e.Func)
+}
+
+func (e errProvide) Error() string { return fmt.Sprint(e) }
+func (e errProvide) Format(w fmt.State, c rune) {
+	formatCauser(e, w, c)
 }
 
 // errConstructorFailed is returned when a user-provided constructor failed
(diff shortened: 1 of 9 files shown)


abhinav:
Not yet ready for review.

abhinav:
Ready for review

internal/digreflect/func_test.go
```diff
@@ -55,7 +56,6 @@ func TestInspectFunc(t *testing.T) {
 		// We don't match the exact file name because $GOPATH can be anywhere
 		// on someone's system. Instead we'll match the suffix.
 		wantFileSuffix string
-		wantStringLike string
```
r-hang: Why does this change mean the removal of the wantStringLike test scenarios?

internal/digreflect/func.go
```diff
@@ -45,8 +45,21 @@ type Func struct {
...
+	return fmt.Sprint(f)
+}
+
+// Format implements fmt.Formatter for Func, printing a single-line
+// representation for %v and a multi-line one for %+v.
+func (f *Func) Format(w fmt.State, c rune) {
+	if w.Flag('+') && c == 'v' {
+		// "path/to/package".MyFunction
+		// 	path/to/file.go:42
+		fmt.Fprintf(w, "%q.%v", f.Package, f.Name)
```
r-hang: Not a quite a nit, but looking at the [https://golang.org/pkg/fmt/](docs for fmt), it seems that there isn't an expectation of "+v" being being multiline. However, this is desirable in the vast majority of use cases -- with the only far fetched exception that I can think of being someone manually parsing the dig logging output and unexpectedly encountering a newline.

error_test.go
```diff
@@ -97,6 +98,26 @@ type consumingFinder interface {
 	Find(got string) (rest string, ok bool)
 }
 
+func runFinders(original string, finders []consumingFinder) error {
```
r-hang: Just to confirm my understanding, runFinders was created so that you could share the old
```
for _, f := range finders {
```
logic among the "single line" and "multi line" tests.

error.go
```diff
@@ -371,21 +370,77 @@ func (e errParamGroupFailed) updateGraph(g *dot.Graph) {
...
-func newErrMissingType(c containerStore, k key) errMissingType {
+// Format prints a string representation of missingType.
+//
+// With %v, it prints a short representation ideal for an itemized list.
+//
+//   io.Writer
+//   io.Writer: did you mean *bytes.Buffer?
+//   io.Writer: did you mean *bytes.Buffer, or *os.File?
+//
+// With %+v, it prints a longer representation ideal for standalone output.
```
r-hang: This point was initially counter-intuitive to me. But the point you are trying to make is that in the world of "+v" or multiline the printed comments can be more verbose because you have more space.

Why not always opt for the shorter message? Even in the shortest form, the message has to be well understood for a minimum UX. If the message is already well understood, why not just use the short message always? The net benefit is less code to maintain.

error.go
```diff
@@ -34,11 +35,44 @@ import (
...
-// want dig-internal causes to be confused with the cause of the user-provided
-// errors. (For example, if the users are using github.com/pkg/errors.)
+// We use unexported methods because we don't want dig-internal causes to be
+// confused with the cause of the user-provided errors. For example, if we
+// used Unwrap(), then user-provided methods would also be unwrapped by
+// RootCause. We want RootCause to eliminate the dig error chain only.
 type causer interface {
+	fmt.Formatter
+
+	// Returns the next error in the chain.
```
r-hang: s/next/preceding/g ?

error.go
```diff
@@ -34,11 +35,44 @@ import (
...
 	cause() error
+
+	// Writes the message or context for this error in the chain.
+	//
+	// verb is either %v or %+v.
+	writeMessage(w io.Writer, verb string)
```
r-hang: what if the verb assumption is violated? Should that be an explcit error case?

error.go
```diff
@@ -187,10 +230,19 @@ type errProvide struct {
 	Reason error
 }
 
-func (e errProvide) cause() error { return e.Reason }
+var _ causer = errProvide{}
```
r-hang: I'm not familiar with the reasoning behind the pattern of var _ causer

internal/digreflect/func_test.go
abhinav: Oh, that's a rider. It was an unused field.

error_test.go
abhinav: Yep

internal/digreflect/func.go
abhinav: Yeah there are no expectations on the output. Starting with multierr's errors, we've been taking the position that if needed, `%+v` is the "richly formatted" version. So this is just carrying that over here.

error.go
abhinav: This divergence was made with the intent of being close to the existing output. Before errMissingType and errManyMissingTypes were merged into errMissingTypes, we had that distinction in the output. The idea was to reduce the number of tests that had to be updated in this change, since we can iterate on the error messages later as well.

error.go
abhinav: No, next. If you visualize the causal relationship between errors as a linked list, the cause of an error is the next node in the linked list.

error.go
abhinav: We control invocations to writeMessage, so an unnecessary thing to guard against.

More generally, the contract is: `%+v` is special, `%v` is the default (per `fmt`), and everything else is undefined. Since you don't want string formatting to error out unnecessarily (you don't want the thing you're using for debugging to fail with a different error), we fall back to `%v` whenever it's not `%+v`.

error.go
abhinav: It's a compile-time check to verify that the interface is satisfied.

If errProvide fails to satisfy the causer interface, this code won't compile.

(We use this convention in a number of places.)

internal/digreflect/func_test.go
r-hang: Ah, okay!

error.go
r-hang: Okay, understood. But do you agree that eventually there should only be one error message?

error.go
r-hang: Just to make sure I have this straight: In the context of 
```
cannot provide function "go.uber.org/dig".TestProvideConstructorErrors\S+ \
```
the cause would be
```
"cannot specify a name for result objects: dig.out embeds dig.Out"
```
As described in ln 1297 of dig_test.go

error.go
r-hang: 👍

error.go
abhinav: Yeah, in "cannot do x: reason y,"  if the current error is "cannot do x", "reason y" is next/cause.

error.go
abhinav: Yeah, they should be similar. If we were writing this error type from scratch, we wouldn't have the two varying error messages.

error.go
r-hang: 👍

PR #262, opened 2020-03-16
Drop dependency on golang.org/x/lint

robbertvanginkel:
We use golang.org/x/lint for dev-time tooling only. We don't
need to declare constraints on versions of these libraries that our
users use.

Refs T5279939


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - GraphViz visualization of the graph now includes names of packages next to
   constructors.
+- Drop library dependency on `golang.org/x/lint`.
```
abhinav:
```suggestion

### Changed
- Drop library dependency on `golang.org/x/lint`.
```

PR #263, opened 2020-03-23
Add support to invoke functions and getting the returned values

fionera:
This is a needed change for the POC on fx

https://github.com/uber-go/fx/pull/699

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -361,21 +361,69 @@ func (c *Container) Provide(constructor interface{}, opts ...ProvideOption) erro
 // The function may return an error to indicate failure. The error will be
 // returned to the caller as-is.
 func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
-	ftype := reflect.TypeOf(function)
-	if ftype == nil {
-		return errors.New("can't invoke an untyped nil")
-	}
-	if ftype.Kind() != reflect.Func {
-		return errf("can't invoke non-function %v (type %v)", function, ftype)
-	}
-
-	pl, err := newParamList(ftype)
+	returned, err := c.invoke(function, opts...)
 	if err != nil {
 		return err
 	}
 
+	if len(returned) == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


terorie:
Any update/required changes on this? The PR looks finished and it would be great to have upstream.

abhinav:
> Any update/required changes on this? The PR looks finished and it would be great to have upstream.

Apologies for the delay in responding.

Unfortunately, we do not yet have plans to support a feature like this in Fx.
Late provides make the Fx dependency graph more dynamic.
That is, dependency graph, which was previously static and fully known when `fx.New` was called,
is now dynamic and arbitrary user-provided code must run before the rest of the graph is visible.

We've wrestled with this idea before, discussing it even during the initial release of Fx.
(In our discussions, we referred to this feature as "dynamic graphs".)

After having a few years of experience running Fx in production,
we're of the opinion that a dynamic graph greatly increases the risk of inscrutable application startup errors.
Today, two of the most common start-up issues we see in our support channels are:

- dependency not specified
- attempt to make an RPC request in a constructor (as opposed to a lifecycle hook) caused a time out

These issues will be greatly amplified if we add support for arbitrary function execution before the graph is even fully visible.

We're open to reconsidering this in the future, but it's not the highest priority feature on our list right now. We have work in flight to significantly alter a number of Fx's internals, so we'd like to focus on that before we consider another complex feature.

However, when we do discuss this feature, we'd probably want the design to account for the following:

We'll want to fully understand how lifecycle hooks will interact with the feature. Currently at startup, Fx effectively has these two phases: invoke everything, run all startup hooks. With Late Provides, the invoke everything stage will now do a lot more. We have no means of guarding against RPC calls being made there so we'd want to make a good choice of how those should interact with lifecycle hooks. Should lifecycle hooks be implemented in terms of Late Provides?

Further, we'd want a deliberate focus on error messaging for such an API so that it is obvious to users at startup why their application failed.

All that said, since we won't be prioritizing this feature in the near future, I'm going to elect to close this PR. I will replicate this comment in #698 for when we come back to this feature.

PR #264, opened 2020-03-24
Add support for flattening value group providers

robbertvanginkel:
Resolves https://github.com/uber-go/fx/issues/693.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -692,6 +692,36 @@ func isFieldOptional(f reflect.StructField) (bool, error) {
 	return optional, err
 }
 
+type group struct {
+	name    string
+	flatten bool
+}
+
+func parseGroup(f reflect.StructField) (group, error) {
+	tag := f.Tag.Get(_groupTag)
+	if tag == "" {
+		return group{}, nil
+	}
+
+	components := strings.Split(tag, ",")
+	var g group
+	for i, c := range components {
+		if i == 0 {
+			g.name = c
+			continue
+		}
+		switch c {
+		case "flatten":
+			g.flatten = true
+		default:
+			return group{}, errf(
+				"invalid value %q for %q tag on field %v",
+				c, _groupTag, f.Name)
+		}
+	}
+	return g, nil
+}
+
 // Checks that all direct dependencies of the provided param are present in
 // the container. Returns an error if not.
 func shallowCheckDependencies(c containerStore, p param) error {
--- a/dig.go
+++ b/dig.go
@@ -37,7 +37,6 @@ import (
 const (
 	_optionalTag = "optional"
 	_nameTag     = "name"
-	_groupTag    = "group"
 )
 
 // Unique identification of an object in the graph.
@@ -690,36 +689,6 @@ func isFieldOptional(f reflect.StructField) (bool, error) {
 	}
 
 	return optional, err
-}
-
-type group struct {
-	name    string
-	flatten bool
-}
-
-func parseGroup(f reflect.StructField) (group, error) {
-	tag := f.Tag.Get(_groupTag)
-	if tag == "" {
-		return group{}, nil
-	}
-
-	components := strings.Split(tag, ",")
-	var g group
-	for i, c := range components {
-		if i == 0 {
-			g.name = c
-			continue
-		}
-		switch c {
-		case "flatten":
-			g.flatten = true
-		default:
-			return group{}, errf(
-				"invalid value %q for %q tag on field %v",
-				c, _groupTag, f.Name)
-		}
-	}
-	return g, nil
 }
 
 // Checks that all direct dependencies of the provided param are present in
--- a/dig_test.go
+++ b/dig_test.go
@@ -2727,47 +2727,6 @@ func TestFailingFunctionDoesNotCreateInvalidState(t *testing.T) {
 	}), "second invoke must fail")
 }
 
-func TestParseGroup(t *testing.T) {
-	tests := []struct {
-		name    string
-		args    reflect.StructField
-		wantG   group
-		wantErr string
-	}{
-		{
-			name:  "no group",
-			args:  reflect.StructField{},
-			wantG: group{},
-		},
-		{
-			name:  "simple group",
-			args:  reflect.StructField{Tag: `group:"somegroup"`},
-			wantG: group{name: "somegroup"},
-		},
-		{
-			name:  "flattened group",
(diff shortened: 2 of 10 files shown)


dig.go
```diff
@@ -692,6 +692,36 @@ func isFieldOptional(f reflect.StructField) (bool, error) {
 	return optional, err
 }
 
+type group struct {
```
abhinav: nit: Placing this in dig.go doesn't seem right. I would prefer to place it
closer to where it's used or in its own place. Suggestions:

- Place in result.go because it's meant for use by resultGrouped. That it's
  also used in param.go can be ignored because that's only for validation.
- Place in a new group.go file, in which case we'll want to move the _groupTag
  const definition there too.

dig.go
```diff
@@ -692,6 +692,36 @@ func isFieldOptional(f reflect.StructField) (bool, error) {
...
+	name    string
+	flatten bool
```
abhinav: Consider exporting fields and methods of unexported types that are meant for
use by other unexported types. This documents the "public" portion of the
private API.

```suggestion
	Name    string
	Flatten bool
```

dig.go
```diff
@@ -692,6 +692,36 @@ func isFieldOptional(f reflect.StructField) (bool, error) {
...
+}
+
+func parseGroup(f reflect.StructField) (group, error) {
```
abhinav: parseGroupTag?

dig.go
```diff
@@ -692,6 +692,36 @@ func isFieldOptional(f reflect.StructField) (bool, error) {
...
+	tag := f.Tag.Get(_groupTag)
+	if tag == "" {
+		return group{}, nil
```
abhinav: Looking at its usage, there's no valid scenario for parseGroup where the tag
isn't set. Should we panic since that's a bug?

dig.go
```diff
@@ -692,6 +692,36 @@ func isFieldOptional(f reflect.StructField) (bool, error) {
...
+	}
+
+	components := strings.Split(tag, ",")
+	var g group
+	for i, c := range components {
+		if i == 0 {
+			g.name = c
+			continue
+		}
```
abhinav: string.Split will always return a non-empty list so we can simplify this a bit:

```suggestion
	g := group{name: components[0]}
	for _, c := range components[1:] {
```

dig.go
```diff
@@ -692,6 +692,36 @@ func isFieldOptional(f reflect.StructField) (bool, error) {
...
+		switch c {
+		case "flatten":
+			g.flatten = true
+		default:
+			return group{}, errf(
+				"invalid value %q for %q tag on field %v",
```
abhinav: nit: s/value/option since everything after the group name is an option.

```suggestion
				"invalid option %q for %q tag on field %v",
```

dig_test.go
```diff
@@ -2698,6 +2727,47 @@ func TestFailingFunctionDoesNotCreateInvalidState(t *testing.T) {
...
+			wantG:   group{},
+			wantErr: `invalid value "abc" for "group" tag on field`,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotG, err := parseGroup(tt.args)
+			if tt.wantErr != "" {
+				assert.Error(t, err)
+				assert.Contains(t, err.Error(), tt.wantErr)
```
abhinav: For error case, should return after checking error because group value is irrelevant in error case.

```suggestion
				assert.Contains(t, err.Error(), tt.wantErr)
				return
```

dig_test.go
```diff
@@ -2698,6 +2727,47 @@ func TestFailingFunctionDoesNotCreateInvalidState(t *testing.T) {
...
+		},
+		{
+			name:  "flattened group",
+			args:  reflect.StructField{Tag: `group:"somegroup,flatten"`},
+			wantG: group{name: "somegroup", flatten: true},
+		},
+		{
+			name:    "error",
+			args:    reflect.StructField{Tag: `group:"somegroup,abc"`},
+			wantG:   group{},
```
abhinav: optional: can drop field for zero value
```suggestion
```

result.go
```diff
@@ -371,6 +371,11 @@ type resultGrouped struct {
 
 	// Type of value produced.
 	Type reflect.Type
+
+	// Indicates elements of a value are to be injected individually, instead of
+	// as a group. Requires the value's slice to be a group. If set, Type will be
+	// the type of individual elements rather than the group.
+	flatten bool
```
abhinav: export please
```suggestion
	Flatten bool
```

result.go
```diff
@@ -386,22 +391,40 @@ func (rt resultGrouped) DotResult() []*dot.Result {
 
 // newResultGrouped(f) builds a new resultGrouped from the provided field.
 func newResultGrouped(f reflect.StructField) (resultGrouped, error) {
-	rg := resultGrouped{Group: f.Tag.Get(_groupTag), Type: f.Type}
-
+	g, err := parseGroup(f)
+	if err != nil {
+		return resultGrouped{}, err
+	}
+	rg := resultGrouped{Group: g.name}
```
abhinav: optional: can simplify logic below with:

```suggestion
	rg := resultGrouped{
		Group: g.name,
		Type: f.Type,
		flatten: g.flatten,
	}
```

And then after the validation `switch`:

```
if rg.flatten {
	rg.Type = rg.Type.Elem()
}
```

result.go
```diff
@@ -386,22 +391,40 @@ func (rt resultGrouped) DotResult() []*dot.Result {
...
 			"name:%q provided with group:%q", name, rg.Group)
 	case optional:
 		return rg, errors.New("value groups cannot be optional")
 	}
+	if g.flatten {
+		rg.Type = f.Type.Elem()
+		rg.flatten = true
+	} else {
+		rg.Type = f.Type
+	}
```
abhinav: Per comment above:

```suggestion
	if rg.flatten {
		rg.Type = rg.Type.Elem()
	}
```

result.go
```diff
@@ -386,22 +391,40 @@ func (rt resultGrouped) DotResult() []*dot.Result {
...
 
 	return rg, nil
 }
 
 func (rt resultGrouped) Extract(cw containerWriter, v reflect.Value) {
+	if rt.flatten {
```
abhinav: nit: invert if to reduce nesting

```
if !rt.flatten {
  cw.submitGroupedValue(..)
  return
}

for ... {
  cw.submitGroupedValue(..)
}
```

PR #266, opened 2020-05-29
Introduce DryRun option which disables invocation of functions supplied to Provide and Invoke.

shirchen:
Introduce `DryRun` Option which, when set to true, disables invocation of functions
 supplied to Provide and Invoke. This option will be used to build no-op containers, 
for example for `fx.ValidateApp` method.

Needed by https://github.com/uber-go/fx/pull/706

Ref T3706201

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -157,8 +157,8 @@ type Container struct {
 	// Defer acyclic check on provide until Invoke.
 	deferAcyclicVerification bool
 
-	// Whether container is being run in dry mode used for testing.
-	dry bool
+	// Invoker calls a function with arguments provided to Provide or Invoke.
+	invoker Invoker
 }
 
 // containerWriter provides write access to the Container's underlying data
@@ -199,8 +199,8 @@ type containerStore interface {
 
 	createGraph() *dot.Graph
 
-	// Returns whether container is running in dry mode.
-	Dry() bool
+	// Returns Invoker function to use when calling arguments.
+	Invoker() Invoker
 }
 
 // provider encapsulates a user-provided constructor.
@@ -234,7 +234,7 @@ func New(opts ...Option) *Container {
 		values:    make(map[key]reflect.Value),
 		groups:    make(map[key][]reflect.Value),
 		rand:      rand.New(rand.NewSource(time.Now().UnixNano())),
-		dry:       false,
+		invoker:   defaultInvoker,
 	}
 
 	for _, opt := range opts {
@@ -265,12 +265,30 @@ func setRand(r *rand.Rand) Option {
 	})
 }
 
-// Dry is an Option when set to true, overrides default behavior of calling functions supplied by
-// Provide and Invoke and allows for validation of graph dependencies.
-func Dry(dry bool) Option {
+// DryRun is an Option which, when set to true, disables invocation of functions supplied to
+// Provide and Invoke. Use this to build no-op containers.
+func DryRun(dry bool) Option {
 	return optionFunc(func(c *Container) {
-		c.dry = dry
+		if dry {
+			c.invoker = dryInvoker
+		}
 	})
+}
+
+// Invoker exports a function to call when invoking a function provided to Provide or Invoke.
+type Invoker func(fn reflect.Value, args []reflect.Value) (results []reflect.Value)
+
+func defaultInvoker(fn reflect.Value, args []reflect.Value) []reflect.Value {
+	return fn.Call(args)
+}
+
+func dryInvoker(fn reflect.Value, _ []reflect.Value) []reflect.Value {
+	results := make([]reflect.Value, fn.Type().NumOut())
+	for i := 0; i < fn.Type().NumOut(); i++ {
+		results[i] = reflect.Zero(fn.Type().Out(i))
+	}
+
+	return results
 }
 
 func (c *Container) knownTypes() []reflect.Type {
@@ -324,9 +342,10 @@ func (c *Container) getProviders(k key) []provider {
 	return providers
 }
 
-// Dry returns whether container is set in dry mode and does not call Provide nor Invoke functions.
-func (c *Container) Dry() bool {
-	return c.dry
+// Invoker return a function to run when calling function provided to Provide or Invoke. Used for
+// running container in dry mode.
+func (c *Container) Invoker() Invoker {
+	return c.invoker
 }
 
 // Provide teaches the container how to build values of one or more types and
@@ -413,18 +432,16 @@ func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
 			Reason: err,
 		}
 	}
-	// Do not invoke if we are testing graph dependencies.
-	if !c.dry {
-		returned := reflect.ValueOf(function).Call(args)
-		if len(returned) == 0 {
-			return nil
-		}
-		if last := returned[len(returned)-1]; isError(last.Type()) {
-			if err, _ := last.Interface().(error); err != nil {
-				return err
-			}
+	returned := c.invoker(reflect.ValueOf(function), args)
+	if len(returned) == 0 {
+		return nil
+	}
+	if last := returned[len(returned)-1]; isError(last.Type()) {
+		if err, _ := last.Interface().(error); err != nil {
+			return err
 		}
 	}
+
 	return nil
 }
 
@@ -686,15 +703,13 @@ func (n *node) Call(c containerStore) error {
 	}
 
 	receiver := newStagingContainerWriter()
-	// Do not call providers if we are in dry mode and  dont want anything to run.
-	if !c.Dry() {
-		results := reflect.ValueOf(n.ctor).Call(args)
-		if err := n.resultList.ExtractList(receiver, results); err != nil {
-			return errConstructorFailed{Func: n.location, Reason: err}
-		}
+	results := c.Invoker()(reflect.ValueOf(n.ctor), args)
+	if err := n.resultList.ExtractList(receiver, results); err != nil {
+		return errConstructorFailed{Func: n.location, Reason: err}
 	}
 	receiver.Commit(c)
 	n.called = true
+
 	return nil
 }
 
--- a/dig.go
+++ b/dig.go
@@ -156,6 +156,9 @@ type Container struct {
 
 	// Defer acyclic check on provide until Invoke.
 	deferAcyclicVerification bool
+
+	// Whether container is being run in dry mode used for testing.
+	dry bool
 }
 
 // containerWriter provides write access to the Container's underlying data
@@ -228,6 +231,7 @@ func New(opts ...Option) *Container {
 		values:    make(map[key]reflect.Value),
 		groups:    make(map[key][]reflect.Value),
 		rand:      rand.New(rand.NewSource(time.Now().UnixNano())),
+		dry:       false,
 	}
 
 	for _, opt := range opts {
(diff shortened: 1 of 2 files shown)


dig.go
```diff
@@ -258,6 +262,13 @@ func setRand(r *rand.Rand) Option {
 	})
 }
 
+// Dry sets container mode to dry used for testing graphs without invocation.
+func Dry(dry bool) Option {
```
SyrieBianco, line 266: 1) Are we attached to the name `Dry` for this option? I feel like `DryRun` or even `ValidateGraphOnly` might be a more descriptive name. 
2) Nice; I like accepting the bool argument here. I could see an argument for having the API be just whether or not you provide the `Dry` option, but I like the ability to pass in, say, environment variables as arguments.

dig.go
```diff
@@ -228,6 +231,7 @@ func New(opts ...Option) *Container {
 		values:    make(map[key]reflect.Value),
 		groups:    make(map[key][]reflect.Value),
 		rand:      rand.New(rand.NewSource(time.Now().UnixNano())),
+		dry:       false,
```
SyrieBianco, line 234: Nit, but i'm not sure it adds much to explicitly set `dry` to false on instantiation.

dig.go
```diff
@@ -258,6 +262,13 @@ func setRand(r *rand.Rand) Option {
 	})
 }
 
+// Dry sets container mode to dry used for testing graphs without invocation.
```
SyrieBianco, line 265: Since this is the exported API, I would make this comment a little more detailed. Maybe something like:
```
// Dry provides a "dry run" testing option that, if set to true, will
// verify the graph without invocation. 
```

dig_test.go
```diff
@@ -1642,6 +1642,34 @@ func TestProvideCycleFails(t *testing.T) {
 		)
 	})
 
+	t.Run("parameters only on dry run", func(t *testing.T) {
```
SyrieBianco, line 1645: So, as it is, this test seems only to be confirming that a Dry container will error on introducing a cycle, but isn't actively asserting that the functions aren't invoked.

A naïve soltuion might be something like:
```
var invocations int

newA := func(*C) *A {
    invocations ++
    return &A{}
}
```

dig_test.go
```diff
@@ -1689,6 +1717,53 @@ func TestProvideCycleFails(t *testing.T) {
 		)
 	})
 
+	t.Run("dig.In based cycle dry run", func(t *testing.T) {
```
SyrieBianco, line 1720: Apply my comments from L1645 to the other test types as well :)

dig.go
```diff
@@ -393,14 +404,16 @@ func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
...
-
-	returned := reflect.ValueOf(function).Call(args)
-	if len(returned) == 0 {
-		return nil
-	}
-	if last := returned[len(returned)-1]; isError(last.Type()) {
-		if err, _ := last.Interface().(error); err != nil {
-			return err
+	// Do not invoke if we are testing graph dependencies.
+	if !c.dry {
```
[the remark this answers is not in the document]
shirchen: Sure, I can go either way though: to me, having 2 `return nil` statements looks awkward. Also, branching on the specific part makes it clear what we want to skip. Imagine someone adds more logic that dry may want to run through. Then you need to redo the branch. 

```
	// Do not invoke if we are testing graph dependencies.
	if c.dry {
		return nil
	}
        ...
	return nil
}
```

In any case, I dont feel strongly and laying out my reasoning.

dig.go
shirchen: Dry is a pretty common option https://en.wikipedia.org/wiki/Dry_run_(testing) and is present for most of git commands, for example, with my understanding that it signals that nothing is going to run and we are testing. And yeah, usually it's a boolean.

dig.go
shirchen: Fair, comment is a bit clumsy.

dig_test.go
shirchen: It's a good idea, but unfortunately in these tests, since we are asserting for an error, we don't get to part where things are invoked, so even if we aren't in Dry mode, we will never call these functions, so invocations will always be 0.

I should create a test case where Dry is true and we don't error, and we can use your suggestion.

shirchen: Added a test when a panic within Provide does not run (or test would fail).

dig.go
SyrieBianco: Right; wikipedia has it as "dry run" like I suggested. it's not a huge deal, but I think `DryRun(true)` is clearer than `Dry(true)` for the api. 

And for the second comment, I meant more that in my CLI experience, the API can either be an explicit boolean — ` --dry-run=true` — or implicit just by adding the flag ` --dry-run`. I guess I'm saying that here, the `dig.Option` could be `Dry()` without accepting the bool argument, or `Dry(true/false)` .  I was just commenting that I agree with the choice to explicitly pass the value.

dig.go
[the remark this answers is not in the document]
SyrieBianco: 👍 agreed, it's very minor. It's my habit to short circuit, but I agree that here it makes it a little more readable/refactorable in terms of what's getting skipped

dig.go
```diff
@@ -664,9 +685,12 @@ func (n *node) Call(c containerStore) error {
 	}
 
 	receiver := newStagingContainerWriter()
-	results := reflect.ValueOf(n.ctor).Call(args)
-	if err := n.resultList.ExtractList(receiver, results); err != nil {
-		return errConstructorFailed{Func: n.location, Reason: err}
+	// Do not call providers if we are in dry mode and  dont want anything to run.
+	if !c.Dry() {
```
SyrieBianco, line 689: Oh, nice, this is exactly what I had messaged you offline about — that provides would still get executed — but this addition already addresses that.

dig.go
```diff
@@ -664,9 +685,12 @@ func (n *node) Call(c containerStore) error {
...
+		results := reflect.ValueOf(n.ctor).Call(args)
+		if err := n.resultList.ExtractList(receiver, results); err != nil {
```
SyrieBianco, line 691: So I'm not sure about this, but is it ok to skip this `ExtractList` call? It seems to me that this is where the results from the provide are registered with the container; if you don't register the return types with the container, will an invoke that depends on those types succeed? 

I'm going to investigate in the code as a sanity check, but I feel like the easiest way to confirm would be to add a test case that confirms that in 1) invokes (as well as provides) are not executed and 2) that invokes that depend on return types from provides don't fail in `Dry` mode.

dig_test.go
```diff
@@ -1611,6 +1611,20 @@ func TestProvideKnownTypesFails(t *testing.T) {
 	})
 }
 
+func TestDrySucceedsOnPanickingFn(t *testing.T) {
+	t.Run("dry mode should succeed on a function that panics if run", func(t *testing.T) {
+		type type1 struct{}
+		panics := func() *type1 {
+			t.Fatal("must not be called")
```
SyrieBianco, line 1618: ty!

dig.go
SyrieBianco: Just reviewed your [Fx PR](https://github.com/uber-go/fx/pull/706/), and you've got these test cases transitively covered by those assertions there, so it looks like this won't be a problem. That said, it probably still makes sense to add a parallel test here, or an assertion to `TestDrySucceedsOnPanickingFn` that the invoke also doesn't get run.

dig.go
```diff
@@ -309,6 +342,12 @@ func (c *Container) getProviders(k key) []provider {
 	return providers
 }
 
+// Invoker return a function to run when calling function provided to Provide or Invoke. Used for
+// running container in dry mode.
+func (c *Container) Invoker() Invoker {
```
abhinav, line 347: should unexport. can rename field to `invokeFn` to avoid conflict with method name.

dig_test.go
```diff
@@ -1611,7 +1611,43 @@ func TestProvideKnownTypesFails(t *testing.T) {
...
 func TestProvideCycleFails(t *testing.T) {
+	t.Run("not dry", func(t *testing.T) {
+		testProvideCycleFails(t, false /* dry run */)
+	})
+	t.Run("dry", func(t *testing.T) {
+		testProvideCycleFails(t, true /* dry run */)
+	})
+}
+
+func testProvideCycleFails(t *testing.T, dryRun bool) {
```
abhinav, line 1650: :+1::+1::+1:

dig.go
```diff
@@ -258,6 +265,32 @@ func setRand(r *rand.Rand) Option {
 	})
 }
 
+// DryRun is an Option which, when set to true, disables invocation of functions supplied to
+// Provide and Invoke. Use this to build no-op containers.
+func DryRun(dry bool) Option {
+	return optionFunc(func(c *Container) {
+		if dry {
+			c.invoker = dryInvoker
+		}
```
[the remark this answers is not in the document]
shirchen: I like having this work done in the Option to contain what DryRun means.

dig.go
```diff
@@ -258,6 +265,35 @@ func setRand(r *rand.Rand) Option {
...
+
+func defaultInvoker(fn reflect.Value, args []reflect.Value) []reflect.Value {
+	return fn.Call(args)
+}
+
+// Generates zero values for results without calling the supplied function.
+func dryInvoker(fn reflect.Value, _ []reflect.Value) []reflect.Value {
+	results := make([]reflect.Value, fn.Type().NumOut())
+	for i := 0; i < fn.Type().NumOut(); i++ {
+		results[i] = reflect.Zero(fn.Type().Out(i))
```
abhinav, line 291: minor: extract fn.Type() into a local variable.

```suggestion
	ft := fn.Type()
	results := make([]reflect.Value, ft.NumOut())
	for i := 0; i < ft.NumOut(); i++ {
		results[i] = reflect.Zero(ft.Out(i))
```

dig_test.go
```diff
@@ -1611,7 +1611,43 @@ func TestProvideKnownTypesFails(t *testing.T) {
 	})
 }
 
+func TestDryModeSuccess(t *testing.T) {
+	t.Run("does not call provides", func(t *testing.T) {
+		type type1 struct{}
+		panics := func() *type1 {
```
abhinav, line 1617:
```suggestion
		provides := func() *type1 {
```

dig_test.go
```diff
@@ -1611,7 +1611,43 @@ func TestProvideKnownTypesFails(t *testing.T) {
...
+			t.Fatal("must not be called")
+			return &type1{}
+		}
+		invokes := func(*type1) {}
+		c := New(DryRun(true))
+		assert.NoError(t, c.Provide(panics))
```
abhinav, line 1623:
```suggestion
		assert.NoError(t, c.Provide(provides))
```

dig_test.go
```diff
@@ -1611,7 +1611,43 @@ func TestProvideKnownTypesFails(t *testing.T) {
...
+		assert.NoError(t, c.Invoke(invokes))
+	})
+	t.Run("does not call invokes", func(t *testing.T) {
+		type type1 struct{}
+		panics := func() *type1 {
```
abhinav, line 1628:
```suggestion
		provides := func() *type1 {
```

dig_test.go
```diff
@@ -1611,7 +1611,43 @@ func TestProvideKnownTypesFails(t *testing.T) {
...
+			t.Fatal("must not be called")
+			return &type1{}
+		}
+		invokes := func(*type1) {
+			t.Fatal("must not be called")
+		}
+		c := New(DryRun(true))
+		assert.NoError(t, c.Provide(panics))
```
abhinav, line 1636:
```suggestion
		assert.NoError(t, c.Provide(provides))
```

PR #270, opened 2020-10-19
Update Makefile

socialshout:
Minor Changes


abhinav (changes requested):
`./.....` is not actually recognized by `go list`.

Issue #273, opened 2021-02-16
dig.In/Out: Allow and ignore unexported fields


abhinav:
dig.In and dig.Out structs explicitly disallow unexported fields on their
structs.

https://github.com/uber-go/dig/blob/4ea249947c78543420fc8c36620326fb811a0094/param.go#L349

Based on anecdotal evidence from our internal support room, this check makes
sense because otherwise users will assume that those unexported fields will
also be filled by Fx.

However, there's value in allowing unexported fields on dig.In tagged structs
regardless, e.g. https://github.com/uber-go/fx/issues/673. We may want to
support an opt-in means of specifying that a provided struct MAY have
unexported fields.

One means of doing this is struct tags on the embedded `dig.In` field:

    type foo struct {
      dig.In `allow-unexported:"true"`  // naming TBD
      
      Foo Foo
      bar bar // ignored by Dig
    }

Internal issue reference: GO-374

PR #274, opened 2021-02-23
Add ability to allow unexported fields in dig.In structs

manjari25:
In order for unexported fields to be allowed, an 
"allowUnexported=true" struct tag should be used on the embedded dig.In 
field.

Resolves: https://github.com/uber-go/dig/issues/273

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -35,8 +35,9 @@ import (
 )
 
 const (
-	_optionalTag = "optional"
-	_nameTag     = "name"
+	_optionalTag        = "optional"
+	_nameTag            = "name"
+	_allowUnexportedTag = "allowUnexported"
 )
 
 // Unique identification of an object in the graph.
@@ -733,6 +734,24 @@ func isFieldOptional(f reflect.StructField) (bool, error) {
 	}
 
 	return optional, err
+}
+
+// Checks if unexported fields in an In struct are allowed. The struct field
+// MUST be an _inType.
+func isUnexportedFieldAllowed(f reflect.StructField) (bool, error) {
+	tag := f.Tag.Get(_allowUnexportedTag)
+	if tag == "" {
+		return false, nil
+	}
+
+	allowed, err := strconv.ParseBool(tag)
+	if err != nil {
+		err = errf(
+			"invalid value %q for %q tag on field %v",
+			tag, _allowUnexportedTag, f.Name, err)
+	}
+
+	return allowed, err
 }
 
 // Checks that all direct dependencies of the provided param are present in
--- a/dig.go
+++ b/dig.go
@@ -35,9 +35,9 @@ import (
 )
 
 const (
-	_optionalTag        = "optional"
-	_nameTag            = "name"
-	_allowUnexportedTag = "allowUnexported"
+	_optionalTag         = "optional"
+	_nameTag             = "name"
+	_ignoreUnexportedTag = "ignore-unexported"
 )
 
 // Unique identification of an object in the graph.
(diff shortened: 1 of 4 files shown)


dig.go
```diff
@@ -35,8 +35,9 @@ import (
 )
 
 const (
-	_optionalTag = "optional"
-	_nameTag     = "name"
+	_optionalTag        = "optional"
+	_nameTag            = "name"
+	_allowUnexportedTag = "allowUnexported"
```
abhinav, line 40: I'm not 100% sold on this tag name.

1. I'm not sure about just "unexported," "allow unexported," or "ignore unexported"
2. I don't think we use camelCase for struct tags usually. [go-flags](https://pkg.go.dev/github.com/jessevdk/go-flags) uses kebab-case so "allow-unexported" or "ignore-unexported" might work. Just "unexported" obviously sidesteps this issue entirely.

I'm personally leaning towards either of the following but I welcome more suggestions.

- `ignore-unexported:"true"`, or
- `unexported:"ignore"` defaulting to `unexported:"error"` (this possibly allows us to add a different behavior for unexported fields in the future and use that name here, e.g. `unexported:"warn"` but I'm just spitballing)

param.go
```diff
@@ -287,13 +287,30 @@ func (po paramObject) DotParam() []*dot.Param {
 func newParamObject(t reflect.Type) (paramObject, error) {
 	po := paramObject{Type: t}
 
+	// Check if the In type supports unexported fields.
+	var allowUnexported bool
+	var err error
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		if f.Type == _inType {
+			allowUnexported, err = isUnexportedFieldAllowed(f)
```
abhinav, line 296: We can reduce the scope of the `err` to where it's first used:

```suggestion
	for i := 0; i < t.NumField(); i++ {
		f := t.Field(i)
		if f.Type == _inType {
			var err error
			allowUnexported, err = isUnexportedFieldAllowed(f)
```

dig.go
manjari25: Following offline discussion, I've decided to go with `ignore-unexported:"true"`.

dig_test.go
```diff
@@ -344,6 +344,29 @@ func TestEndToEndSuccess(t *testing.T) {
...
+			return &type1{}, &type2{}
+		}
+
+		c := New()
+		type param struct {
+			In `ignore-unexported:"true"`
+
+			T1 *type1 // regular 'ol type
+			T2 *type2 `optional:"true"` // optional type present in the graph
+			t3 *type3
```
abhinav, line 361: can we have that or a different constructor produce `type3` so we can verify that the unexported field t3 is not set?

param.go
```diff
@@ -287,13 +287,30 @@ func (po paramObject) DotParam() []*dot.Param {
 func newParamObject(t reflect.Type) (paramObject, error) {
 	po := paramObject{Type: t}
 
+	// Check if the In type supports ignoring unexported fields.
+	var ignoreUnexported bool
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		if f.Type == _inType {
+			var err error
+			ignoreUnexported, err = isIgnoreUnexportedSet(f)
```
abhinav, line 296: optional: should we invert the if condition?

```
if f.Type != _inType { continue }

var err error
// ...
```

abhinav (commented):
LGTM besides minor thing: can we add type3 to the graph for the `ignore-unexported"true"` case in dig_test.go so that we can verify we never accidentally set the unexported field (even though that's not allowed by reflect).

param.go
manjari25: Hmm, we could. I went with `break` because we can stop processing the fields as soon as we find the `_inType` field . With `continue` we would continue looking through all fields.

abhinav (approved):

abhinav:
Per Slack conversation, this likely won't be enough for Fx because Fx declares In and Out as 

```
type In struct{ dig.In }
type Out struct{ dig.Out }
```

So the struct tag on an fx.In field will never make it to dig.

Since the embed-In is a leftover from before type aliases were supported in Go, we can now switch it to aliases to solve this issue.

```
package fx

type In = dig.In

type Out = dig.Out
```

evgeny-myasishchev:
Any plans to release master? This change is wanted :)

abhinav:
> Any plans to release master? This change is wanted :)

Oops, we missed this. Yes, we intend to release this change as soon as we make the corresponding change in https://github.com/uber-go/fx.

PR #275, opened 2021-05-05
Switching from Travis to GitHub Actions

shirchen:
Due to Travis CI deprecation, switching to GitHub Workflows. 

Ref GO-555


.github/workflows/go.yml
```diff
@@ -0,0 +1,51 @@
...
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+
+    - name: Download Dependencies
+      run: go mod download
+
+    - name: Lint
+      run: make lint
```
abhinav, line 42: Missing a `if: matrix.latest` here that will ensure that we lint only with the latest Go.

```suggestion
      if: matrix.latest
      run: make lint
```

.github/workflows/go.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    - name: Test
+      run: make test
```
abhinav, line 45: This should be `make cover` so that we generate a coverage report to upload to codecov.
```suggestion
      run: make cover
```

abhinav (commented):
Thanks for doing this!

abhinav (approved):

Issue #254, opened 2019-11-22
dig: Provide-time introspection


abhinav:
Currently, when you `c.Provide` a constructor to a dig container, you get no
visibility into what dig got out of it. If the constructor had a dig/fx.In
and/or a dig/fx.Out, the caller would have to implement its own reflection
logic to determine the list of actual types, named values, or value groups,
provided or consumed by that constructor. In fact, Fx has to do so already to
be able to log what was provided to each constructor.

I suggest that we add support for a means to extract information about the
provided constructor. It can take the form of a handle-style API which gives
us access to an object that represents a unique reference to the constructor
in the dig container, or it can take the form of an info-style API which
merely returns information it extracted.

To elaborate what I mean by that:

### Handle-style API

Under this API, we would expose a means of requesting a 'handle' to a
constructor into the container.

    var h dig.Handle
    c.Provide(constructor, dig.ProvideHandle(&h))
    
This handle would expose a means of introspection.

    h.ID()  // unique ID of the constructor in the container.
    h.Inputs()
    h.Outputs()
    
Theoretically, this could also expose a means of deletion and provide access
to the original constructor:

    h.Delete()

### Info-style API

In the Info-style API, we would only provide a means of introspection.

    var info dig.Info
    c.Provide(constructor, dig.ProvideInfo(&info))

    info.ID
    info.Inputs
    info.Outputs

----

Barring names, which are TBD, I don't have a strong preference between the
approaches, especially since nothing prevents us from providing both. Info
would be quicker to implement and exposes a smaller API surface, so perhaps
we should go for that first.

We can eventually provide accessors on errors returned by dig that point to
relevant constructors by their `Info/Handle` IDs.

Ref: T4578111

sywhang:
Closed via #280

PR #280, opened 2021-06-17
Add WithInfo and ConstructorInfo

sywhang:
This adds WithInfo and ConstructorInfo, which allow users to get what Dig was able to extract out of the `Provide`d constructors.

Specifically, ConstructorInfo contains ID of the constructor (which is unique among all provided ctors), Inputs (parameters), and Outputs (results) types as Strings.

With these, users can introspect the constructor info that Dig was able to get out using reflection. For example:
```go
info := ConstructorInfo{}
c.Provide(someCtor, WithInfo(&info))

fmt.Println(info.ID) // Will print the ID of someCtor

for _, param := info.Inputs {
    fmt.Println(param) // will print the input parameters types of someCtor
}
 
for _, res := info.Outputs {
    fmt.Println(res) // will print the return types of someCtor
}
```

Refs GO-169 and #254.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -131,6 +131,10 @@ func Group(group string) ProvideOption {
 
 // ConstructorInfo provides information about the constructor's inputs and outputs
 // types as strings, as well as the ID of the constructor supplied to the Container.
+//
+// ID is a unique integer representing the constructor node in the dependency graph.
+// Inputs is a string slice with the types of the input parameters of the constructor.
+// Outputs is a string slice with the types of the results produced by the constructor.
 type ConstructorInfo struct {
 	ID      int
 	Inputs  []string
--- a/dig.go
+++ b/dig.go
@@ -129,21 +129,40 @@ func Group(group string) ProvideOption {
 	})
 }
 
+// ID is a unique integer representing the constructor node in the dependency graph.
+type ID int
+
 // ConstructorInfo provides information about the constructor's inputs and outputs
 // types as strings, as well as the ID of the constructor supplied to the Container.
-//
-// ID is a unique integer representing the constructor node in the dependency graph.
-// Inputs is a string slice with the types of the input parameters of the constructor.
-// Outputs is a string slice with the types of the results produced by the constructor.
+// It contains ID for the constructor, as well as slices of Input and Output types,
+// which are Stringers that report the types of the parameters and results respectively.
 type ConstructorInfo struct {
-	ID      int
-	Inputs  []string
-	Outputs []string
+	ID      ID
+	Inputs  []*Input
+	Outputs []*Output
 }
 
-// WithInfo is a ProvideOption that writes info on what Dig was able to get out
+// Input is a stringer that report the type of an input parameters of the constructor.
+type Input struct {
+	t reflect.Type
+}
+
+func (i *Input) String() string {
+	return i.t.String()
+}
+
+// Output is a stringer that report the types of an output produced by the constructor.
+type Output struct {
+	t reflect.Type
+}
+
+func (o *Output) String() string {
+	return o.t.String()
+}
+
+// FillInfo is a ProvideOption that writes info on what Dig was able to get out
 // out of the provided constructor into the provided ConstructorInfo.
-func WithInfo(info *ConstructorInfo) ProvideOption {
+func FillInfo(info *ConstructorInfo) ProvideOption {
 	return provideOptionFunc(func(opts *provideOptions) {
 		opts.Info = info
 	})
@@ -522,20 +541,22 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
 	c.nodes = append(c.nodes, n)
 
 	// Record introspection info for caller if Info option is specified
-	if opts.Info != nil {
-		opts.Info.ID = (int)(n.id)
-		opts.Info.Inputs = make([]string, 0)
-		opts.Info.Outputs = make([]string, 0)
+	if info := opts.Info; info != nil {
+		params := n.ParamList().DotParam()
+		results := n.ResultList().DotResult()
 
-		for _, res := range n.ResultList().DotResult() {
-			opts.Info.Outputs = append(opts.Info.Outputs, res.Type.String())
+		info.ID = (ID)(n.id)
+		info.Inputs = make([]*Input, len(params))
+		info.Outputs = make([]*Output, len(results))
+
+		for i, param := range params {
+			info.Inputs[i] = &Input{t: param.Type}
 		}
 
-		for _, param := range n.ParamList().DotParam() {
-			opts.Info.Inputs = append(opts.Info.Inputs, param.Type.String())
+		for i, res := range results {
+			info.Outputs[i] = &Output{t: res.Type}
 		}
 	}
-
 	return nil
 }
 
--- a/dig.go
+++ b/dig.go
@@ -527,11 +527,11 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
 		opts.Info.Inputs = make([]string, 0)
 		opts.Info.Outputs = make([]string, 0)
 
-		for _, res := range n.resultList.DotResult() {
+		for _, res := range n.ResultList().DotResult() {
 			opts.Info.Outputs = append(opts.Info.Outputs, res.Type.String())
 		}
 
-		for _, param := range n.paramList.DotParam() {
+		for _, param := range n.ParamList().DotParam() {
 			opts.Info.Inputs = append(opts.Info.Inputs, param.Type.String())
 		}
 	}
--- a/dig.go
+++ b/dig.go
@@ -65,7 +65,6 @@ type provideOptions struct {
 	Info  *ConstructorInfo
 }
 
-
 func (o *provideOptions) Validate() error {
 	if len(o.Group) > 0 && len(o.Name) > 0 {
 		return errf(
@@ -130,7 +129,8 @@ func Group(group string) ProvideOption {
 	})
 }
 
-func Info(info *ConstructorInfo) ProvideOption {
+// WithInfo is a ProvideOption that lets Info.
+func WithInfo(info *ConstructorInfo) ProvideOption {
 	return provideOptionFunc(func(opts *provideOptions) {
 		opts.Info = info
 	})
@@ -506,11 +506,22 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
 		}
 		c.isVerifiedAcyclic = true
 	}
-
-	opts.Info.ID = int(n.ID())
-	opts.Info.Inputs = n.ParamList()
-	opts.Info.Outputs = n.ResultList()
 	c.nodes = append(c.nodes, n)
+
+	// Record introspection info for caller if Info option is specified
+	if opts.Info != nil {
+		opts.Info.ID = (int)(n.id)
+		opts.Info.Inputs = make([]string, 0)
+		opts.Info.Outputs = make([]string, 0)
+
+		for _, res := range n.resultList.DotResult() {
+			opts.Info.Outputs = append(opts.Info.Outputs, res.Type.String())
+		}
+
+		for _, param := range n.paramList.DotParam() {
+			opts.Info.Inputs = append(opts.Info.Inputs, param.Type.String())
+		}
+	}
 
 	return nil
 }
--- a/dig.go
+++ b/dig.go
@@ -144,7 +144,9 @@ type ConstructorInfo struct {
 
 // Input is a stringer that report the type of an input parameters of the constructor.
 type Input struct {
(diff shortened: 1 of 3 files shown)


dig.go
```diff
@@ -498,9 +519,23 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
...
-
 	c.nodes = append(c.nodes, n)
 
+	// Record introspection info for caller if Info option is specified
+	if opts.Info != nil {
+		opts.Info.ID = (int)(n.id)
+		opts.Info.Inputs = make([]string, 0)
+		opts.Info.Outputs = make([]string, 0)
+
+		for _, res := range n.resultList.DotResult() {
```
shirchen, line 530: nit: `n.ResultList()`?

dig.go
```diff
@@ -498,9 +519,23 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
...
+			opts.Info.Outputs = append(opts.Info.Outputs, res.Type.String())
+		}
+
+		for _, param := range n.paramList.DotParam() {
```
shirchen, line 534: nit: `n.ParamList()`

dig.go
```diff
@@ -128,6 +129,26 @@ func Group(group string) ProvideOption {
...
 
+// ConstructorInfo provides information about the constructor's inputs and outputs
+// types as strings, as well as the ID of the constructor supplied to the Container.
+//
+// ID is a unique integer representing the constructor node in the dependency graph.
+// Inputs is a string slice with the types of the input parameters of the constructor.
+// Outputs is a string slice with the types of the results produced by the constructor.
+type ConstructorInfo struct {
+	ID      int
+	Inputs  []string
```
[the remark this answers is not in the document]
sywhang: Hmm.. That's a good point. The main rationale behind exposing this as strings was the usage in `fx` but I can see keeping backwards compatibility becoming a mess if we want to expose more things in the future. 

What if I add a `ConstructorInfoOption`s and expose that as a parameter to `NewConstructorInfo`?

```go
func NewConstructorInfo(opts ...ConstructorInfoOption) ConstructorInfo
```

That way in the future if we want to expose not only the names of the types of the parameters/returns but the type value themselves, or `Group`/`Name`, user can construct the ConstructorInfo with whatever options they want to get out from `WithInfo`:

```go
var name string
var group string
var paramType []reflect.Type

info := NewConstructorInfo(WithNameInfo(&name), WithGroupInfo(&group), WithParamTypeInfo(&paramType))
c.Provide(ctor, WithInfo(&info))
```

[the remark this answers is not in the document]
abhinav: Yeah, I'm reluctant to expose just the strings as well. What do you think about opaque `Input` and `Output` types that happen to implement `fmt.Stringer` (`String() string`), reporting the current strings, but leave us room for exposing more information in the future?

```
type ConstructorInfo struct {
  ID int
  Inputs []*Input
  Outputs []*Output
}

type Input struct{ /* contains unexported fields */ }

func (*Input) String() string
```

We might even make `ConstructorInfo` completely opaque for now and add `Inputs() []*Input` and `Outputs() []*Output` methods if we're unsure, but that one is easier to mess with already since it's the "entry" point here.

dig.go
```diff
@@ -128,6 +129,26 @@ func Group(group string) ProvideOption {
...
 }
 
+// ConstructorInfo provides information about the constructor's inputs and outputs
+// types as strings, as well as the ID of the constructor supplied to the Container.
+//
+// ID is a unique integer representing the constructor node in the dependency graph.
+// Inputs is a string slice with the types of the input parameters of the constructor.
+// Outputs is a string slice with the types of the results produced by the constructor.
+type ConstructorInfo struct {
+	ID      int
```
abhinav, line 139: I would make `ID` a new type so that it's distinct from `int`s.

```
type ID int

type ConstructorInfo struct {
  ID ID
  // ...
}
```

dig.go
```diff
@@ -128,6 +129,26 @@ func Group(group string) ProvideOption {
...
+	Inputs  []string
+	Outputs []string
+}
+
+// WithInfo is a ProvideOption that writes info on what Dig was able to get out
+// out of the provided constructor into the provided ConstructorInfo.
+func WithInfo(info *ConstructorInfo) ProvideOption {
```
abhinav, line 146: Couple things: The other ProvideOptions don't have a With prefix. The divergence is fine because those were written before we agreed that the prefix was desirable on most options, but in this case, "WithFoo" usually suggests you're supplying Foo, not requesting Foo. Should we name this something else? Like, `FillConstructorInfo` (or better?—that's just the first idea that came to mind).

dig.go
```diff
@@ -498,9 +519,23 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
...
 		c.isVerifiedAcyclic = true
 	}
-
 	c.nodes = append(c.nodes, n)
 
+	// Record introspection info for caller if Info option is specified
+	if opts.Info != nil {
+		opts.Info.ID = (int)(n.id)
+		opts.Info.Inputs = make([]string, 0)
+		opts.Info.Outputs = make([]string, 0)
```
abhinav, line 528: This can be shortened with `:=` inside if:
```suggestion
	if info := opts.Info; info != nil {
		info.ID = (int)(n.id)
		info.Inputs = make([]string, 0)
		info.Outputs = make([]string, 0)
```

dig.go
```diff
@@ -498,9 +519,23 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
...
+
+		for _, res := range n.ResultList().DotResult() {
```
abhinav, line 530: If we know the number of items, we should pre-allocate the slice to avoid resizing. https://github.com/uber-go/guide/blob/master/style.md#specifying-slice-capacity

Either:

```
outputs := make([]string, 0, len(results))
for _, res := range results { outputs = append(outputs, res.Type.String()) }
```

Or,

```
outputs := make([]string, len(results))
for i, res := range results { outputs[i] = res.Type.String() }
```

abhinav (commented):
I agree with the reluctance to make the `[]string`s the main API.
If possible, we should leave room for exposing more information.

dig.go
sywhang: That's fair. I'll go with `FillInfo`.

dig.go
```diff
@@ -128,6 +129,45 @@ func Group(group string) ProvideOption {
...
+// which are Stringers that report the types of the parameters and results respectively.
+type ConstructorInfo struct {
+	ID      ID
+	Inputs  []*Input
+	Outputs []*Output
+}
+
+// Input is a stringer that report the type of an input parameters of the constructor.
+type Input struct {
+	t reflect.Type
```
abhinav, line 147: Looks like we're losing some information:

- name of the input, if any
- whether the input is optional
- value group of the input, if any

The dot.Param has this information, FWIW.

dig.go
```diff
@@ -128,6 +129,45 @@ func Group(group string) ProvideOption {
...
+}
+
+func (i *Input) String() string {
+	return i.t.String()
+}
+
+// Output is a stringer that report the types of an output produced by the constructor.
+type Output struct {
+	t reflect.Type
```
abhinav, line 156: Same. Looks like we're losing whether this output has a name or group.

dig.go
```diff
@@ -498,9 +538,25 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
...
+	if info := opts.Info; info != nil {
+		params := n.ParamList().DotParam()
+		results := n.ResultList().DotResult()
+
+		info.ID = (ID)(n.id)
+		info.Inputs = make([]*Input, len(params))
+		info.Outputs = make([]*Output, len(results))
+
+		for i, param := range params {
+			info.Inputs[i] = &Input{t: param.Type}
```
abhinav, line 553: Maybe, 

```
info.Inputs[i] = &Input{
    t: param.Type,
    optional: param.Optional,
    name: param.Name,
    group: param.Group,
}
```

And the `String()` for Input then needs to account for some of this information. We can pick whatever format we prefer for it. I'm leaning something like:

```
$type[optional, name = "$name", group = "$group"]
```

Where the `optional`, `name = "$name"` or `group = "$group"` is added only if those values are set, and the `[`, `]` is added only if at least one entry was added. So  the following are all valid:

```
*db.Conn
*zap.Logger[optional]
*db.Conn[optional, name = "readonly"]
[]http.Handler[group = "http"]
```

dig_test.go
```diff
@@ -2990,3 +2990,115 @@ func TestUnexportedFieldsFailures(t *testing.T) {
 			`bad argument 1: invalid value "foo" for "ignore-unexported" tag on field In: strconv.ParseBool: parsing "foo": invalid syntax`)
 	})
 }
+
+func TestConstructorInfoOption(t *testing.T) {
+	t.Parallel()
```
abhinav, line 2995: optional: FYI, this only makes the top-level test parallel. If the subtests are parallel as well, you may add t.Parallel at the start of each.

nit: an empty line after t.Parallel call.

abhinav (commented):
Excellent. This leaves us a fair bit of room for expansion.

abhinav:
One question I have, @sywhang is whether we want to also produce information about invokes? Perhaps that can be a separate thing later based on needs. 

RE: naming: What do you think about ProvideInfo instead of ConstructorInfo? That'll make it easy to name the Invoke variant InvokeInfo if/when we add it.

sywhang:
> One question I have, @sywhang is whether we want to also produce information about invokes? Perhaps that can be a separate thing later based on needs.

That sounds reasonable. I can work on that as a follow-up to this. 

Re: names, I'll change it to `ProvideInfo` and rename `FillInfo` to `FillProvideInfo`.

dig_test.go
```diff
@@ -2990,3 +2990,125 @@ func TestUnexportedFieldsFailures(t *testing.T) {
...
+		type type2 struct{}
+		ctor := func() (*type1, *type2) {
+			return &type1{}, &type2{}
+		}
+
+		c := New()
+		info := ProvideInfo{}
+		require.NoError(t, c.Provide(ctor, FillProvideInfo(&info)))
+
+		assert.Equal(t, 0, len(info.Inputs))
```
abhinav, line 3007:
```suggestion
		assert.Empty(t, info.Inputs)
```

dig_test.go
```diff
@@ -2990,3 +2990,125 @@ func TestUnexportedFieldsFailures(t *testing.T) {
...
+		assert.Equal(t, 4, len(info.Inputs))
+		assert.Equal(t, 1, len(info.Outputs))
+
+		assert.Equal(t, "*dig.type3", info.Outputs[0].String())
+		assert.Equal(t, "*dig.type1", info.Inputs[0].String())
+		assert.Equal(t, `*io.Writer[optional, name = "rw"]`, info.Inputs[1].String())
+		assert.Equal(t, `*io.Reader[name = "ro"]`, info.Inputs[2].String())
+		assert.Equal(t, `[]string[group = "server"]`, info.Inputs[3].String())
+	})
+
```
abhinav, line 3061: nice

dig_test.go
```diff
@@ -2990,3 +2990,125 @@ func TestUnexportedFieldsFailures(t *testing.T) {
...
+	t.Parallel()
+	t.Run("two outputs", func(t *testing.T) {
+		type type1 struct{}
+		type type2 struct{}
+		ctor := func() (*type1, *type2) {
+			return &type1{}, &type2{}
+		}
+
+		c := New()
+		info := ProvideInfo{}
```
abhinav, line 3004: minor stylistic nit: use var form for zero value.

https://github.com/uber-go/guide/blob/master/style.md#use-var-for-zero-value-structs
```suggestion
		var info ProvideInfo
```

dig_test.go
```diff
@@ -2990,3 +2990,125 @@ func TestUnexportedFieldsFailures(t *testing.T) {
...
+
+	t.Run("two inputs and one output", func(t *testing.T) {
+		type type1 struct{}
+		type type2 struct{}
+		type type3 struct{}
+		ctor := func(*type1, *type2) *type3 {
+			return &type3{}
+		}
+		c := New()
+		info := ProvideInfo{}
```
abhinav, line 3022:
```suggestion
		var info ProvideInfo
```

dig_test.go
```diff
@@ -2990,3 +2990,125 @@ func TestUnexportedFieldsFailures(t *testing.T) {
...
+			ConnNames    []string   `group:"server"`
+		}
+
+		type type3 struct{}
+
+		ctor := func(*type1, GatewayParams) (*type3, error) {
+			return &type3{}, nil
+		}
+		c := New()
+		info := ProvideInfo{}
```
abhinav, line 3049:
```suggestion
		var info ProvideInfo
```

abhinav (approved):
LGTM!

PR #282, opened 2021-06-21
Add LocationForPC ProvideOption

### LINKED ISSUE #281: ProvideOption to override digreflect.Func
## Problem

It is really hard to debug errors with constructors created by reflection because the error will just say `"reflect".makeFuncStub (/Users/arc/sdk/go1.16.3/src/reflect/asm_amd64.s:14)`.

## Proposed Solution

I propose creating a `ProvideOption` that allows setting `digreflect.Func` manually so that frameworks that are building functions with reflection can override this information to provide better error messages to users. I imagine this would improve Fx as well because `Supply` builds a constructor by reflection.

aaronc:
Closes #281

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -205,7 +205,7 @@ func FillProvideInfo(info *ProvideInfo) ProvideOption {
 // and DOT graphs. This option is intended to be used with functions created
 // with the reflect.MakeFunc method whose error messages are otherwise hard to
 // understand
-func LocationForPC(funcPc uintptr) ProvideOption {
+func LocationForPC(pc uintptr) ProvideOption {
 	return provideOptionFunc(func(opts *provideOptions) {
 		opts.Location = digreflect.InspectFuncPC(funcPc)
 	})
--- a/dig.go
+++ b/dig.go
@@ -207,7 +207,7 @@ func FillProvideInfo(info *ProvideInfo) ProvideOption {
 // understand
 func LocationForPC(pc uintptr) ProvideOption {
 	return provideOptionFunc(func(opts *provideOptions) {
-		opts.Location = digreflect.InspectFuncPC(funcPc)
+		opts.Location = digreflect.InspectFuncPC(pc)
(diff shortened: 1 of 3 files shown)


internal/digreflect/func.go
```diff
@@ -66,9 +66,15 @@ func (f *Func) Format(w fmt.State, c rune) {
...
 func InspectFunc(function interface{}) *Func {
 	fptr := reflect.ValueOf(function).Pointer()
-	f := runtime.FuncForPC(fptr)
+	return InspectFuncPC(fptr)
+}
+
+// InspectFuncPC inspects and returns runtime information about the function
+// at the given program counter address.
+func InspectFuncPC(pc uintptr) *Func {
+	f := runtime.FuncForPC(pc)
```
sywhang, line 75: Can we add a check for `f` being `nil` here? Upon an invalid PC we'll panic in the next line.

dig.go
```diff
@@ -198,6 +199,18 @@ func FillProvideInfo(info *ProvideInfo) ProvideOption {
 	})
 }
 
+// LocationForPC is a ProvideOption which specifies an alternate function program
+// counter address to be used for debug information. The package, name, file and
+// line number of this alternate function address will be used in error messages
+// and DOT graphs. This option is intended to be used with functions created
+// with the reflect.MakeFunc method whose error messages are otherwise hard to
+// understand
+func LocationForPC(funcPc uintptr) ProvideOption {
```
sywhang, line 208: Can we just call the argument `pc` instead of `funcPc`? I think it's a little redundant to call this "FuncPc" since `pc` already encapsulates the info that this should be pointing to a func.

```suggestion
func LocationForPC(pc uintptr) ProvideOption {
```

sywhang (commented):
Thank you for the contribution! Left couple of comments but looks mostly good to me otherwise.

sywhang (approved):

PR #288, opened 2021-08-19
Add support to GroupInvoke

RobyFerro:
This implementation allows developers to invoke specific function between multiple containers.
This may be useful in case of you have a container instantiated at application startup (singleton container) and a second container instantiated just before handle every HTTP request (factory container).

With _GroupInvoke_ you can resolve dependencies across singleton and factory container.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -23,6 +23,7 @@ package dig
 import (
 	"errors"
 	"fmt"
+	"log"
 	"math/rand"
 	"reflect"
 	"sort"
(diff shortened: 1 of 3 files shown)


abhinav:
@RobyFerro, I know that you've closed this PR,
but I want to take a second to call out that
Dig and Fx rely heavily on reflection to implement their functionality.
For that reason, we advise that you use them only at startup,
and do not use them on the request path.

Creating per-request containers to (presumably) build a request-scoped handler
is not something we suggest doing with Dig. We instead recommend that you work
around this limitation by manually implementing request-scoped handlers.

For example, suppose your service interface is the following.

```go
package keyvalue

type Interface interface {
  Set(context.Context, *SetRequest) error
  Get(context.Context, *GetRequest) (*GetResponse, error)
}
```

You can build a `Handler` that pretends it's request scoped:

```go
type Handler struct {
  // ...
  cache map[something]somethingElse
}

func (h *Handler) Set(context.Context, *keyvalue.SetRequest) error {
  h.cache[...]
}
```

But instead of building a constructor for it, use of a factory type, which is constructed with dependency injection.

```go
type HandlerFactory struct {
  log *zap.Logger
  // ..
}

func NewHandlerFactory(log *zap.Logger) *HandlerFactory { ... }

// In main, fx.Provide(NewHandlerFactory)

func (f *HandlerFactory) NewHandler(...) *Handler {
  return &Handler{
    log: log,
    cache: make(...),
  }
}
```

And have HandlerFactory implement keyvalue.Interface, but delegate to a new Handler each request.

```
func (f *HandlerFactory) Set(ctx context.Context, req *keyvalue.SetRequest) error {
  return f.NewHandler(...).Set(ctx, req)
}

func (f *HandlerFactory) Get(ctx context.Context, req *keyvalue.GetRequest) (*keyvalue.GetResponse, error) {
  return f.NewHandler(...).Get(ctx, req)
}
```

Then in your main, instead of passing `Handler` as the implementation of `Interface`, pass in `HandlerFactory`.

```
fx.Provide(func(f *HandlerFactory) keyvalue.Interface { return f })
```

PR #289, opened 2021-08-19
Add support to invoke a function between multiple containers

RobyFerro:
This implementation allows developers to invoke specific function between multiple containers.
This may be useful in case of you have a container instantiated at application startup (singleton container) and a second container instantiated just before handle every HTTP request (factory container).

With _GroupInvoke_ you can resolve dependencies across singleton and factory container.

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -23,7 +23,6 @@ package dig
 import (
 	"errors"
 	"fmt"
-	"log"
 	"math/rand"
 	"reflect"
 	"sort"
@@ -138,6 +137,7 @@ func GroupInvoke(function interface{}, containers ...*Container) error {
 	}
 
 	reflect.ValueOf(function).Call(arguments)
+
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


RobyFerro:
This method can be also used to invoke reflected interface:
```
if err := dig.GroupInvoke(method.Interface(), container, singletonIOC); err != nil {
    // Handle error
}
```

abhinav:
Oops, I missed that this was a re-open of #288.
Copying part of https://github.com/uber-go/dig/pull/288#issuecomment-961485428  here.

---

Dig and Fx rely heavily on reflection to implement their functionality.
For that reason, we advise that you use them only at startup,
and do not use them on the request path.

Creating per-request containers to (presumably) build a request-scoped handler
is not something we suggest doing with Dig. We instead recommend that you work
around this limitation by manually implementing request-scoped handlers.
[Original comment has an example of the recommended workaround.]

PR #293, opened 2021-09-15
Add Dig.As ProvideOption

sywhang:
This brings back `Dig.As` from #233 with some changes. Specifically, `Dig.Provide(t1, As(t2))`used to provide type `t2` as well as type `t1` but that is no longer the case.

For example, `c.Provide(newBuffer, dig.As(new(io.Reader), new(io.Writer)))` will provide `io.Reader` and `io.Writer`, but not `buffer`. 

Refs: GO-451

### DIFF
--- a/dig.go
+++ b/dig.go
@@ -63,13 +63,19 @@ type provideOptions struct {
 	Name  string
 	Group string
 	Info  *ProvideInfo
+	As    []interface{}
 }
 
 func (o *provideOptions) Validate() error {
-	if len(o.Group) > 0 && len(o.Name) > 0 {
-		return errf(
-			"cannot use named values with value groups",
-			"name:%q provided with group:%q", o.Name, o.Group)
+	if len(o.Group) > 0 {
+		if len(o.Name) > 0 {
+			return fmt.Errorf(
+				"cannot use named values with value groups: name:%q provided with group:%q", o.Name, o.Group)
+		}
+		if len(o.As) > 0 {
+			return fmt.Errorf(
+				"cannot use dig.As with value groups: dig.As provided with group:%q", o.Group)
+		}
 	}
 
 	// Names must be representable inside a backquoted string. The only
@@ -81,6 +87,23 @@ func (o *provideOptions) Validate() error {
 	}
 	if strings.ContainsRune(o.Group, '`') {
 		return errf("invalid dig.Group(%q): group names cannot contain backquotes", o.Group)
+	}
+
+	for _, i := range o.As {
+		t := reflect.TypeOf(i)
+
+		if t == nil {
+			return fmt.Errorf("invalid dig.As(nil): argument must be a pointer to an interface")
+		}
+
+		if t.Kind() != reflect.Ptr {
+			return fmt.Errorf("invalid dig.As(%v): argument must be a pointer to an interface", t)
+		}
+
+		pointingTo := t.Elem()
+		if pointingTo.Kind() != reflect.Interface {
+			return fmt.Errorf("invalid dig.As(*%v): argument must be a pointer to an interface", pointingTo)
+		}
 	}
 	return nil
 }
@@ -195,6 +218,55 @@ func (o *Output) String() string {
 func FillProvideInfo(info *ProvideInfo) ProvideOption {
 	return provideOptionFunc(func(opts *provideOptions) {
 		opts.Info = info
+	})
+}
+
+// As is a ProvideOption that specifies that the value produced by the
+// constructor implements one or more other interfaces.
+//
+// As expects one or more pointers to the implemented interfaces. Values
+// produced by constructors will be made available in the container as
+// implementations of all of those interfaces.
+//
+// For example, the following will make the buffer available in the container
+// as io.Reader and io.Writer.
+//
+//   c.Provide(newBuffer, dig.As(new(io.Reader), new(io.Writer)))
+//
+// That is, the above is equivalent to the following.
+//
+//   c.Provide(func(...) (*bytes.Buffer, io.Reader, io.Writer) {
+//     b := newBuffer(...)
+//     return b, b, b
+//   })
+//
+// If used with dig.Name, the type produced by the constructor and the types
+// specified with dig.As will all use the same name. For example,
+//
+//   c.Provide(newFile, dig.As(new(io.Reader)), dig.Name("temp"))
+//
+// The above is equivalent to the following.
+//
+//   type Result struct {
+//     dig.Out
+//
+//     File   *os.File  `name:"temp"`
+//     Reader io.Reader `name:"temp"`
+//   }
+//
+//   c.Provide(func(...) Result {
+//     f := newFile(...)
+//     return Result{
+//       File: f,
+//       Reader: f,
+//     }
+//   })
+//
+// This option cannot be provided for constructors which produce result
+// objects.
+func As(i ...interface{}) ProvideOption {
+	return provideOptionFunc(func(opts *provideOptions) {
+		opts.As = append(opts.As, i...)
 	})
 }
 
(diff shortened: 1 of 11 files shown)


sywhang:
Hey @alessandrozucca, I know it's been a long time but we're bringing back `Dig.As`. Could you sign the CLA so that we can move forward with this? Thanks!

alessandrozucca:
@sywhang all signed 🙌

sywhang:
Awesome! Thank you :)

dig.go
```diff
@@ -674,32 +746,24 @@ func (cv connectionVisitor) Visit(res result) resultVisitor {
...
+				*cv.err = err
+				return nil
 			}
-
-			*cv.err = errf(
-				"cannot provide %v from %v", k, path,
-				"already provided by %v", strings.Join(cons, "; "),
-			)
-			return nil
+			cv.keyPaths[k] = path
```
abhinav, line 764: Should both these `cv.keyPaths[k] = path` statements be in checkKey?

dig_test.go
```diff
@@ -627,6 +629,69 @@ func TestEndToEndSuccess(t *testing.T) {
...
 	})
 
+	t.Run("struct constructor with as interface option", func(t *testing.T) {
+		c := New()
+
+		provider := c.Provide(
+			func() *bytes.Buffer {
+				var buf bytes.Buffer
+				buf.WriteString("foo")
+				return &buf
```
abhinav, line 639: optional: can replace with `bytes.NewBufferString("foo")`

dig_test.go
```diff
@@ -627,6 +629,69 @@ func TestEndToEndSuccess(t *testing.T) {
...
+			},
+			As(new(fmt.Stringer), new(io.Reader)),
+		)
+
+		require.NoError(t, provider, "provide failed")
```
abhinav, line 644: `provider` appears to be an error. it should probably be named `err`.
```suggestion
		require.NoError(t, err, "provide failed")
```

dig_test.go
```diff
@@ -627,6 +629,69 @@ func TestEndToEndSuccess(t *testing.T) {
...
+
+		require.NoError(t, c.Invoke(
+			func(s fmt.Stringer, r io.Reader) {
+				require.Equal(t, "foo", s.String(), "invoke got new buffer")
+				got, err := ioutil.ReadAll(r)
+				assert.NoError(t, err, "failed to read from reader")
+				require.Equal(t, "foo", string(got), "invoke got new buffer")
+			},
+		), "invoke failed")
+	})
```
abhinav, line 654: I would add another check to this test to make sure there's no `*bytes.Buffer` in the container.

```
require.Error(t, c.Invoke(func(*bytes.Buffer) {
    t.Fatalf("must not be called")
}), "must not have a *bytes.Buffer in the container")
```

dig_test.go
```diff
@@ -627,6 +629,69 @@ func TestEndToEndSuccess(t *testing.T) {
...
+			In
+
+			Buffer *bytes.Buffer `name:"buff"`
+			Reader io.Reader     `name:"buff"`
+		}
+
+		require.Error(t, c.Invoke(func(got in) {
+			body, err := ioutil.ReadAll(got.Reader)
+			require.NoError(t, err, "failed to read buffer body")
+			assert.Equal(t, "foo", string(body))
```
abhinav, line 677: If the body is not meant to run, a `t.Fatal` might be merited there.

dig_test.go
```diff
@@ -1537,16 +1602,111 @@ func TestProvideInvalidGroup(t *testing.T) {
...
+			name:        "as param is a nil",
+			param:       nil,
+			expectedErr: "invalid dig.As(nil): argument must be a pointer to an interface",
+		},
+		{
+			name:        "as param is a func",
+			param:       func() {},
+			expectedErr: "invalid dig.As(func()): argument must be a pointer to an interface",
+		},
+	}
```
abhinav, line 1643: another one for `dig.As` on a function that returns a `dig.Out`?

dig_test.go
```diff
@@ -1537,16 +1602,111 @@ func TestProvideInvalidGroup(t *testing.T) {
...
+		type in struct {
+			In
+
+			Buffer *bytes.Buffer `name:"buff"`
+			Reader io.Reader     `name:"buff"`
+		}
+
+		require.Error(t, c.Invoke(func(got in) {
+			assert.Nil(t, got.Buffer, "buffer should not be provided.")
+			assert.NotNil(t, got.Reader, "reader should be provided.")
```
abhinav, line 1683: the function body will never be called because `*bytes.Reader` with that name won't be present.

dig_test.go
```diff
@@ -1537,16 +1602,111 @@ func TestProvideInvalidGroup(t *testing.T) {
...
 
-func TestProvideGroupAndName(t *testing.T) {
+func TestProvideInvalidAs(t *testing.T) {
+	ptrToStruct := &struct {
+		name string
+	}{
+		name: "example",
+	}
+	var nilInterface io.Reader
+	c := New()
```
abhinav, line 1612: i would do container per subtest. there doesn't seem to be any need for sharing?

dig_test.go
```diff
@@ -1537,16 +1602,111 @@ func TestProvideInvalidGroup(t *testing.T) {
...
+		},
+		{
+			name:        "as param is a func",
+			param:       func() {},
+			expectedErr: "invalid dig.As(func()): argument must be a pointer to an interface",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := c.Provide(
```
abhinav, line 1646: consider adding t.Parallel

testdata/dig_as_two.dot
```diff
@@ -0,0 +1,16 @@
...
+	rankdir=RL;
+	graph [compound=true];
+	
+		subgraph cluster_0 {
+			label = "go.uber.org/dig";
+			constructor_0 [shape=plaintext label="TestVisualize.func4.1"];
+			
+			"io.Reader" [label=<io.Reader>];
+			"io.Writer" [label=<io.Writer>];
+
```
abhinav, line 11: hell yeah

abhinav (commented):
Looks pretty good. A few comments here and there but the actual implementation is straightforward. Thanks!

dig_test.go
```diff
@@ -1537,16 +1603,123 @@ func TestProvideInvalidGroup(t *testing.T) {
...
+	t.Parallel()
+
+	t.Run("group and name", func(t *testing.T) {
+		c := New()
+		err := c.Provide(func() io.Reader {
+			panic("this function must not be called")
+		}, Group("foo"), Name("bar"))
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "cannot use named values with value groups: "+
+			"name:\"bar\" provided with group:\"foo\"")
```
abhinav, line 1711: optional: can use tilde for string literals to avoid escaping `"`

```suggestion
			`name:"bar" provided with group:"foo"`)
```

result.go
```diff
@@ -243,21 +247,69 @@ func (rl resultList) ExtractList(cw containerWriter, values []reflect.Value) err
...
+
+	for _, as := range opts.As {
+		ifaceType := reflect.TypeOf(as).Elem()
+		if !t.Implements(ifaceType) {
+			return r, fmt.Errorf("invalid dig.As: %v does not implement %v", t, ifaceType)
+		}
+		if ifaceType == t {
+			// Special case:
+			//   c.Provide(func() io.Reader, As(new(io.Reader)))
+			// Ignore instead of erroring out.
```
abhinav, line 272: Should this check be before the `t.Implements` then?

abhinav (approved):

PR #301, opened 2021-11-16
Refactor graph representation 

sywhang:
This is the first PR to prepare Dig for in-graph modifications. 

This PR contains several changes to allow further extensibility in the graph representation used in Dig, and decouple reflection Type info from the graph algorithms as much as possible.

Specifically, this PR makes the following changes:

- Add `internal/graph` package which contains Dig-agnostic graph algorithm used for cycle detection
- Remove the somewhat duplicated embedded cycle detection logic in Container and cycle.go
- Add `graphHolder` type in `Container` which holds all the graph nodes and implements the graph iterator interface used for graph algorithm in `internal/graph`.
- Add `paramGroupedSlice` as part of node in graph.
- Remove the param visitor funcs and related helpers.
- Rename `node` to `constructorNode`.

There is also a change in the way cycle error messages are displayed. Previously, the errors looked like:

```
typeA provided by funcName (file_location.go:line_number)
  depends on typeB provided by funcName (file_location.go:line_number)
  depends on typeC provided by funcName (file_location.go:line_number)
 ... and so on
```

This error message generation depended on knowing the "edge" info between nodes, which is the reflect.Type that connects the two constructor nodes. 

Now that the graph representation holds not only constructor nodes but also value groups, and because graph representation does not know what type holds the nodes together, this needs to be changed.

Now, the error messages look like:
```
func(*typeB) *typeA provided by funcName(file_location.go:line_number)
  depends on func(*typeC) *typeB provided by funcName(file_location.go:line_number)
  depends on func(*typeA) *typeC provided by funcName(file_location.go:line_number)
... and so on
```

(Note that the function signature is now printed instead of the type that holds the nodes as dependencies).

For param value groups, the error message skips over it so that the error message doesn't look cluttered but that can be subject to change in the future.


internal/graph/graph.go
```diff
@@ -0,0 +1,112 @@
...
+type cycleInfo struct {
+	// order is the number of nodes in the graph
+	order int
+
+	// records whether ith node has been visited.
+	visited []bool
+	// records whether ith node is currently on the recursion stack.
+	onStack []bool
+	// back-tracks each edge info to form cycle path if one is detected.
+	backtrack map[int]int
```
abhinav, line 101: We might be able to collapse these into a single slice.

abhinav: We can probably use a single slice of a richer structure instead.

```
type cycleNode struct {
  Visited bool
  OnStack bool
  Backtrack int
}
```

shirchen:
High-level comment: can you refactor existing dig_test file to add DeferAcyclicVerification option to some tests. I think we are currently missing some code coverage around it. Ref GO-76

param.go
```diff
@@ -81,61 +81,6 @@ func newParam(t reflect.Type) (param, error) {
...
-
-// walkParam walks the param tree for the given param with the provided
-// visitor.
-//
-// paramVisitor.Visit will be called on the provided param and if a non-nil
-// paramVisitor is received, this param's descendants will be walked with that
-// visitor.
-//
-// This is very similar to how go/ast.Walk works.
-func walkParam(p param, v paramVisitor) {
```
abhinav, line 117: thanks for deleting this

dig.go
```diff
@@ -1038,3 +1131,8 @@ func shuffledCopy(rand *rand.Rand, items []reflect.Value) []reflect.Value {
 	}
 	return newItems
 }
+
+type graphNode struct {
```
abhinav, line 1135: minor: this should be declared near graphHolder. Better reading order.

dig.go
```diff
@@ -314,6 +314,47 @@ type Container struct {
 
 	// invokerFn calls a function with arguments provided to Provide or Invoke.
 	invokerFn invokerFn
+
+	gh *graphHolder
+}
+
+// graphHolder represents the dependency graph for a Container. Specifically,
```
abhinav, line 321: nit: can we move graphHolder and graphNode to a graph.go. (There's already one, but that's for graph visualization and can be renamed to visualize.go.)

dig.go
```diff
@@ -314,6 +314,47 @@ type Container struct {
...
+// it saves constructorNodes and paramGroupedSlices (value groups) as graphNodes
+// and implements the Graph interface defined in internal/graph to run graph
+// algorithms on it. It has a 1-to-1 correspondence with a Container whose graph
+// it represents.
+type graphHolder struct {
+	// all the nodes defined in the graph.
+	allNodes []*graphNode
+
+	// Maps each graphNode to its order.
```
abhinav, line 330: High-level comment here and in-general about the documentation:
That "graph order" refers to the number of vertices is obvious,
but that "node order" refers to the index of the node in allNodes is not obvious.

dig.go
```diff
@@ -314,6 +314,47 @@ type Container struct {
...
+	n := gh.allNodes[u]
+
+	var orders []int
+
+	switch w := n.Wrapped.(type) {
+	case *constructorNode:
+		for _, param := range w.paramList.Params {
+			orders = append(orders, getParamOrder(gh, param)...)
+		}
+	case *paramGroupedSlice:
```
abhinav, line 351: It feels like breaking a boundary here to use paramGroupedSlice as the graph node
because a value group technically has two ends: param and result.

Would it be reasonable to declare a separate `valueGroupNode` type for this?

```go
type valueGroupNode struct {
  Group string
  Type reflect.Type
}
```

param.go
```diff
@@ -277,9 +223,29 @@ func (po paramObject) DotParam() []*dot.Param {
...
+	switch p := param.(type) {
+	case paramSingle:
+		providers := gh.c.getValueProviders(p.Name, p.Type)
+		for _, provider := range providers {
+			v := gh.orders[key{t: provider.CType()}]
+			orders = append(orders, v)
+		}
+	case paramGroupedSlice:
+		v := gh.orders[key{t: p.Type, group: p.Group}]
+		orders = append(orders, v)
```
abhinav, line 237: Is this right? It seems significantly different from the caller's handling of paramGroupedSlice.

https://github.com/uber-go/dig/pull/301/files#diff-a27f75d4c63afb2160bde845e6453abce79e88c5ee81d146711382455f05c1e3R352-R355

I also can't tell for sure if this will actually happen, to be honest.
I _think_ that this is resolving the dependency from function argument to value group node,
but it's a bit unclear.
If that's right, _maybe_ separating paramGroupedSlice from valueGroupNode might make that clearer?

dig.go
```diff
@@ -380,18 +424,27 @@ type provider interface {
...
 		rand:      rand.New(rand.NewSource(time.Now().UnixNano())),
 		invokerFn: defaultInvoker,
 	}
 
+	gh := &graphHolder{
+		orders: make(map[key]int),
+		c:      c,
+	}
+
+	c.gh = gh
```
abhinav, line 446: nit:
```suggestion
	c.gh = &graphHolder{
		orders: make(map[key]int),
		c:      c,
	}
```

dig.go
```diff
@@ -579,8 +632,8 @@ func (c *Container) Invoke(function interface{}, opts ...InvokeOption) error {
 	}
 
 	if !c.isVerifiedAcyclic {
-		if err := c.verifyAcyclic(); err != nil {
-			return err
+		if ok, cycle := graph.IsAcyclic(c.gh); !ok {
+			return errf("cycle detected in dependency graph", c.cycleDetectedError(cycle))
```
abhinav, line 636: minor unintentional change in behavior here:
We won't mark the container as acyclic `if ok`, which we were doing before.
We should probably still mark it acyclic afterwards so that we don't check again for every invoke until it's dirty again.

dig.go
```diff
@@ -954,36 +1022,52 @@ func isIgnoreUnexportedSet(f reflect.StructField) (bool, error) {
...
+			if ns := c.getValueProviders(p.Name, p.Type); len(ns) == 0 && !p.Optional {
+				missingDeps = append(missingDeps, p)
+			}
+		case paramObject:
+			for _, f := range p.Fields {
+				missingDeps = append(missingDeps, findMissingDependencies(c, f.Param)...)
+
+			}
+		case paramList:
+			missingDeps = append(missingDeps, findMissingDependencies(c, p.Params...)...)
```
abhinav, line 1058: This case is impossible. We can't have a paramList here because that represents the parameters of the constructor. There can't be another list of parameters inside it.

dig.go
```diff
@@ -954,36 +1022,52 @@ func isIgnoreUnexportedSet(f reflect.StructField) (bool, error) {
...
 	}
 	return nil
 }
 
+func findMissingDependencies(c containerStore, params ...param) []paramSingle {
+	var missingDeps []paramSingle
+
+	for _, param := range params {
+		switch p := param.(type) {
+		case paramSingle:
```
abhinav, line 1048: Do we need to address the paramGroupedSlice case here at all?

abhinav (commented):
So one thing I noticed with looking up orders in the `map[key]int`:
If there's no match, we get back a zero.
This seems like it will be easy to cause a silent bug with,
because zero will resolve to the first node.
We should maybe place a sentinel invalid node at index zero.
That'll help us find issues if we accidentally end up on it.

---

It's also a bit unclear how keying on the constructor type is safe?
Is it not possible that we provide two constructors with the same signature?
Like, say two constructors pushing to the same value group?

sywhang:
> It's also a bit unclear how keying on the constructor type is safe?
> Is it not possible that we provide two constructors with the same signature?
> Like, say two constructors pushing to the same value group?

@abhinav So I thought about this for a while and used multiple ways to test it, and current implementation does work. I also thought this wouldn't work, but it turns out that it's still fine. Here's why:

Currently the way this works is that `graphHolder.orders` is a map that maps a type of either a constructor or a value group to an order. 

Conceptually, if we provide two constructors of the same type, then it is okay for the graph representation to contain a single node to represent that, because currently, dependency resolution is still performed by the `providers` map defined in `Container`, and that is a map of key to a list of `constructorNode`s.  The graph representation still *holds* because cycle detection algorithm (which is the only graph algorithm that we currently run on this representation) does not care if there is one node or multiple constructorNodes with the same types.

For example, given 4 constructors ctor1, ctor2, ctor3, and ctor4 where ctor3 and ctor4 are of the same type, this graph:

Screen Shot 2021-12-02 at 9 45 56 PM

is the "same" graph as this one:

Screen Shot 2021-12-02 at 9 46 24 PM.

dig.go
sywhang: not really, since this is only checking for direct dependencies of the provided function, and paramGroupedSlice is a part of paramObject always.

internal/graph/graph_test.go
```diff
@@ -0,0 +1,156 @@
...
+			},
+		},
+	}
+	for _, tt := range testCases {
+		g := newTestGraph()
+		for i, neighbors := range tt.edges {
+			g.Nodes[i+1] = neighbors
+		}
+		ok, _ := IsAcyclic(g)
+		assert.True(t, ok)
```
abhinav, line 96: We could add a message if a cycle was detected:

```suggestion
		ok, cycle := IsAcyclic(g)
		assert.True(t, ok, "expected acyclic, got cycle %v", cycle)
```

internal/graph/graph.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// nodes is the information for a given node.
+	nodes []cycleNode
+}
+
+func newCycleInfo(order int) *cycleInfo {
+	return &cycleInfo{
+		order: order,
+		// +1 because 0 is always a sentinel value.
+		nodes: make([]cycleNode, order+1),
```
abhinav, line 128: Wait, this is off. The internal graph type and its algorithm doesn't have a concept of a sentinel node.
In Dig's usage of this graph, it happens to put a sentinel node at position 0,
and that node will have no dependencies or dependents,
but that shouldn't leak to this code here?

internal/graph/graph_test.go
```diff
@@ -0,0 +1,156 @@
...
+				{6},
+				nil,
+			},
+		},
+	}
+	for _, tt := range testCases {
+		g := newTestGraph()
+		for i, neighbors := range tt.edges {
+			g.Nodes[i+1] = neighbors
+		}
```
abhinav, line 94: Yeah, that `+1` there is also a result of leaking dig's use of a sentinel node to the graph algorithm here.

internal/graph/graph_test.go
```diff
@@ -0,0 +1,156 @@
...
+	for _, tt := range testCases {
+		g := newTestGraph()
+		for i, neighbors := range tt.edges {
+			g.Nodes[i] = neighbors
+		}
+		ok, c := IsAcyclic(g)
+		assert.False(t, ok)
+		for _, node := range tt.cycle {
+			assert.Contains(t, c, node)
+		}
```
abhinav, line 154: Should we be asserting the exact path here? 

```suggestion
		assert.Equal(t, tt.cycle, c)
```

param.go
abhinav: @sywhang I might be misunderstanding some of the code here. Is this not a concern?

internal/graph/graph.go
sywhang: oh, yes. That's correct. I'll correct this.

internal/graph/graph_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (g TestGraph) EdgesFrom(u int) []int {
+	return g.Nodes[u]
+}
+
+func TestGraphIsAcyclic(t *testing.T) {
+	testCases := []struct {
+		edges [][]int
+	}{
+		// 1
```
shirchen, line 51: Can you add some comments about how one would add a test case here?

cycle.go
```diff
@@ -39,10 +39,10 @@ type errCycleDetected struct {
 func (e errCycleDetected) Error() string {
```
shirchen, line 39: nit: Consider renaming this file to cycle_error.go.
nit: Rename `cycleEntry` into something more related to the error.

param.go
sywhang: talked offline on Slack but for records:

these two are doing different things: EdgesFrom tries to get the orders of the dependencies of a paramGroupedSlice. 
getParamOrder tries to get the order of a paramGroupedSlice from the graph.

dig.go
```diff
@@ -641,16 +664,13 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) error {
 	}
 
 	for k := range keys {
-		c.isVerifiedAcyclic = false
-		oldProviders := c.providers[k]
 		c.providers[k] = append(c.providers[k], n)
```
sywhang, line 667: actually this might be an unintentional behavior change because c.providers[k] should not be changed when a cycle is introduced by a newly provided constructor, but with the current changes, it will be. Let me fix that.

graph.go
```diff
@@ -93,27 +70,31 @@ func (gh *graphHolder) EdgesFrom(u int) []int {
 	case *paramGroupedSlice:
 		providers := gh.c.getGroupProviders(w.Group, w.Type.Elem())
 		for _, provider := range providers {
-			orders = append(orders, gh.orders[key{t: provider.CType()}])
+			orders = append(orders, provider.Order())
```
sywhang, line 77: nice

abhinav (approved):

PR #302, opened 2021-12-04
Enable staticcheck and fix issues

abhinav:
This adds staticcheck to our `make lint` and fixes issues reported by it.

Of the failures reported by staticcheck, the following were actual
issues:

- shallowCheckDependencies: addMissingNodes was unused. I think this was
  intended to go into visualiziation somehow, but we never did it.
- digSentinel fields on dig.In/Out can be hidden away
- A test validating that we don't allow the `Group` provide option for
  `dig.Out`-based structs was not actually using the result struct
  declared there so it wasn't testing what it claimed to test.

Besides that, there were a number of unused fields in test structs.
These are intentionally unused fields so I've appeased the linter by
referencing them with `_ = ..` assignments.


r-hang (approved):

sywhang (approved):
LGTM

PR #303, opened 2021-12-04
dig.go: Split into files by purpose

abhinav:
The dig.go file has grown past readability.
Break it apart into separate files:

- constructor.go defines constructorNode, its method, and the ability
  to call a constructor.
- container.go defines the container type and its related interfaces and
  methods.
- types.go was renamed to inout.go, defining the In and Out types.
- invoke.go defines the Invoke method of the Container type.
- provide.go does the same for the provide method of the Container type.
- stringer.go was deleted in favor of string methods next to their
  types.

This change contains no behavioral changes. Code has only been moved.


param.go
```diff
@@ -168,6 +178,24 @@ func (ps paramSingle) DotParam() []*dot.Param {
 	}
 }
 
+func (sp paramSingle) String() string {
```
sywhang, line 181:
```suggestion
func (ps paramSingle) String() string {
```

param.go
```diff
@@ -168,6 +178,24 @@ func (ps paramSingle) DotParam() []*dot.Param {
...
+	// tally.Scope[optional] means optional
+	// tally.Scope[optional, name="foo"] means named optional
+
+	var opts []string
+	if sp.Optional {
```
sywhang, line 186:
```suggestion
	if ps.Optional {
```

param.go
```diff
@@ -168,6 +178,24 @@ func (ps paramSingle) DotParam() []*dot.Param {
...
+		opts = append(opts, "optional")
+	}
+	if sp.Name != "" {
```
sywhang, line 189:
```suggestion
	if ps.Name != "" {
```

param.go
```diff
@@ -168,6 +178,24 @@ func (ps paramSingle) DotParam() []*dot.Param {
...
+		opts = append(opts, fmt.Sprintf("name=%q", sp.Name))
+	}
+
+	if len(opts) == 0 {
+		return fmt.Sprint(sp.Type)
+	}
+
+	return fmt.Sprintf("%v[%v]", sp.Type, strings.Join(opts, ", "))
```
sywhang, line 197:
```suggestion
	return fmt.Sprintf("%v[%v]", ps.Type, strings.Join(opts, ", "))
```

param.go
```diff
@@ -168,6 +178,24 @@ func (ps paramSingle) DotParam() []*dot.Param {
...
+	var opts []string
+	if sp.Optional {
+		opts = append(opts, "optional")
+	}
+	if sp.Name != "" {
+		opts = append(opts, fmt.Sprintf("name=%q", sp.Name))
+	}
+
+	if len(opts) == 0 {
+		return fmt.Sprint(sp.Type)
```
sywhang, line 194:
```suggestion
		return fmt.Sprint(ps.Type)
```

param.go
```diff
@@ -168,6 +178,24 @@ func (ps paramSingle) DotParam() []*dot.Param {
...
+func (sp paramSingle) String() string {
+	// tally.Scope[optional] means optional
+	// tally.Scope[optional, name="foo"] means named optional
+
+	var opts []string
+	if sp.Optional {
+		opts = append(opts, "optional")
+	}
+	if sp.Name != "" {
+		opts = append(opts, fmt.Sprintf("name=%q", sp.Name))
```
sywhang, line 190:
```suggestion
		opts = append(opts, fmt.Sprintf("name=%q", ps.Name))
```

param.go
```diff
@@ -223,6 +251,14 @@ func (po paramObject) DotParam() []*dot.Param {
 	return types
 }
 
+func (op paramObject) String() string {
```
sywhang, line 254:
```suggestion
func (po paramObject) String() string {
```

param.go
```diff
@@ -223,6 +251,14 @@ func (po paramObject) DotParam() []*dot.Param {
...
+	fields := make([]string, len(op.Fields))
```
sywhang, line 255:
```suggestion
	fields := make([]string, len(po.Fields))
```

param.go
```diff
@@ -223,6 +251,14 @@ func (po paramObject) DotParam() []*dot.Param {
...
+	for i, f := range op.Fields {
```
sywhang, line 256:
```suggestion
	for i, f := range po.Fields {
```

sywhang (approved):
LGTM modulo lint changes

abhinav:
Thanks!

PR #304, opened 2021-12-06
Options: Implement String()

abhinav:
Change option implementations from function pointers
to objects implementing `String()`.

This makes it easier to debug options by printing them
and aligns with the recommended pattern for functional options
in [our style guide].

  [our style guide]: https://github.com/uber-go/guide/blob/master/style.md#functional-options


sywhang (approved):
👍

PR #305, opened 2021-12-08
Introduce Scope to Dig

sywhang:
This introduces the concept of Dig.Scope, which is a scoped container for modifications to the dependency graph. A Scope may have one or more Scopes as children. A Scope also has one parent Scope, except for `Container`, which is the root Scope. This change is necessary before we can start working on APIs for graph modifications and decorations.

Things to note:
1. Modifications to a Scope is "private" by default. For example, `Provide` to a Scope means it will be made available to itself and any of its descendent Scopes (More on this later).
2. `Container`, which is the existing API for DI container, is the just the root `Scope`. 

To make things easier to review, I have broken down the PRs into few parts:
1. Introduce `Dig.Scope` type (this PR)
2. Add exported provide option to Provide, which will allow `Provide`s to a `Scope` to be propagated to all the Scopes without having to touch the root Scope. 
3. Add `Decorate` method to Dig.Scope, which lets you decorate a type provided to a `Scope`. 

Notes on Implementation:

Each `Scope` struct keeps track of the following things:
* `constructorNode`s, `providers`, `groups`, and `values` *all* track things that were provided directly to this Scope. It does not contain anything that was provided by its parent, child, or anything in its ancestry tree. This is so that we can avoid having to copy over these every single time a new Scope is created, which is expensive in terms of memory and computation.
* `gh` still holds all the nodes in the entire dependency graph for nodes that impact this Scope. This is so that we can avoid re-computing the dependency graph every single time a graph algorithm needs to be run.

Ref: GO-939


sywhang:
Marking this as WIP because I want to play around with an alternate implementation for `Scope.Scope()` where I don't have to copy all the providers, values, and groups each time a new Scope is created. But other than that, other parts (such as tests, Container -> Scope renames) are ready to review.

param.go
```diff
@@ -146,9 +146,15 @@ func (pl paramList) Build(containerStore) (reflect.Value, error) {
 // to the underlying constructor.
 func (pl paramList) BuildList(c containerStore) ([]reflect.Value, error) {
 	args := make([]reflect.Value, len(pl.Params))
+	allContainers := c.getStoresUntilRoot()
 	for i, p := range pl.Params {
 		var err error
-		args[i], err = p.Build(c)
+		// iterate through the tree path of scopes.
+		for _, ac := range allContainers {
```
r-hang, line 153: nit: c instead of ac? There's no singular for all.

param.go
```diff
@@ -146,9 +146,15 @@ func (pl paramList) Build(containerStore) (reflect.Value, error) {
...
+			if args[i], err = p.Build(ac); err == nil {
+				break
```
r-hang, line 155: What are you thoughts on a small comment on break
// Build arg from most relevant(?) scope

Appreciation: I thought about inverting the loops here, but then there would be no opportunity to stop computation early since all param should be processed.

provide.go
```diff
@@ -348,6 +348,31 @@ type provider interface {
...
+// The first argument of Provide is a function that accepts zero or more
+// parameters and returns one or more results. The function may optionally
+// return an error to indicate that it failed to build the value. This
+// function will be treated as the constructor for all the types it returns.
+// This function will be called AT MOST ONCE when a type produced by it, or a
+// type that consumes this function's output, is requested via Invoke. If the
+// same types are requested multiple times, the previously produced value will
+// be reused.
+//
+// In addition to accepting constructors that accept dependencies as separate
```
r-hang, line 366: More succinct rewording:

Provide accepts argument types or dig.In structs as dependencies as well as separate return values or dig.Out structs for results.

provide.go
```diff
@@ -411,25 +441,31 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) (err error) {
...
+			s.isVerifiedAcyclic = true
 		}
-		c.isVerifiedAcyclic = true
 	}
-	c.nodes = append(c.nodes, n)
+
+	// Before appending to the nodes, check if there are
+	// any child Scopes. If there are any, recurse down
+	// the descendents to provide the constructor to them.
+	s.nodes = append(s.nodes, n)
```
r-hang, line 468: Where is this check happening? I only see the append.

scope.go
```diff
@@ -0,0 +1,260 @@
...
+	"sort"
+)
+
+// A ScopeOption modifies the default behavior of Scope; currently,
+// there are no implementations.
+type ScopeOption interface {
+	applyScopeOption(*scopeOptions)
+}
+
+type scopeOptions struct {
```
r-hang, line 37: nit: type ... struct {}

scope.go
```diff
@@ -0,0 +1,260 @@
...
+	name string
+	// Mapping from key to all the constructor node that can provide a value for that
+	// key.
+	providers map[key][]*constructorNode
+
+	// constructorNodes provided directly to this Scope. i.e. it does not include
+	// any nodes that were provided to the parent Scope this inherited from.
+	nodes []*constructorNode
+
+	// Values that have already been generated directly in the Scope.
```
r-hang, line 53: wording: can we strike 'that have already been' and say generated directly?

scope.go
```diff
@@ -0,0 +1,260 @@
...
+	for i, graphNode := range s.gh.nodes {
+		child.gh.nodes[i] = graphNode
+	}
+
+	s.childScopes = append(s.childScopes, child)
+	return child
+}
+
+// getScopesUntilRoot creates a list of Scopes
+// have to traverse through until the current node.
```
r-hang, line 113: wording: list of Scopes that traverse from current node to the root? Until "the current node" implies top down and doesn't match the function name.

scope.go
```diff
@@ -0,0 +1,260 @@
...
+	allScopes := []*Scope{s}
+	if len(s.childScopes) > 0 {
+		for _, cs := range s.childScopes {
+			allScopes = append(allScopes, cs.getAllLeafScopes()...)
+		}
+	}
+	return allScopes
+}
+
+func (s *Scope) getStoresUntilRoot() []containerStore {
```
r-hang, line 131: Since a change in order would break BuildList, would be helpful to reinforce the ordering with a comment?

scope_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestScopedOperations(t *testing.T) {
+	t.Parallel()
+
+	t.Run("verify private provides", func(t *testing.T) {
```
r-hang, line 33: nit: I think we can drop the "verify" in the run comment.

scope_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+		assert.NoError(t, grandchild.Invoke(useA))
+		assert.NoError(t, grandchild.Invoke(useB))
+		assert.Error(t, c.Invoke(useB))
+	})
+
+	t.Run("provides to top-level Container propogates to all scopes", func(t *testing.T) {
+		type A struct{}
+
+		// Scope tree:
```
r-hang, line 76: appreciation: graph is helpful

provide.go
```diff
@@ -411,25 +441,31 @@ func (c *Container) provide(ctor interface{}, opts provideOptions) (err error) {
...
-	if !c.deferAcyclicVerification {
-		if ok, cycle := graph.IsAcyclic(c.gh); !ok {
-			// When a cycle is detected, recover the old providers to reset
-			// the providers map back to what it was before this node was
-			// introduced.
-			for k, ops := range oldProviders {
-				c.providers[k] = ops
+	for _, s := range allScopes {
+		s.isVerifiedAcyclic = false
+		if !s.deferAcyclicVerification {
```
moisesvega, line 450: We can reduce nesting here by returning early if s.deferAcyclicVerification is true. 

```
if s.deferAcyclicVerification { continue }
```

scope.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+func (s *Scope) getAllLeafScopes() []*Scope {
+	allScopes := []*Scope{s}
+	if len(s.childScopes) > 0 {
+		for _, cs := range s.childScopes {
+			allScopes = append(allScopes, cs.getAllLeafScopes()...)
+		}
+	}
+	return allScopes
```
abhinav, line 128: FWIW this will do a lot of allocs because every call will allocate that slice only for it to be thrown away and copied into the parent's slice. An alternative API here is:

```
func (s *Scope) appendLeafScopes(dest []*Scope) []*Scope {
  dest = append(dest, s)
  for _, c := range s.childScopes {
    dest = c.appendLeafScopes(dest)
  }
  return dest
}
```

Usage: `s.appendLeafScopes(nil)` (because nil is a valid slice).
This will allocate a single slice in total.

scope.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+func (s *Scope) getStoresUntilRoot() []containerStore {
+	if s.parentScope != nil {
+		return append(s.parentScope.getStoresUntilRoot(), s)
+	}
+	return []containerStore{s}
```
abhinav, line 135: Same as getScopesUntilRoot: iteration might be better than recursion here.

abhinav (commented):
Gave it a high level first pass. Will give it another go later.
I think with the current implementation, invokes will ignore legitimate errors from constructors?

provide.go
sywhang: sorry, that's an outdated comment. will remove

scope.go
```diff
@@ -0,0 +1,260 @@
...
+// made to it in the future will be propagated to the child scope.
+// However, no modifications made to the child scope being created will be propagated
+// to the parent Scope.
+func (s *Scope) Scope(name string, opts ...ScopeOption) *Scope {
+	child := &Scope{
+		name:        name,
+		parentScope: s,
+		providers:   make(map[key][]*constructorNode, len(s.providers)),
+		values:      make(map[key]reflect.Value, len(s.values)),
+		groups:      make(map[key][]reflect.Value, len(s.groups)),
```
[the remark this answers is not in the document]
sywhang: yeah, this is also a remnant from old code. will remove

scope.go
sywhang: will add a test to prevent regression

moisesvega (approved):

param.go
```diff
@@ -146,10 +146,35 @@ func (pl paramList) Build(containerStore) (reflect.Value, error) {
...
+		// scopes had the type. This should be reported.
+		if !argsBuilt[i] {
+			return nil, err
+		}
+
+		// If argument has successfully been built, it's possible
+		// for these errors to occur in child scopes that don't
+		// contain the given parameter type. We can safely ignore
+		// these.
+		// kf it's an error other than missing types/dependencies,
```
r-hang, line 172: "If" instead of "kf"

container.go
```diff
@@ -109,7 +86,7 @@ type containerStore interface {
 	containerWriter
 
 	// Adds a new graph node to the Container and returns its order.
```
r-hang, line 88: There's no return anymore, should we update the doc?

scope.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// child copies the parent's graph nodes.
+	child.gh.nodes = append(child.gh.nodes, s.gh.nodes...)
+
+	s.childScopes = append(s.childScopes, child)
+	return child
+}
+
+// getScopesFromRoot creates a list of Scopes
+// have to traverse through from root until the current node.
```
r-hang, line 112: can remove 'have'

r-hang (approved):

scope.go
```diff
@@ -0,0 +1,262 @@
...
+	"time"
+)
+
+// A ScopeOption modifies the default behavior of Scope; currently,
+// there are no implementations.
+type ScopeOption interface {
+	noScopeOption() //yet
+}
+
+// Scope is a scoped DAG of types and their dependencies.
```
shirchen, line 38: maybe mention that this implements containerStore interface. it wasn't clear (at least to me).

cycle_error.go
```diff
@@ -46,6 +47,7 @@ func (e errCycleDetected) Error() string {
 	//
 	b := new(bytes.Buffer)
 
+	fmt.Fprintf(b, "In Scope %s: \n", e.scope.name)
```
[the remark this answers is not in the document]
sywhang: Yeah, so this is not a debug print. It's here so that the error message contains which Scope the cycle was detected from.

scope.go
sywhang: containerStore is not a public interface so maybe I'll add that inside the struct def so that it's not public.

scope.go
```diff
@@ -0,0 +1,262 @@
...
+	s.gh = newGraphHolder(s)
+	return s
+}
+
+// Scope creates a new Scope with the given name and options from current Scope.
+// Any constructors that the current Scope knows about, as well as any modifications
+// made to it in the future will be propagated to the child scope.
+// However, no modifications made to the child scope being created will be propagated
+// to the parent Scope.
+func (s *Scope) Scope(name string, opts ...ScopeOption) *Scope {
```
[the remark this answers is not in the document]
sywhang: there isn't, I can add that now.

scope_test.go
```diff
@@ -0,0 +1,180 @@
...
+			return &A{}
+		})
+
+		// top-level provide should be available in all the scopes.
+		for _, scope := range allScopes {
+			assert.NoError(t, scope.Invoke(func(a *A) {}))
+		}
+	})
+
+	t.Run("private provides to child should be available to grandchildren, but not root", func(t *testing.T) {
```
shirchen, line 108: hmm.. yeah, im not sure whether this test is a true positive or a negative and it should be grouped below. leaning towards putting this into `TestScopeFailures` but just bringing this up to your attention.

scope.go
```diff
@@ -0,0 +1,262 @@
...
+			})
+		}
+	}
+	return errCycleDetected{Path: path, scope: s}
+}
+
+// String representation of the entire Scope
+func (s *Scope) String() string {
+	b := &bytes.Buffer{}
+	fmt.Fprintln(b, "nodes: {")
```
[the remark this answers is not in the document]
sywhang: yea it's a valid point, though we're writing it to a bytes.Buffer and not some external stream, and doing err check for every single Fprintln in this func can be expensive + hurts readability. i can add it though if you think the tradeoff is worth.

scope_test.go
sywhang: hmm, you're right. this should probably go to failure case. I'll move it now. Thanks for catching it!

shirchen (approved):
lgtm. talked offline about adding permutations for tests for options that could interfere with this diff eg `DryRun(true)` and `DeferAcyclicVerification()`. Maybe there are others I didn't think of. I think its fine not to do that to existing ones, but for new ones, let's try and test more options.

scope.go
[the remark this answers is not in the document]
shirchen: nah, i think you are just adding existing code. just thought i would mention in case it comes up again.

cycle_error.go
abhinav, line 50: nit: I would do something like `[scope "foo"]` here so that the error is:

```
[scope "foo"]
foo provided by bar
  depends on baz provided by ...
```

param.go
```diff
@@ -146,10 +146,35 @@ func (pl paramList) Build(containerStore) (reflect.Value, error) {
...
+		// If argument has successfully been built, it's possible
+		// for these errors to occur in child scopes that don't
+		// contain the given parameter type. We can safely ignore
+		// these.
+		// If it's an error other than missing types/dependencies,
+		// this means some constructor returned an error that must
+		// be reported.
+		_, isErrMissingTypes := err.(errMissingTypes)
+		_, isErrMissingDeps := err.(errMissingDependencies)
+		if err != nil && !isErrMissingTypes && !isErrMissingDeps {
```
abhinav, line 177: I'm not sure I understand what's going here fully:
`err` is assigned in a loop above. The value of `err` here will be the last thing assigned to it before the loop exits.
So this will report an error only if the a constructor in the middle somewhere failed, but one of the ones after it succeeded.

---

Is there risk here in implementing the scope-resolution logic at such a high level in the parameter tree?
If we implemented it on the leaf nodes (paramSingle and maybe paramGroupedSlice), we might be able to control how these operate in a multi-scope environment. Like, paramSingle.Build could call getValueProviders on current containerStore, then its parent, and so on until it finds a provider or runs out of stores.

scope.go
```diff
@@ -0,0 +1,270 @@
...
+	// Parent of this Scope.
+	parentScope *Scope
+
+	// All the child scopes of this Scope.
+	childScopes []*Scope
+}
+
+func newScope() *Scope {
+	s := &Scope{
+		name:      "container",
```
abhinav, line 85: Two things:

1. Is there a reason we went with "container" as the name of the root node instead of empty string?
2. Since the name must always be set and is the only thing that changes, should this perhaps be a parameter (`newScope(name)`)?

scope.go
```diff
@@ -0,0 +1,270 @@
...
+		scopes[i], scopes[j] = scopes[j], scopes[i]
+	}
+	return scopes
+}
+
+func (s *Scope) appendLeafScopes(dest []*Scope) []*Scope {
+	dest = append(dest, s)
+	for _, cs := range s.childScopes {
+		dest = cs.appendLeafScopes(dest)
+	}
```
abhinav, line 136: I don't recall if we already discussed naming for this operation.
"leaf scopes" seems inaccurate; these are all nodes in the subtree rooted at this node.
Perhaps "appendSubtree" would be more appropriate?

scope.go
```diff
@@ -0,0 +1,270 @@
...
+
+func (s *Scope) getStoresFromRoot() []containerStore {
+	var stores []containerStore
+	for s := s; s != nil; s = s.parentScope {
+		stores = append(stores, s)
+	}
+	for i, j := 0, len(stores)-1; i < j; i, j = i+1, j-1 {
+		stores[i], stores[j] = stores[j], stores[i]
+	}
+	return stores
```
abhinav, line 148: This is getScopesFromRoot but with a different type.
We can probably just call getScopesFromRoot, and then build a slice casting those elements:

```
scopes := getScopesFromRoot()
stores := make([]containerStore, len(scope))
for i, s := range scopes {
  stores[i] = s
}
```

scope.go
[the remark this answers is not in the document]
abhinav: FWIW bytes.Buffer.Write [always returns a nil error](https://pkg.go.dev/bytes#Buffer.Write), so checking errors here would be unnecessary. I don't personally use errcheck much, but it will also ignore this case when the writer is known to be a bytes.Buffer at compile time.

PR #306, opened 2021-12-22
Add Export option to Provide

sywhang:
This adds `Export` option to `Provide`, which specifies that the provided constructor should be made available to all the Scopes in the same scope tree.

For example:
```
c := New()
s1 := c.Scope("child 1")
s2 := c.Scope("child 2")
s2.Provide(func() *A { ... })
c.Invoke(func(a *A) { ... }) // errors
s1.Invoke(func(a *A) { ...}) // errors
```
will error out on `Invoke` because constructor providing `*A` is provided to the `s2` only. 

With `Export` option, the child can provide this to all the Scopes in effect:
```
c := New()
s1 := c.Scope("child 1")
s2 := c.Scope("child 2")
s2.Provide(func() *A { ... }, Export(true))
c.Invoke(func(a *A) { ... })  // works!
s1.Invoke(func(a *A) { ... }) // works!
```
The implementation is quite simple - if this option is set, we provide it to the root Container. That provides the constructor visibility to all the scopes in effect.

### DIFF
--- a/provide.go
+++ b/provide.go
@@ -304,7 +304,18 @@ func (o provideLocationOption) applyProvideOption(opts *provideOptions) {
 
 // Export is a ProvideOption which specifies that the provided function should
 // be made available to all Scopes available in the application, regardless
-// of from which Scope it was provided.
+// of from which Scope it was provided. By default, it is false.
+//
+// For example,
+//  c := New()
+//  s1 := c.Scope("child 1")
+//  s2:= c.Scope("child 2")
+//  s1.Provide(func() *bytes.Buffer { ... })
+// does not allow the constructor returning *bytes.Buffer to be made available to
+// the root Container c or its sibling Scope s2.
+//
+// With Export, you can make this constructor available to all the Scopes:
+//  s1.Provide(func() *bytes.Buffer { ... }, Export(true))
 func Export(export bool) ProvideOption {
 	return provideExportOption{exported: export}
 }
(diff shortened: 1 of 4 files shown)


moisesvega (approved):

provide.go
```diff
@@ -301,6 +302,34 @@ func (o provideLocationOption) applyProvideOption(opts *provideOptions) {
 	opts.Location = o.loc
 }
 
+// Export is a ProvideOption which specifies that the provided function should
+// be made available to all Scopes available in the application, regardless
+// of from which Scope it was provided. By default, it is false.
```
r-hang, line 307: nit: regardless of which Scope it was provided from.

provide_test.go
```diff
@@ -86,3 +86,11 @@ func TestLocationForPCString(t *testing.T) {
 	opt := LocationForPC(reflect.ValueOf(func() {}).Pointer())
 	assert.Contains(t, fmt.Sprint(opt), `LocationForPC("go.uber.org/dig".TestLocationForPCString.func1 `)
 }
+
+func TestExportString(t *testing.T) {
+	opt1 := Export(true)
```
r-hang, line 91: optional: Could inline into fmt.Sprint(...)

r-hang (approved):

abhinav (commented):
LGTM. Pretty straightforward.

PR #311, opened 2022-01-05
Misc. fixes from #305

sywhang:
This contains some miscellaneous fixes from additional feedback
from #305. Notably:

- Make provider resolution for parameter building iterate the other way around (from self to root) and break once a provider is found.
- Rename appendLeafScopes to appendSubscopes.
- Rename getScopesFromRoot to ancestors
- Change getStoresFromRoot to use storesToRoot.

### DIFF
--- a/param.go
+++ b/param.go
@@ -146,7 +146,6 @@ func (pl paramList) Build(containerStore) (reflect.Value, error) {
 // to the underlying constructor.
 func (pl paramList) BuildList(c containerStore) ([]reflect.Value, error) {
 	args := make([]reflect.Value, len(pl.Params))
-	argsBuilt := make([]bool, len(pl.Params))
 	allContainers := c.getStoresFromRoot()
 	for i, p := range pl.Params {
 		var err error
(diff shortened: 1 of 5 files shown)


param.go
```diff
@@ -146,36 +146,26 @@ func (pl paramList) Build(containerStore) (reflect.Value, error) {
 // to the underlying constructor.
 func (pl paramList) BuildList(c containerStore) ([]reflect.Value, error) {
 	args := make([]reflect.Value, len(pl.Params))
-	argsBuilt := make([]bool, len(pl.Params))
 	allContainers := c.getStoresFromRoot()
 	for i, p := range pl.Params {
-		var err error
-		var arg reflect.Value
 		// iterate through the tree path of scopes.
 		for _, c := range allContainers {
```
abhinav, line 152: Just confirming: We do this in parent->child order. Is the reason for this that we want the value to be instantiated in the parent and shared across all descendants if possible?

param.go
```diff
@@ -146,36 +146,26 @@ func (pl paramList) Build(containerStore) (reflect.Value, error) {
...
-			if arg, err = p.Build(c); err == nil {
+			arg, err := p.Build(c)
+			if err == nil {
 				args[i] = arg
```
abhinav, line 155: Do we need to `break` the allContainers loop here to stop instantiating the value in the children?

```
containerLoop:
for _, c := range allContainers {
  // ..
  if err == nil {
    args[i] = arg
    break containerLoop
```

Otherwise this will call the child's constructors for that type as well, right?

---

While we're at it, is this valid:

```
root := New()
root.Provide(func() Foo { ... })

child := root.Child("foo")
child.Provide(func() Foo { ... })
```

If so, does an `Invoke(func(Foo))` mean we want the child's version?
And if that's the case, the scope iteration order should probably be reversed?

abhinav (approved):
I would update the commit message while merging since this is a bit different now.

PR #312, opened 2022-01-05
param: Search scopes during Build, support value groups

abhinav:
This significantly changes how we instantiate dependencies for a
constructor and adds support for value groups.

Previously, for each parameter of a constructor, we attempted to
instantiate the value in every scope until one succeeded without
dependency errors. Effectively:

    for param in constructor
        for scope in scopes_to_root
            err = scope.instantiate(param)
            if err != missing_dependencies {
                return err
            }

This made the behavior less certain because we were relying on fuzzily
matching errors.

With this change, we search the container for constructors as needed
for singular parameters (paramSingle) and for value groups
(paramGroupedSlice). Roughly:

    for param in constructor
        switch param {
        case paramSingle:
            for scope in scopes_to_root {
                if scope.provides(param) {
                    found = true
                }
            }
            // ...
        
        case paramGroupedSlice:
            for scope in scopes_to_root {
                providers.append(scope.providers_for(param))
            }
        
        default:
            // ...
        }

This is cleaner because it works by searching the container for what it
needs rather than trying and letting it fail.

This includes a subtle change to constructorNode.Call:
previously, when a constructor ran, its results were stored in the
containerStore that it used to instantiate its dependencies.
Now, results are stored in the containerStore (Scope) to which that
constructor was provided. This works because the leaf param types
(paramSingle and paramGroupedSlice) will search the scope tree to find
the value.

### DIFF
--- a/constructor.go
+++ b/constructor.go
@@ -151,7 +151,7 @@ func (n *constructorNode) Call(c containerStore) error {
 
 	// Commit the result to the original container that this constructor
 	// was supplied to. The provided constructor is only used for a view of
-	// the rest of the graph to instantiate the dependnecies of this
+	// the rest of the graph to instantiate the dependencies of this
 	// container.
 	receiver.Commit(n.s)
 	n.called = true
--- a/constructor.go
+++ b/constructor.go
@@ -148,7 +148,12 @@ func (n *constructorNode) Call(c containerStore) error {
 	if err := n.resultList.ExtractList(receiver, results); err != nil {
 		return errConstructorFailed{Func: n.location, Reason: err}
 	}
-	receiver.Commit(c)
+
+	// Commit the result to the original container that this constructor
+	// was supplied to. The provided constructor is only used for a view of
+	// the rest of the graph to instantiate the dependnecies of this
+	// container.
+	receiver.Commit(n.s)
 	n.called = true
 
 	return nil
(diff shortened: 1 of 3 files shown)


constructor.go
```diff
@@ -148,7 +148,12 @@ func (n *constructorNode) Call(c containerStore) error {
 	if err := n.resultList.ExtractList(receiver, results); err != nil {
 		return errConstructorFailed{Func: n.location, Reason: err}
 	}
-	receiver.Commit(c)
+
+	// Commit the result to the original container that this constructor
+	// was supplied to. The provided constructor is only used for a view of
+	// the rest of the graph to instantiate the dependnecies of this
```
sywhang, line 154:
```suggestion
	// the rest of the graph to instantiate the dependencies of this
```

param.go
```diff
@@ -146,28 +146,11 @@ func (pl paramList) Build(containerStore) (reflect.Value, error) {
...
-			// this means some constructor returned an error that must
-			// be reported.
-			_, isErrMissingTypes := err.(errMissingTypes)
-			_, isErrMissingDeps := err.(errMissingDependencies)
-			if err != nil && !isErrMissingTypes && !isErrMissingDeps {
-				return nil, err
-			}
+		var err error
+		args[i], err = p.Build(c)
+		if err != nil {
```
sywhang, line 152:
```suggestion
		if args[i], err := p.Build(c); err != nil {
```

sywhang (approved):
couple of nits but lgtm otherwise

param.go
abhinav: Can't do this because we're assigning to a slice element. `:=` only works if we create a new variable.

PR #313, opened 2022-01-07
Add Scope.Decorate

sywhang:
This adds `Scope.Decorate` which lets the user decorate an existing provider for a given type by overriding that with another provider. 

For example,

```
c := dig.New()

c.Provide(func() *Logger {
  return Logger {
    Name: "Default",
  }
})

c.Decorate(func(l *Logger) *Logger {
  return Logger {
    Name: "Decorated",
  }
})
```

the code snippet above shows a provider that injects a `*Logger` type into the top-level scope. Then it adds a decorator which takes in the `*Logger` type and replaces its name with something else. 

Once that has been done, the following `Invoke` will get the decorated name:
```
dig.Invoke(func(l *Logger) {
  l.Log(l.Name) // will log "Decorated"
})
```

Decorations are limited to its scope only. i.e. a child's decorator does not affect the parent scope or any of the ancestor scopes. A parent's decorator does affect the child and the descendant scopes.

One limitation: a Scope can only decorate a type once. It is possible, however, to create a child scope and then decorate once more in the child scope. In such case, the parent's decorator will be executed prior to the child's decorator. 

In terms of implementation, an additional set of types were added. Namely:

- `decorator`: This is an interface that abstract a decorator that was provided to a Scope.
- `decoratorNode`: This implements the `decorator` interface.
- `DecorateOption`: Functional options for `Scope.Decorate`. Unimplemented.

In addition, `Scope` also has new fields to keep track of all the decorators and decorated values/value groups that were injected in the scope directly. The way they work is quite similar to providers and values/value groups list. Basically, decorators are analogous to providers (constructors), and it keeps a separated list of decorated values and value groups in addition to provider-injected values and value groups.


shirchen:
took a first pass with a shallow walk through code. going to go deeper on next pass through.

decorate_test.go
```diff
@@ -0,0 +1,463 @@
...
+		require.NoError(t, c.Provide(func() string { return "b" }, dig.Group("values")))
+
+		require.NoError(t, c.Decorate(func(p Param) Result {
+			return Result{
+				Values: append(p.Values, "c"),
+			}
+		}))
+
+		assert.NoError(t, c.Invoke(func(p Param) {
+			assert.Equal(t, 3, len(p.Values))
```
abhinav: For this one, should we check if A is nil, and assign it a value for the invoke?

decorate_test.go
```diff
@@ -0,0 +1,463 @@
...
+			dig.In
+
+			Animals []string `group:"animals"`
+		}
+
+		type Result struct {
+			dig.Out
+
+			Animals []string `group:"animals"`
+		}
```
abhinav, line 176: This is pretty cool.
We'll need to document under Decorate that for value groups, the result must be a `[]T` as well.

decorate.go
```diff
@@ -0,0 +1,254 @@
...
+	if err != nil {
+		return errArgumentsFailed{
+			Func:   n.location,
+			Reason: err,
+		}
+	}
+
+	results := reflect.ValueOf(n.dcor).Call(args)
+	if err != nil {
+		return nil
```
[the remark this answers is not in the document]
sywhang: oops, that was dead code. thanks for catching that, removing this.

scope.go
```diff
@@ -82,11 +91,14 @@ type Scope struct {
...
-		invokerFn: defaultInvoker,
-		rand:      rand.New(rand.NewSource(time.Now().UnixNano())),
+		providers:       make(map[key][]*constructorNode),
+		decorators:      make(map[key][]*decoratorNode),
+		values:          make(map[key]reflect.Value),
+		decoratedValues: make(map[key]reflect.Value),
+		groups:          make(map[key][]reflect.Value),
+		decoratedGroups: make(map[key][]reflect.Value),
+		invokerFn:       defaultInvoker,
+		rand:            rand.New(rand.NewSource(time.Now().UnixNano())),
```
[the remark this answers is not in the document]
sywhang: The randomness here is here just so that people don't rely on the ordering of grouped values when they use Dig; I don't think it not being "random enough" would matter that much, but we could change it. Thoughts?

param.go
```diff
@@ -207,7 +217,34 @@ func (ps paramSingle) getValue(c containerStore) (reflect.Value, bool) {
 	return _noValue, false
 }
 
-func (ps paramSingle) Build(c containerStore) (reflect.Value, error) {
+func (ps paramSingle) Build(c containerStore, decorating bool) (reflect.Value, error) {
+	if !decorating {
```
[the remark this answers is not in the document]
sywhang: 👍

decorate_test.go
```diff
@@ -0,0 +1,463 @@
...
+				result = append(result, "good "+val)
+			}
+			return DecorateOut{
+				Values: result,
+			}
+		})
+
+		child.Invoke(func(i InvokeIn) {
+			assert.ElementsMatch(t, []string{"happy dog", "happy cat", "good dog", "good cat"}, i.Values)
+		})
```
abhinav, line 304: Each level of decorate on a value group is an append, not a replace?
That's surprising behavior, IMO.
Should we do a replace, and if the user wants to append, they can copy items from the original slice too?

(Also, I guess for the other value group case above, we should also use ElementsMatch because this was not obvious with just the Contains check.)

param.go
```diff
@@ -49,7 +49,7 @@ type param interface {
 	// Container.
 	//
 	// This MAY panic if the param does not produce a single value.
-	Build(containerStore) (reflect.Value, error)
+	Build(containerStore, bool) (reflect.Value, error)
```
abhinav, line 52: Consider naming the parameter here for documentation purposes.

param.go
```diff
@@ -207,7 +217,44 @@ func (ps paramSingle) getValue(c containerStore) (reflect.Value, bool) {
...
+			}
+			return _noValue, errParamSingleFailed{
+				CtorID: 1,
+				Key:    key{t: ps.Type, name: ps.Name},
+				Reason: err,
+			}, true
+		}
+		v, _ = c.getDecoratedValue(ps.Name, ps.Type)
+	}
+	return v, err, found
```
abhinav: Can reduce nesting here by inverting the if condition
```suggestion
	if len(decorators) == 0 {
		return _noValue, nil, false
	}

	for _, d := range decorators {
		err := d.Call(c)
		if err == nil {
			continue
		}
		if _, ok := err.(errMissingDependencies); ok && ps.Optional {
			v, err = reflect.Zero(ps.Type), nil
		}
		return _noValue, errParamSingleFailed{
			CtorID: 1,
			Key:    key{t: ps.Type, name: ps.Name},
			Reason: err,
		}, true
	}
	v, _ = c.getDecoratedValue(ps.Name, ps.Type)
	return v, err, true
```

param.go
```diff
@@ -241,7 +288,7 @@ func (ps paramSingle) Build(c containerStore) (reflect.Value, error) {
 
 		// If we're missing dependencies but the parameter itself is optional,
 		// we can just move on.
-		if _, ok := err.(errMissingDependencies); ok && ps.Optional {
+		if errors.As(err, new(errMissingDependencies)) && ps.Optional {
```
abhinav, line 292: You sure this will work? Dig's internal errors don't implement `Unwrap`.

param.go
```diff
@@ -485,23 +532,85 @@ func newParamGroupedSlice(f reflect.StructField, c containerStore) (paramGrouped
...
+	}
 
+	// Check if we have decorated values
+	decoratedItems := pt.getDecoratedValues(c)
+	if len(decoratedItems) != 0 {
+		result := reflect.MakeSlice(pt.Type, 0, len(decoratedItems))
+		for _, item := range decoratedItems {
+			result = reflect.Append(result, item)
+		}
+		return result, nil
```
abhinav, line 604: I think maybe getDecoratedValues should return an `ok bool` because an empty slice should be a valid replacement via decoration?

scope.go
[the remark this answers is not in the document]
abhinav: For security purposes, crypto/rand is indeed the thing to use.
To shuffle the Dig value groups, pseudo-randomness from math/rand is enough.
(Slightly related: crypto/rand will be really slow for this.)

result.go
```diff
@@ -479,12 +483,17 @@ func newResultGrouped(f reflect.StructField) (resultGrouped, error) {
...
 		cw.submitGroupedValue(rt.Group, rt.Type, v)
 		return
 	}
 	for i := 0; i < v.Len(); i++ {
-		cw.submitGroupedValue(rt.Group, rt.Type, v.Index(i))
+		if !decorated {
+			cw.submitGroupedValue(rt.Group, rt.Type, v.Index(i))
+		} else {
+			cw.submitDecoratedGroupedValue(rt.Group, rt.Type.Elem(), v.Index(i))
+		}
```
abhinav, line 497: I think maybe this is where—if we're decorating—instead of submitting individual values, we submit the entire decorated slice, so that scope level knows whether it decorated that group.

This will allow replacing the decorated value groups wholesale.

param.go
```diff
@@ -485,23 +533,85 @@ func newParamGroupedSlice(f reflect.StructField, c containerStore) (paramGrouped
 	return pg, nil
 }
 
-func (pt paramGroupedSlice) Build(c containerStore) (reflect.Value, error) {
-	var itemCount int
+func (pt paramGroupedSlice) getDecoratedValues(c containerStore) []reflect.Value {
+	var items []reflect.Value
+	stores := c.storesToRoot()
+	for _, c := range stores {
```
r-hang, line 539: possible nit: could inline: for _, c := range c.storesToRoot()

param.go
```diff
@@ -485,23 +533,85 @@ func newParamGroupedSlice(f reflect.StructField, c containerStore) (paramGrouped
...
+		items = append(items, c.getDecoratedValueGroup(pt.Group, pt.Type.Elem())...)
+	}
+	return items
+}
+
+// searches the given container and its parent for matching group decorators
```
r-hang, line 545: :s/searches/search/g
:s/calls/call/g

possible suggestion: If any decorators return an error, that error is returned immediately. If all decorators succeed, nil is returned.

param.go
```diff
@@ -485,23 +533,85 @@ func newParamGroupedSlice(f reflect.StructField, c containerStore) (paramGrouped
...
+// and calls them to commit values. If any of the decorators returns an error,
+// it returns that error immediately. If all decorators succeeds, it returns nil.
+func (pt paramGroupedSlice) callGroupDecorators(c containerStore) error {
+	stores := c.storesToRoot()
+	for _, c := range stores {
```
r-hang, line 550: same possible nit from above about inlining storesToRoot().

param.go
```diff
@@ -485,23 +533,85 @@ func newParamGroupedSlice(f reflect.StructField, c containerStore) (paramGrouped
...
+					Key:    key{group: pt.Group, t: pt.Type.Elem()},
+					Reason: err,
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// searches the given container and its parent for matching group providers and
```
r-hang, line 564: suggestion: Search the given container and its parent for matching group providers and call them to commit values.
Return the number of providers called and an a non-nil error from the first provided, if an error is encountered.

decorate.go
```diff
@@ -0,0 +1,254 @@
...
+func (n *decoratorNode) ID() dot.CtorID { return n.id }
+
+// DecorateOption modifies the default behavior of Provide.
+// Currently there is no implementation of it yet.
+type DecorateOption interface {
+	applyDecorateOption(*decorateOptions)
+}
+
+type decorateOptions struct {
+}
```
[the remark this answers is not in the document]
sywhang: @r-hang yes.

constructor.go
```diff
@@ -135,7 +136,7 @@ func (n *constructorNode) Call(c containerStore) error {
 		}
 	}
 
-	args, err := n.paramList.BuildList(c)
+	args, err := n.paramList.BuildList(c, false)
```
moisesvega, line 139:
```suggestion
	args, err := n.paramList.BuildList(c, false /* decorating */)
```
per https://github.com/uber-go/guide/blob/master/style.md#avoid-naked-parameters

invoke.go
```diff
@@ -82,7 +82,7 @@ func (s *Scope) Invoke(function interface{}, opts ...InvokeOption) error {
 		s.isVerifiedAcyclic = true
 	}
 
-	args, err := pl.BuildList(s)
+	args, err := pl.BuildList(s, false)
```
moisesvega, line 85:
```suggestion
	args, err := pl.BuildList(s, false /* decorating */)
```
per https://github.com/uber-go/guide/blob/master/style.md#avoid-naked-parameters

constructor.go
```diff
@@ -145,7 +146,7 @@ func (n *constructorNode) Call(c containerStore) error {
 
 	receiver := newStagingContainerWriter()
 	results := c.invoker()(reflect.ValueOf(n.ctor), args)
-	if err := n.resultList.ExtractList(receiver, results); err != nil {
+	if err := n.resultList.ExtractList(receiver, false, results); err != nil {
```
moisesvega, line 149:
```suggestion
	if err := n.resultList.ExtractList(receiver, false /* decorated */, results); err != nil {
```

decorate.go
```diff
@@ -0,0 +1,213 @@
...
+	args, err := n.params.BuildList(n.s, true /* decorating */)
+	if err != nil {
+		return errArgumentsFailed{
+			Func:   n.location,
+			Reason: err,
+		}
+	}
+
+	results := reflect.ValueOf(n.dcor).Call(args)
+	if err := n.results.ExtractList(n.s, true, results); err != nil {
```
moisesvega, line 110:
```suggestion
	if err := n.results.ExtractList(n.s, true /* decorated */, results); err != nil {
```

decorate.go
```diff
@@ -0,0 +1,250 @@
...
+		return err
+	}
+	n.called = true
```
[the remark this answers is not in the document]
sywhang: I think it's technically more correct to set it here once we confirmed there is no error. 

For example, you could imagine doing something like trying to decorate with a function whose dependencies haven't been met - in such case, line 94 or 101 will fail. But we shouldn't mark this decorator as "called" because it'll simply return without attempting to call it again. 

Basically marking called as true earlier than here would mean the following scenario would break (when it's not supposed to):
1. Provide only part of the dependency of a decorator function f. 
2. Try to decorate using f. 
3. Call something that uses f produces. (and run into an error)
4. Provide the remainder of the dependency of f.
5. Call that something again from step 3. 
6. We should expect that second call to succeed.

decorate_test.go
sywhang: Fixed!

container.go
```diff
@@ -111,6 +124,14 @@ type containerStore interface {
 	// type across all the Scopes that are in effect of this containerStore.
 	getAllValueProviders(name string, t reflect.Type) []provider
 
+	// Returns the decorators that can produce values for the given name and
+	// type.
```
shirchen, line 128: nit: for consistency with comment on line 112, since we are iterating over a map, returned providers are probably also undefined? or do we not care about ordering here?

decorate.go
```diff
@@ -0,0 +1,213 @@
...
+	return nil
+}
+
+func (n *decoratorNode) ID() dot.CtorID { return n.id }
+
+// DecorateOption modifies the default behavior of Provide.
+// Currently there is no implementation of it yet.
```
shirchen, line 120: nit: change to Currently,

decorate.go
```diff
@@ -0,0 +1,213 @@
...
+// Decorate provides a decorator for a type that has already been provided in the Scope.
+//
+// Similar to Provide, Decorate takes in a function with zero or more dependencies and one
+// or more results. Decorate can be used to modify a type that was already introduced to the
+// Scope, or completely replace it with a new object.
+//
+// For example,
+//  s.Decorate(func(log *zap.Logger) *zap.Logger {
+//    return log.Named("myapp")
+//  }
```
shirchen, line 141: dangling ) here and on 148

invoke.go
```diff
@@ -124,13 +124,14 @@ func findMissingDependencies(c containerStore, params ...param) []paramSingle {
 	for _, param := range params {
 		switch p := param.(type) {
 		case paramSingle:
-			if ns := c.getAllValueProviders(p.Name, p.Type); len(ns) == 0 && !p.Optional {
+			allProviders := c.getAllValueProviders(p.Name, p.Type)
+			_, hasDecoratedValue := c.getDecoratedValue(p.Name, p.Type)
+			if len(allProviders) == 0 && !hasDecoratedValue && !p.Optional {
```
shirchen, line 129: i had to pause and think about why we want to exempt list of deps if they are decorated. can u maybe explain in code and here about the special case.

param.go
```diff
@@ -49,7 +49,7 @@ type param interface {
 	// Container.
 	//
 	// This MAY panic if the param does not produce a single value.
-	Build(containerStore) (reflect.Value, error)
+	Build(store containerStore, decorating bool) (reflect.Value, error)
```
shirchen, line 52: Let's fix docstring and make it `//Build this dependency...`

decorate_test.go
```diff
@@ -0,0 +1,509 @@
...
+		c.RequireInvoke(func(a *A) {
+			assert.Equal(t, "A", a.name, "expected name to not be decorated yet.")
+		})
+
+		c.RequireDecorate(func(a *A) *A { return &A{name: a.name + "'"} })
+
+		c.RequireInvoke(func(a *A) {
+			assert.Equal(t, "A'", a.name, "expected name to equal decorated name.")
+		})
+
```
shirchen, line 52: nit: these newlines are fine, but they wouldnt pass https://github.com/mvdan/gofumpt

decorate_test.go
```diff
@@ -0,0 +1,509 @@
...
+			assert.Equal(t, "A'", a.name, "expected decorated name in parent")
+		})
+
+		child.RequireInvoke(func(a *A) {
+			assert.Equal(t, "A''", a.name, "expected double-decorated name in child")
+		})
+
+		sibling := c.Scope("sibling")
+		grandchild := child.Scope("grandchild")
+		sibling.Invoke(func(a *A) {
```
shirchen, line 117: let's check for no errors on 117, 120, 282, 292 and 302

decorate_test.go
```diff
@@ -0,0 +1,509 @@
...
+		}, dig.Group("value"))
+
+		c.RequireDecorate(func(i DecorateIn) DecorateOut {
+			return DecorateOut{Values: i.Values}
+		})
+
+		err := c.Invoke(func(c InvokeIn) {})
+		require.Error(t, err, "expected one of the group providers for a decorator to fail")
+		assert.Contains(t, err.Error(), `could not build value group string[group="decoratedVal"]`)
+		assert.Contains(t, err.Error(), `received non-nil error from function "go.uber.org/dig_test".TestDecorateFailure.func4.3`)
```
shirchen, line 423: any way to make this under 100, or whatever we have set as max column width?

decorate_test.go
```diff
@@ -0,0 +1,509 @@
...
+			dig.In
+
+			Value string `name:"val"`
+		}
+
+		c := digtest.New(t)
+		c.RequireDecorate(func(p Param) string { return p.Value })
+		err := c.Invoke(func(s string) {})
+		require.Error(t, err, "expected missing dep check to fail the decorator")
+		assert.Contains(t, err.Error(), `missing dependencies for function "go.uber.org/dig_test".TestDecorateFailure.func7.2`)
```
shirchen, line 470: a bit too wide here too.

result.go
```diff
@@ -304,11 +304,15 @@ func (rs resultSingle) DotResult() []*dot.Result {
 	return dotResults
 }
 
-func (rs resultSingle) Extract(cw containerWriter, v reflect.Value) {
-	cw.setValue(rs.Name, rs.Type, v)
+func (rs resultSingle) Extract(cw containerWriter, decorated bool, v reflect.Value) {
+	if decorated {
+		cw.setDecoratedValue(rs.Name, rs.Type, v)
+	} else {
+		cw.setValue(rs.Name, rs.Type, v)
```
moisesvega, line 311: We can do something like this to reduce nesting. https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

```suggestion
func (rs resultSingle) Extract(cw containerWriter, decorated bool, v reflect.Value) {
	if decorated {
		return cw.setDecoratedValue(rs.Name, rs.Type, v)
	}
	cw.setValue(rs.Name, rs.Type, v)
```

shirchen (commented):
this looks great and the abundance of unit tests is fantastic.

result.go
```diff
@@ -479,12 +483,18 @@ func newResultGrouped(f reflect.StructField) (resultGrouped, error) {
...
+	if !decorated && !rt.Flatten {
 		cw.submitGroupedValue(rt.Group, rt.Type, v)
 		return
 	}
-	for i := 0; i < v.Len(); i++ {
-		cw.submitGroupedValue(rt.Group, rt.Type, v.Index(i))
+
+	if decorated {
+		cw.submitDecoratedGroupedValue(rt.Group, rt.Type, v)
+	} else {
```
moisesvega, line 495: We can return early to reduce nesting. 
```suggestion
	if decorated {
		cw.submitDecoratedGroupedValue(rt.Group, rt.Type, v)
		return 
	}
```

scope.go
```diff
@@ -184,6 +215,23 @@ func (s *Scope) getGroupProviders(name string, t reflect.Type) []provider {
...
+func (s *Scope) getValueDecorators(name string, t reflect.Type) []decorator {
+	return s.getDecorators(key{name: name, t: t})
+}
+
+func (s *Scope) getGroupDecorators(name string, t reflect.Type) []decorator {
+	return s.getDecorators(key{group: name, t: t})
+}
+
+func (s *Scope) getDecorators(k key) []decorator {
+	nodes := s.decorators[k]
```
moisesvega, line 227: Maybe we should guard ourself and make sure the key exist before continue.

container.go
sywhang: The ordering here isn't undefined since it's iterating over a single entry in a map. The value of that map is a list so the ordering will be consistent.

decorate_test.go
```diff
@@ -0,0 +1,509 @@
...
+		}
+
+		type Result struct {
+			dig.Out
+
+			Animals []string `group:"animals"`
+		}
+
+		c := digtest.New(t)
+		c.RequireProvide(func() string { return "dog" }, dig.Group("animals"))
```
r-hang, line 179: humorous non-suggestion: I would expect one of these to be a gopher.

param.go
```diff
@@ -197,7 +197,17 @@ func (ps paramSingle) String() string {
 	return fmt.Sprintf("%v[%v]", ps.Type, strings.Join(opts, ", "))
 }
 
-// searches the given container and its parent for a matching value.
+// search the given container and its parent for decorated values.
```
r-hang, line 200:
```suggestion
// search the given container and its ancestors for a decorated value.
```

param.go
```diff
@@ -197,7 +197,17 @@ func (ps paramSingle) String() string {
...
+func (ps paramSingle) getDecoratedValue(c containerStore) (reflect.Value, bool) {
+	for _, c := range c.storesToRoot() {
+		if v, ok := c.getDecoratedValue(ps.Name, ps.Type); ok {
+			return v, ok
+		}
+	}
+	return _noValue, false
+}
+
+// search the given container and its parent for a matching value.
```
r-hang, line 210:
```suggestion
// search the given container and its ancestors for a matching value.
```

result.go
r-hang, line 495: Can we reduce nesting even further by dropping the else statement for an early return?
```
...
if decorated {
    cw.submitDecoratedGroupedValue(rt.Group, rt.Type, v)
    return 
}
for i := 0; i < v.Len(); i++ {
  cw.submitGroupedValue(rt.Group, rt.Type, v.Index(i))
}
```

moisesvega (approved):

decorate_test.go
```diff
@@ -0,0 +1,525 @@
...
+		c.RequireDecorate(func() A {
+			return A{From: "decorator"}
+		})
+
+		c.RequireInvoke(func(a A) {
+			assert.Equal(t, a.From, "decorator", "value should be from decorator")
+		})
+	})
+
+	t.Run("group value decorator from parent and child", func(t *testing.T) {
```
abhinav, line 250: 🎉

abhinav (approved):

PR #315, opened 2022-01-27
Parallel container startup with deferred values

xandris:
Hi dig maintainers,

My organization made an internal tool that has many different startup tasks that all produce a unique Go type. Before dig/fx, some tasks wrote to a shared struct, some tasks read from it, and some did both. This meant having to carefully order functions, keeping in mind these implicit dependencies hidden in func bodies. I'm busy removing this shared struct and rewriting each task as a dig/fx provider and letting dig order things automatically. It's wonderful! Can't thank you enough.

However—and sorry for frontloading this so much—our existing code, poor quality as it is, does perform many slow tasks in parallel. Forcing these to happen in serial would push our execution time over our execution frequency. But dependency injection with dig is a perfect match with our codebase, so I would really like to work something out.

I don't believe the core of dig is incompatible with parallel execution. To prove this to myself, as a first pass I made paramList use an Errgroup to call all its children and wait for them to finish, and I made the store maps sync.Map. Where two goroutines reached the same constructorNode, I put a mutex to protect the `called` field. It worked but felt too simplistic and hard to control.

The approach I settled on lets all dig code run in one goroutine, and user-provided constructors can either run in the same goroutine (default) or in pooled/ephemeral goroutines. This keeps mutexes out of dig's data structures and lets users decide how much concurrency they want.

I tried this several ways, but this pull request is the cleanest way I found. It creates a new `deferred` type params and constructorNodes return. A `deferred` can be observed and resolved like a very simple Promise. Since params and constructors pass values indirectly via a Scope, `deferred` doesn't need much state.

I would very much like to get concurrent initialization into dig by any means; it doesn't have to be this pull request. I'm willing to any changes you deem necessary; let's work together!

Refs GO-1191


container.go
```diff
@@ -208,6 +211,29 @@ func dryInvoker(fn reflect.Value, _ []reflect.Value) []reflect.Value {
...
+
+// MaxConcurrency run constructors in this container with a fixed pool of executor
+// goroutines. max is the number of goroutines to start.
+func MaxConcurrency(max int) Option {
+	return maxConcurrencyOption(max)
+}
+
+func (m maxConcurrencyOption) applyOption(container *Container) {
+	container.scope.sched = &parallelScheduler{concurrency: int(m)}
+}
```
abhinav, line 224: I'm flexible on this, but how do we feel about merging the two options into one.
MaxConcurrency's behavior can be:

    0: no concurrency (default)
    >0: fixed number of workers
    -1: unbounded number of workers

If we think "negative means unlimited" is smelly, we can keep it as-is.

CC @sywhang

dig_test.go
```diff
@@ -3566,3 +3567,91 @@ func TestEndToEndSuccessWithAliases(t *testing.T) {
...
+
+		type (
+			A int
+			B int
+			C int
+		)
+
+		var (
+			timer           = time.NewTimer(10 * time.Second)
+			max       int32 = 3
```
abhinav, line 3584: minor: We can define a `const Max = 3`; it'll have type "untyped int" and won't need to be cast to int32.

dig_test.go
```diff
@@ -3566,3 +3567,91 @@ func TestEndToEndSuccessWithAliases(t *testing.T) {
...
+			done            = make(chan struct{})
+			running   int32 = 0
+			waitForUs       = func() error {
+				if atomic.AddInt32(&running, 1) == max {
```
abhinav, line 3588: We typically use [go.uber.org/atomic](https://pkg.go.dev/go.uber.org/atomic) for atomics to prevent unintentional direct access to the atomic variable.

```
var running atomic.Int32

// ...

if running.Inc() == Max { ... }
```

abhinav (commented):
Hi @xandris,

Thanks for the contribution and the explanation! We would like to help find a good solution to your problem. We really appreciate the thoughtfulness you put into the PR.

- If dig was to support concurrent initialization, I really like the idea of keeping dig's internal state management in its own goroutine. Especially because we've put zero effort into making dig's state management safe for concurrent use.
- I like how clean the `deferred` implementation is.

In the past, we've been reluctant to add concurrency to Dig largely because we felt that it would add significant complexity and risk to the code. Our position was that since Dig relies heavily on reflection, there's no expectation of high performance from it, and anything that suggests otherwise should be avoided.

*However*, we discussed the problem you're experiencing and this PR, and I think we're happy for Dig to support concurrent initialization. This reduces the scope of the problem (and our concerns) to a much safer subset, and introduces minimal new API surface area.

That said, the timing of this change was a bit unfortunate. As you might have noticed, we're deep in the middle of refactoring and cleanup to support decorations and replacements in the container. Apologies for the conflicts, but if you don't mind resolving them, we can work with you to get this merged.

(I have included some preliminary review comments; I haven't had a chance to do a full review yet. I suspect we'll want to move deferred and scheduler to internal/ to unit test them independently.)

CC @sywhang @shirchen

xandris:
> That said, the timing of this change was a bit unfortunate. As you might have noticed, we're deep in the middle of refactoring and cleanup to support decorations and replacements in the container. Apologies for the conflicts, but if you don't mind resolving them, we can work with you to get this merged.

Roger! I've started working on the merge conflicts and I'll do a force push once those are resolved. I have some training to do today, but after that I'll start running review comments to ground. Very happy we can work on this together!

One comment I have for the decorators impl–not that anyone asked–is that there seems to be a lot of code duplication between constructors and decorators. There are limits to DRY just by Go's nature but I wonder if there's opportunities here...

abhinav:
> One comment I have for the decorators impl–not that anyone asked–is that there seems to be a lot of code duplication between constructors and decorators. There are limits to DRY just by Go's nature but I wonder if there's opportunities here...

Feedback is absolutely welcome! Yes, there is room for DRYing up there. The codebase has some amount of historical debt which we're aware of, so we've been trying to find a good balance between "refactor it all until it's nice and clean" and actually getting [`fx.Replace`](https://github.com/uber-go/fx/issues/653) working. Right now we're leaning towards deferring some of the non-essential clean-ups until afterwards.

sywhang:
> There are limits to DRY just by Go's nature but I wonder if there's opportunities here...

Definitely! Thanks for pointing that out. As @abhinav said, we're kind of pushing hard to get the fx.Replace feature out ASAP so there are definitely some debts being left around. We'll be revisiting those after we make fx-side changes.

container.go
xandris: That makes sense to me; I went ahead and implemented your suggestion. If max<0 proves controversial we can split it off to a separate option.

dig_test.go
xandris: Thanks, that's a language feature I tend to forget about!

dig_test.go
xandris: Roger, I was thinking about using that but didn't want to add another dependency unless you all were okay with it.

abhinav (commented):
Thanks for making the updates @xandris.

We want Dig to have this feature, but we're slightly reluctant to merge the promise-based approach right now; we want to make sure that this is the only clean way to do this (since there is otherwise a maintainability risk associated with it). Going over the code a couple times, my intuition agrees with you that this is the most straightforward way to keep Dig's state management separate from the initialization goroutines. However, we'd like to give this a deeper look than we've been able to.

Unfortunately—and this ties into the other changes you've been seeing in Dig and Fx—we're currently heads-down with getting fx.Replace ready, so can't investigate too much right now.

Further, the new fx.Replace feature comes with a fair amount of risk because of the amount of internal refactoring it required. Adding another thing that significantly alters the core of how Dig operates would further compound the risk of this release.

Would you be willing to pin to your branch of Dig while we make more time for this on our end?

xandris:
> Would you be willing to pin to your branch of Dig while we make more time for this on our end?

The risk assessment makes sense to me. Standing by; we can use module replacements until this is ready. What do you mean by "pin"?

abhinav:
Excellent! Again, thanks for the PR.

> What do you mean by "pin"?

Oops, I meant module replacements. "Pin to X" (as in, "pin to branch X" or "pin to version "Y") is what we've been using to generically refer to module replacements. I think maybe the phrase was more accurate before Go modules (with dep and glide) when we had to sometimes pin a package to a specific version to ensure it did not get upgraded.

sywhang:
Hi @xandris , apologies for the delayed review. We had a few other priorities to take care of, and I just found time to get back to review this PR. I'll spend some time this week to test this internally as well in our codebase. 

But before I start doing that, I see that this branch has gone out of sync with master by quite a bit, so let me first address the merge conflicts first. Thank you for the patience!

xandris:
It's been a minute. How are things?

PR #317, opened 2022-02-06
Fix decoration logic for paramSingle

### LINKED ISSUE #316: Decorators at parent level scope affects child only after an invoke at parent scope
Suppose:
```
type A struct {
  Name string
}


c := dig.New()
child := c.Scope("child")

c.Provide(func() *A { return &A{Name: "A") })
c.Decorate(func(a *A) *A { return &A{Name: a.Name + "'" }})
```

Then:
```
c.Invoke(func(a *A) { fmt.Println(a.Name) })
child.Invoke(func(a *A) { fmt.Println(a.Name) })
```

prints two lines of `A'`.

But if we do this instead:
```
child.Invoke(func(a *A) { fmt.Println(a.Name) })
```

It prints `A` instead of `A'`.

sywhang:
When building paramSingle using decorators, there was a descrepancy when
the invoking Scope does not have a decorator for the type but one of its
parent Scopes does. This resulted in the parent Scope having to Invoke
a function that uses the decorated type for the child to start seeing it
too.

This fixes paramSingle to search all effective scopes for a decorator
and invoke that instead of searching just the scope it was invoked from.

Fixes #316

### DIFF
--- a/decorate_test.go
+++ b/decorate_test.go
@@ -34,56 +34,95 @@ func TestDecorateSuccess(t *testing.T) {
 	t.Run("simple decorate without names or groups", func(t *testing.T) {
 		t.Parallel()
 		type A struct {
-			name string
+			Name string
 		}
 
 		c := digtest.New(t)
-		c.RequireProvide(func() *A { return &A{name: "A"} })
+		c.RequireProvide(func() *A { return &A{Name: "A"} })
 
 		c.RequireInvoke(func(a *A) {
-			assert.Equal(t, "A", a.name, "expected name to not be decorated yet.")
+			assert.Equal(t, "A", a.Name, "expected name to not be decorated yet.")
 		})
 
-		c.RequireDecorate(func(a *A) *A { return &A{name: a.name + "'"} })
+		c.RequireDecorate(func(a *A) *A { return &A{Name: a.Name + "'"} })
 
 		c.RequireInvoke(func(a *A) {
-			assert.Equal(t, "A'", a.name, "expected name to equal decorated name.")
+			assert.Equal(t, "A'", a.Name, "expected name to equal decorated name.")
 		})
 	})
 
 	t.Run("simple decorate a provider from child scope", func(t *testing.T) {
 		t.Parallel()
 		type A struct {
-			name string
+			Name string
 		}
 
 		c := digtest.New(t)
 		child := c.Scope("child")
-		child.RequireProvide(func() *A { return &A{name: "A"} }, dig.Export(true))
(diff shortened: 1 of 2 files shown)


decorate_test.go
```diff
@@ -34,56 +34,95 @@ func TestDecorateSuccess(t *testing.T) {
 	t.Run("simple decorate without names or groups", func(t *testing.T) {
 		t.Parallel()
 		type A struct {
-			name string
+			Name string
```
abhinav, line 37: curious why this was needed?

decorate_test.go
```diff
@@ -34,56 +34,95 @@ func TestDecorateSuccess(t *testing.T) {
...
+			Values []string `group:"values"`
+		}
+
+		c := digtest.New(t)
+		child := c.Scope("child")
+
+		c.RequireProvide(func() *A { return &A{Name: "A"} }, dig.Export(true))
+		c.RequireProvide(func() string { return "val1" }, dig.Export(true), dig.Group("values"))
+		c.RequireProvide(func() string { return "val2" }, dig.Export(true), dig.Group("values"))
+		c.RequireProvide(func() string { return "val3" }, dig.Export(true), dig.Group("values"))
```
abhinav, line 96: Is dig.Export needed on the root container?

param.go
```diff
@@ -217,10 +217,19 @@ func (ps paramSingle) getValue(c containerStore) (reflect.Value, bool) {
...
-// builds the parameter using decorators, if any. If there are no decorators associated
-// with this parameter, _noValue is returned.
+// builds the parameter using decorators in all scopes that affect the
+// current scope, if there are any. If there are multiple Scopes that decorates
+// this parameter, the closest one to the Scope that invoked this will be used.
+// If there are no decorators associated with this parameter, _noValue is returned.
 func (ps paramSingle) buildWithDecorators(c containerStore) (v reflect.Value, found bool, err error) {
-	decorators := c.getValueDecorators(ps.Name, ps.Type)
+	var decorators []decorator
+	var decoratingScope containerStore
```
abhinav, line 226: nit: group these
```suggestion
	var (
		decorators []decorator
		decoratingScope containerStore
	)
```

https://github.com/uber-go/guide/blob/master/style.md#group-similar-declarations

decorate_test.go
sywhang: It wasn't - I think a change I was playing around with locally got merged in with this branch. I undid these changes.

decorate_test.go
sywhang: it is not. Removed

PR #322, opened 2022-02-17
Fix Invoke order changing execution result in multiple decorations of same type

### LINKED ISSUE #321: Result of multiple decorations of the same type depends on invocation order
Consider: 

```go
type A struct {
	name string
}

c := dig.New()
child := c.Scope("child")
c.Provide(func() *A { return &A{name: "A"} })

c.Decorate(func(a *A) *A { return &A{name: a.name + "'"} })
child.Decorate(func(a *A) *A { return &A{name: a.name + "'"} })

c.Invoke(func(a *A) {
    fmt.Println(a.name)
})

child.Invoke(func(a *A) {
    fmt.Println(a.name)
})
```

This prints `A''` and `A'` respectively to the console.

Now let's say the invokes were reversed in order:
```go
child.Invoke(func(a *A) {
    fmt.Println(a.name)
})

c.Invoke(func(a *A) {
    fmt.Println(a.name)
})
```
That ends up printing `A'` and `A'` on the screen.

The order of invocation should not alter the committed value - this is likely a bug in the paramSingle.Build() logic where it's being too "lazy" with the lookups for decorated values.

sywhang:
This fixes the multiple decoration logic for paramSingle where there
are more than one scopes that contain decorators for a type. The lookup
logic only looked up one-level for the decorator. For example, if Scope S
has a decorator and its parent Sp also has a decorator for the same type, it never
executed Sp's decorator.

This caused the order of the Scopes in which a function that used such type
was Invoked from to change the execution result.

This PR addresses the issue by more diligently looking up the types by
traversing for more decorators until there is no more decorators found
in the path to the root Scope.

Also added several tests that cover this case.

Fixes #321 and https://github.com/uber-go/fx/issues/841
Internal Ref: GO-1228

### DIFF
--- a/decorate_test.go
+++ b/decorate_test.go
@@ -589,20 +589,20 @@ func TestMultipleDecorates(t *testing.T) {
 	t.Run("decorate same type from parent and child, invoke child first", func(t *testing.T) {
 		t.Parallel()
 		type A struct{ value int }
-		c := digtest.New(t)
-		child := c.Scope("child")
+		root := digtest.New(t)
+		child := root.Scope("child")
 
 		decorator := func(a A) A {
 			return A{value: a.value + 1}
 		}
-		c.RequireProvide(func() A { return A{value: 0} })
-		c.RequireDecorate(decorator)
+		root.RequireProvide(func() A { return A{value: 0} })
+		root.RequireDecorate(decorator)
 		child.RequireDecorate(decorator)
 
 		child.Invoke(func(a A) {
 			assert.Equal(t, 2, a.value)
 		})
-		c.Invoke(func(a A) {
+		root.Invoke(func(a A) {
 			assert.Equal(t, 1, a.value)
 		})
 	})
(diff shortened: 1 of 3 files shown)


decorate_test.go
```diff
@@ -586,6 +585,139 @@ func TestDecorateFailure(t *testing.T) {
...
+
+		// decorate +1 to each element in child
+		child.RequireDecorate(decorator)
+
+		// assert child sees 2, 3, 4
+		child.RequireInvoke(func(a A) {
+			assert.ElementsMatch(t, []int{2, 3, 4}, a.Values)
+		})
+
+		// assert parent sees 1, 2, 3
```
abhinav, line 669: nit: unnecessary comments
```suggestion
		child.RequireInvoke(func(a A) {
			assert.ElementsMatch(t, []int{2, 3, 4}, a.Values)
		})

```

abhinav (approved):

PR #325, opened 2022-02-24
Allow providing types that depend on decorated types from child modules

sywhang:
This fixes the case where a function that's Invoked depends on a type that's decorated. 

Specifically, if a type has a decorator in a child Scope, and there's another exported constructor consuming that type within the same Scope, and a function is invoked outside of that Scope, we still should see the decorated type.

Implementation-wise, this means that the constructors should know which Scope it was originally provided to, and invoke it from that Scope if and only if that Scope has a more "specific" Scope than the Scope it was invoked from. To make this comparison of "specific-ness" of the Scopes cheap, a new field was added to Scope to keep track of its distance from the root Container. A Scope with higher degree is the more "specific" Scope than one with a lower degree.

### DIFF
--- a/param.go
+++ b/param.go
@@ -299,13 +299,7 @@ func (ps paramSingle) Build(c containerStore, decorating bool) (reflect.Value, e
 	}
 
 	for _, n := range providers {
-		var err error
-		if n.Exported() {
-			err = n.Call(n.OrigScope())
-		} else {
-			err = n.Call(c)
-		}
-
+		err := n.Call(n.OrigScope())
 		if err == nil {
 			continue
 		}
--- a/param.go
+++ b/param.go
@@ -305,9 +305,9 @@ func (ps paramSingle) Build(c containerStore, decorating bool) (reflect.Value, e
 			// a "larger" scope.
 			// We use the more "specific" Scope to call.
 			if n.OrigScope().getDegree() > c.getDegree() {
-				n.Call(n.OrigScope())
+				err = n.Call(n.OrigScope())
 			} else {
-				n.Call(c)
+				err = n.Call(c)
 			}
 		} else {
 			err = n.Call(c)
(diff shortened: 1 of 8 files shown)


param.go
```diff
@@ -299,7 +299,13 @@ func (ps paramSingle) Build(c containerStore, decorating bool) (reflect.Value, e
 	}
 
 	for _, n := range providers {
-		err := n.Call(c)
+		var err error
+		if n.Exported() {
+			err = n.Call(n.OrigScope())
+		} else {
+			err = n.Call(c)
+		}
```
abhinav, line 307: Could this just be `n.Call(n.OrigScope())` unconditionally? Unless I'm misunderstanding, either `Exported() == true` or `OrigScope() == c`.


```suggestion
		err := n.Call(n.OrigScope())
```

As an experiment, I tried this in a local checkout and only this test started failing:
https://github.com/uber-go/dig/blob/3a148a28d5453a4527f72b23dc44366ac39ca3db/scope_test.go#L351

Upon inspection, looks like the failure is correct? The `child.Provide`s are not exported so the parent should not see them. If I add `dig.Export(true)` to them, the test passes again.

```diff
diff --git a/scope_test.go b/scope_test.go
index cd7ec10..3710f75 100644
--- a/scope_test.go
+++ b/scope_test.go
@@ -371,9 +371,9 @@ func TestScopeValueGroups(t *testing.T) {
 		})
 
 		child := root.Scope("child")
-		child.RequireProvide(func() string { return "a" }, dig.Group("foo"))
-		child.RequireProvide(func() string { return "b" }, dig.Group("foo"))
-		child.RequireProvide(func() string { return "c" }, dig.Group("foo"))
+		child.RequireProvide(func() string { return "a" }, dig.Group("foo"), dig.Export(true))
+		child.RequireProvide(func() string { return "b" }, dig.Group("foo"), dig.Export(true))
+		child.RequireProvide(func() string { return "c" }, dig.Group("foo"), dig.Export(true))
 
 		// Invocation in child should see values provided to the child,
 		// even though the constructor we're invoking is provided in
```

abhinav (approved):

PR #332, opened 2022-03-15
Bubble location from LocationForPC's ProvideOption up to errors

josephinedotlee:
Previously, if a location was passed to Provide via the
LocationForPC ProvideOption, the location would not be taken
into account in the case of an error where a constructor could
not be Provided.

This change updates the error returned, making it more readable by
basing the error message on alternate function provided by
LocationForPC rather than the constructor.

For example, the following error:
```
Failed: cannot provide function "reflect".makeFuncStub (/usr/local/Cellar/go/1.17.6/libexec/src/reflect/asm_amd64.s:30):
cannot provide *fx_test.a[name="a"] from [0].Field0: already provided by
"go.uber.org/fx_test".TestAnnotate.func1 (/gocode/src/github.com/uber-go/fx/annotated_test.go:515)
```

Becomes:
```
Failed: cannot provide function "go.uber.org/fx_test".TestAnnotate.func1 (/gocode/src/github.com/uber-go/fx/annotated_test.go:515):
cannot provide *fx_test.a[name="a"] from [0].Field0: already provided by
"go.uber.org/fx_test".TestAnnotate.func1 (/gocode/src/github.com/uber-go/fx/annotated_test.go:515)
```

### DIFF
--- a/provide.go
+++ b/provide.go
@@ -417,8 +417,15 @@ func (s *Scope) Provide(constructor interface{}, opts ...ProvideOption) error {
 	}
 
 	if err := s.provide(constructor, options); err != nil {
+		var errFunc *digreflect.Func
+		if options.Location == nil {
+			errFunc = digreflect.InspectFunc(constructor)
+		} else {
+			errFunc = options.Location
+		}
+
 		return errProvide{
-			Func:   digreflect.InspectFunc(constructor),
+			Func:   errFunc,
 			Reason: err,
 		}
 	}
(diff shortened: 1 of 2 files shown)


provide.go
```diff
@@ -417,8 +417,15 @@ func (s *Scope) Provide(constructor interface{}, opts ...ProvideOption) error {
 	}
 
 	if err := s.provide(constructor, options); err != nil {
+		var errFunc *digreflect.Func
+		if options.Location == nil {
+			errFunc = digreflect.InspectFunc(constructor)
+		} else {
+			errFunc = options.Location
```
sywhang, line 424: @josephinedotlee Please add tests here :)

sywhang (approved):

abhinav (approved):

PR #333, opened 2022-03-15
Fix issue with multiple decorations of value group in same Scope

sywhang:
This fixes a similar issue to #329 in value groups. When a value group is decorated, and its decorator depends on another decorated type within the same scope, it was skipping over due to incorrect logic in attempt to break the infinite recursion loop. 

This fix uses the concept introduced in #329 - we check if the decorator's "state" to make sure we don't run into infinite recursion, so that we can safely search the current scope for additional decorators instead of dumbly skipping over it all the time when there is a decoration happening.

This also enables us to remove the no-longer used parameter `decorating` from the parameter interfaces.


abhinav (approved):
yay. that "decorating" parameter always did feel weird.

Issue #343, opened 2022-07-07
Invoke order affects private provided types


sywhang:
@hbdf reported this issue to me.

The ordering of Invoke seems to affect the value for private-provided types. 

For example:
```go
parent := dig.New()
child := parent.Scope("child")

child.Provide(func() string {
  return "child"
}, dig.Export(false))

parent.Provide(func() string {
  return "parent"
})

parent.Invoke(func(s string) {
  fmt.Println(s)
})
child.Invoke(func(s string) {
  fmt.Println(s)
})
```

prints:
```
parent
parent
```

But the following code:
```go
parent := dig.New()
child := parent.Scope("child")

child.Provide(func() string {
  return "child"
}, dig.Export(false))

parent.Provide(func() string {
  return "parent"
})

child.Invoke(func(s string) {
  fmt.Println(s)
})
parent.Invoke(func(s string) {
  fmt.Println(s)
})
```

prints:
```
child
parent
```

The first code snippet should've really printed:
```
parent
child
```

sywhang:
Fixed by #344.

PR #344, opened 2022-07-14
Fix issue with wrong invoke for multiple private providers

hbdf:
This fixes issue #343 - we check if the value can be provided in each scope from current to root, instead of first searching for pre built ones in all scopes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,3 @@
 *.coverprofile
 coverage.txt
 *.pprof
-*.swp
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 *.coverprofile
 coverage.txt
 *.pprof
+*.swp
--- a/scope_test.go
+++ b/scope_test.go
@@ -71,6 +71,25 @@ func TestScopedOperations(t *testing.T) {
 		assert.Error(t, c.Invoke(useB))
 	})
 
+	t.Run("private provides doesn't depend on invoke order", func(t *testing.T) {
+		c := digtest.New(t)
+
+		c.RequireProvide(func() string { return "container" })
+
+		child := c.Scope("child")
+		child.RequireProvide(func() string { return "child" })
+
+		verifyContainerStr := func(s string) {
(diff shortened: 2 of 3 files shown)


scope_test.go
```diff
@@ -71,6 +71,25 @@ func TestScopedOperations(t *testing.T) {
...
 	})
 
+	t.Run("private provides doesn't depend on invoke order", func(t *testing.T) {
+		c := digtest.New(t)
+
+		c.RequireProvide(func() string { return "container" })
+
+		child := c.Scope("child")
+		child.RequireProvide(func() string { return "child" })
+
```
moisesvega, line 81:
```suggestion
		child := c.Scope("child")
		child.RequireProvide(func() string { return "child" })
		c.RequireProvide(func() string { return "container" })
```

scope_test.go
```diff
@@ -71,6 +71,25 @@ func TestScopedOperations(t *testing.T) {
...
+		verifyContainerStr := func(s string) {
+			assert.Equal(t, "container", s)
+		}
+		verifyChildStr := func(s string) {
+			assert.Equal(t, "child", s)
+		}
+
+		c.RequireInvoke(verifyContainerStr)
+		child.RequireInvoke(verifyChildStr)
```
moisesvega, line 90:
```suggestion
		child.RequireInvoke(verifyChildStr)
		c.RequireInvoke(verifyContainerStr)
```

.gitignore
```diff
@@ -10,3 +10,4 @@
 *.coverprofile
 coverage.txt
 *.pprof
+*.swp
```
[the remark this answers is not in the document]
hbdf: done!

scope_test.go
hbdf: done!

scope_test.go
hbdf: I don't think this will test the fix, calling the invoke from the child scope first won't cause problem. Am I missing something?

sywhang (approved):
Thanks for the contribution @hbdf :)

PR #348, opened 2022-07-29
Add support for soft value groups

EstebanOlmedo:
This adds the soft attribute to value parameter value groups. Types in
a soft value group won't call their providers if they are the only value
that it's provided by these constructors. This helps to avoid providing
unnecessary values to the dependency graph and also prevent calling
unnecessary constructors.

Example:
```
type Param struct {
    dig.In

    Values []string `group:"foo,soft"`
}
type Result1 struct {
    dig.Out

    Value1 string `group:"foo"`
}

// This constructor won't be called as it only provides a value to the
// soft group
func Result1Ctr() Result1 {
    return Result1{ Value1: "hello" }
}
type Result2 struct {
    dig.Out

    Value1 string `group:"foo"`
    Value2 int
}

// This constructor will be called as we need the provided integer
func Result2Ctr() Result2 {
    return Result2{ Value1: "bye", Value2: 10 }
}

c := dig.New()
c.Provide(Result1Ctr)
c.Provide(Result2Ctr)
// The string "bye" is provided because its constructor is called anways
c.Invoke(func(num int, p Param) {
    // p.Values = {"bye"}
    // num = 10
})
```

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -1380,7 +1380,7 @@ func TestGroups(t *testing.T) {
 		assert.Contains(t, err.Error(), "flatten can be applied to slices only")
 	})
 
-	t.Run("check providers for value groups aren't called if only they're provided", func(t *testing.T) {
+	t.Run("a soft value group provider is not called when only that value group is consumed", func(t *testing.T) {
 		type Param struct {
 			dig.In
 
@@ -1394,7 +1394,7 @@ func TestGroups(t *testing.T) {
 		c := digtest.New(t)
 
 		c.RequireProvide(func() (Result, int) {
-			require.FailNow(t, "This shouldn't be called")
+			require.FailNow(t, "this function should not be called")
 			return Result{Value: "sad times"}, 20
 		})
 
@@ -1403,7 +1403,7 @@ func TestGroups(t *testing.T) {
 		})
 	})
 
-	t.Run("soft value group", func(t *testing.T) {
+	t.Run("soft value group is provided", func(t *testing.T) {
 		type Param1 struct {
 			dig.In
 
@@ -1460,7 +1460,12 @@ func TestGroups(t *testing.T) {
 			assert.Equal(t, float32(3.1416), p.Value3)
 		})
 	})
-
+	t.Run("soft in a result value group", func(t *testing.T) {
+		c := digtest.New(t)
+		err := c.Provide(func() int { return 10 }, dig.Group("foo,soft"))
+		require.Error(t, err, "failed to privide")
+		assert.Contains(t, err.Error(), "cannot use soft with result value groups")
+	})
 	t.Run("value group provided after a hard dependency is provided", func(t *testing.T) {
 		type Param struct {
 			dig.In
--- a/dig_test.go
+++ b/dig_test.go
@@ -1380,6 +1380,29 @@ func TestGroups(t *testing.T) {
 		assert.Contains(t, err.Error(), "flatten can be applied to slices only")
 	})
 
+	t.Run("check providers for value groups aren't called if only they're provided", func(t *testing.T) {
+		type Param struct {
+			dig.In
+
+			Values []string `group:"foo,soft"`
+		}
+		type Result struct {
+			dig.Out
+
+			Value string `group:"foo"`
+		}
+		c := digtest.New(t)
+
+		c.RequireProvide(func() (Result, int) {
+			require.FailNow(t, "This shouldn't be called")
+			return Result{Value: "sad times"}, 20
+		})
+
+		c.RequireInvoke(func(p Param) {
+			assert.ElementsMatch(t, []string{}, p.Values)
+		})
+	})
+
 	t.Run("soft value group", func(t *testing.T) {
 		type Param1 struct {
 			dig.In
--- a/dig_test.go
+++ b/dig_test.go
@@ -53,7 +53,6 @@ func TestEndToEndSuccess(t *testing.T) {
 			require.NotNil(t, got, "invoke got nil buffer")
 			require.True(t, got == b, "invoke got wrong buffer")
 		})
-
 	})
 
 	t.Run("nil pointer constructor", func(t *testing.T) {
@@ -65,7 +64,6 @@ func TestEndToEndSuccess(t *testing.T) {
 		c.RequireInvoke(func(b *bytes.Buffer) {
 			require.Nil(t, b, "expected to get nil buffer")
 		})
-
 	})
 
 	t.Run("struct constructor", func(t *testing.T) {
@@ -80,7 +78,6 @@ func TestEndToEndSuccess(t *testing.T) {
 			// ensure we're getting back the buffer we put in
 			require.Equal(t, "foo", b.String(), "invoke got new buffer")
 		})
-
 	})
 
 	t.Run("slice constructor", func(t *testing.T) {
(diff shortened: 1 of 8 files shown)


result.go
```diff
@@ -90,6 +90,10 @@ func newResult(t reflect.Type, opts resultOptions) (result, error) {
 				"cannot parse group %q", opts.Group, err)
 		}
 		rg := resultGrouped{Type: t, Group: g.Name, Flatten: g.Flatten}
+		if g.Soft {
+			return nil, errf("cannot use soft with result value groups",
+				"soft was used with group:%q", g.Name)
+		}
```
sywhang, line 96: Let's add a test case for covering this error case

result.go
```diff
@@ -469,6 +473,9 @@ func newResultGrouped(f reflect.StructField) (resultGrouped, error) {
 	case g.Flatten && f.Type.Kind() != reflect.Slice:
 		return rg, errf("flatten can be applied to slices only",
 			"field %q (%v) is not a slice", f.Name, f.Type)
+	case g.Soft:
+		return rg, errf("cannot use soft with result value groups",
+			"soft was used with group:%q", rg.Group)
```
sywhang, line 478:
```suggestion
			"soft was used with group %q", rg.Group)
```

dig_test.go
```diff
@@ -1433,6 +1379,89 @@ func TestGroups(t *testing.T) {
...
+		c.Provide(func() (Result, int) { return Result{Value1: "a"}, 2 })
+
+		c.RequireInvoke(func(param Param) {
+			assert.ElementsMatch(t, []string{}, param.Value)
+		})
+		c.RequireInvoke(func(int) {})
+		c.RequireInvoke(func(param Param) {
+			assert.ElementsMatch(t, []string{"a"}, param.Value)
+		})
+	})
```
sywhang, line 1464: Let's add a test case for the canonical example of a constructor that provides 2 types:
1. a type that is only consumed by a soft value group
2. a type that is not consumed by anything

And add a `assert.Error(t, "this constructor should not be called")` in that constructor to make sure it's not getting called.

dig_test.go
```diff
@@ -1433,6 +1379,89 @@ func TestGroups(t *testing.T) {
...
+
+		// The only value that must be in the group is the one that's provided
+		// because it would be provided anyways by another dependency, in
+		// this case we need an int, so the first constructor is called, and
+		// this provides a string, which is the one in the group
+		c.RequireInvoke(func(p2 int, p1 Param1) {
+			assert.ElementsMatch(t, []string{"a"}, p1.Values)
+		})
+	})
+
```
sywhang, line 1408: Can you also add a test case for using this with Decorate?

sywhang (commented):
Looking good! Left a couple of suggestions on tests to add.

param.go
```diff
@@ -395,7 +395,22 @@ func newParamObject(t reflect.Type, c containerStore) (paramObject, error) {
...
+			if p.Soft {
+				groupQueue = append(groupQueue, f)
+				continue
+			}
+		}
+		v, err := f.Build(c)
+		if err != nil {
+			return dest, err
+		}
+		dest.Field(f.FieldIndex).Set(v)
```
sywhang, line 411: looks like there's some amount of dup code here. i think we can simplify this block to reduce some complexity -- you're essentially re-organizing the ordering of the fields so that you build non-groups first. Maybe just build out two queues first by iterating through the whole fields and then merge the queues to actually call Build on them.

Also you might want to add a comment on why this is necessary as it's not immediately obvious why we need to do this.

dig_test.go
```diff
@@ -1433,6 +1379,112 @@ func TestGroups(t *testing.T) {
 		require.Error(t, err, "failed to provide")
 		assert.Contains(t, err.Error(), "flatten can be applied to slices only")
 	})
+
+	t.Run("check providers for value groups aren't called if only they're provided", func(t *testing.T) {
```
r-hang, line 1383: What are your thoughts about:
"a soft value group provider is not called when only that value group is consumed"

dig_test.go
```diff
@@ -1433,6 +1379,112 @@ func TestGroups(t *testing.T) {
...
+		}
+		type Result struct {
+			dig.Out
+
+			Value string `group:"foo"`
+		}
+		c := digtest.New(t)
+
+		c.RequireProvide(func() (Result, int) {
+			require.FailNow(t, "This shouldn't be called")
```
r-hang, line 1397: nit: The majority of the tests in this file use "should not be called". Uniformity makes it easier to ctrl-f.

dig_test.go
```diff
@@ -1433,6 +1379,112 @@ func TestGroups(t *testing.T) {
...
+			return Result{Value: "sad times"}, 20
+		})
+
+		c.RequireInvoke(func(p Param) {
+			assert.ElementsMatch(t, []string{}, p.Values)
+		})
+	})
+
+	t.Run("soft value group", func(t *testing.T) {
```
r-hang, line 1406: wording: "soft value group is provided"

dig_test.go
```diff
@@ -1433,6 +1379,112 @@ func TestGroups(t *testing.T) {
...
+
+			Value1 []string `group:"foo,soft"`
+			Value2 []int    `group:"bar,soft"`
+			Value3 float32
+		}
+
+		type result struct {
+			dig.Out
+
+			Result1 []string `group:"foo,flatten"`
```
r-hang, line 1444: appreciation: Nice flattened + non-flattened exercise.

param.go
```diff
@@ -395,7 +395,22 @@ func newParamObject(t reflect.Type, c containerStore) (paramObject, error) {
 
 func (po paramObject) Build(c containerStore) (reflect.Value, error) {
 	dest := reflect.New(po.Type).Elem()
+	var groupQueue []paramObjectField
 	for _, f := range po.Fields {
+		p, ok := f.Param.(paramGroupedSlice)
+		if ok {
```
r-hang, line 401: Can this combined with ```if ok && p.Soft {``` to reduce nesting?

r-hang:
Looking good! Appreciate all the great tests, left some minor comments.

param.go
```diff
@@ -485,6 +499,9 @@ type paramGroupedSlice struct {
 	// Type of the slice.
 	Type reflect.Type
 
+	// Soft
```
sywhang, line 502: 😆 
Let's either remove this comment or make it more descriptive.

param.go
```diff
@@ -395,7 +395,21 @@ func newParamObject(t reflect.Type, c containerStore) (paramObject, error) {
...
 func (po paramObject) Build(c containerStore) (reflect.Value, error) {
 	dest := reflect.New(po.Type).Elem()
+	// We have to build soft groups after all other fields, to avoid cases
+	// when a field calls a provider for a soft value group, but the value is
+	// not provided to it because the value group is declared before the field
+	var softGroupsQueue []paramObjectField
+	var fields []paramObjectField
 	for _, f := range po.Fields {
+		p, ok := f.Param.(paramGroupedSlice)
+		if ok && p.Soft {
```
sywhang, line 405: Can inline this:
```suggestion
		if p, ok := f.Param.(paramGroupedSlice); ok && p.Soft {
```

sywhang (approved):
👍 excellent work! Couple of minor comments but LGTM other than those.

PR #352, opened 2022-08-03
Use Go 1.19 / Remove deprecated API usage

sywhang:
Go 1.19 is out. Let's use it in the CI.

Also, this updates staticcheck version and removes deprecated API usage.


abhinav (approved):
probably gotta run gofmt

Issue #356, opened 2022-09-12
Ability to check if the error belongs to dig


AnatolyRugalev:
**Is your feature request related to a problem? Please describe.**
During our application initialization I want to distinguish errors that my constructors throw from DI graph issues. I need to know this to provide different explanations to the developer. Also, I don't want to show the whole DI graph when the error is trivial (e.g. env var is not set).

**Describe the solution you'd like**
I would like to see a function like `dig.DigError(err)` or `dig.CausedByConstructor(err)` (ugly)

**Describe alternatives you've considered**
An API like go errors would also work here: `dig.IsError`, `dig.AsError` etc. It would be tricky to test against a number of dig errors though, so this seems suboptimal to me

**Is this a breaking change?**
No, it's a new public method

abhinav:
@AnatolyRugalev thanks for the request. 
I think some way of distinguishing dig errors makes sense.

Meanwhile, is [dig.RootCause](https://pkg.go.dev/go.uber.org/dig#RootCause) an option?
It won't report whether something is a dig error, but it will unpack to the first non-dig error.

AnatolyRugalev:
Yeah, I am using `RootCause` and some secondary assertions to get the job done:

```go
cause := dig.RootCause(err)
myErr := new(myerror.MyError)
isStreamErr := errors.As(cause, &myError)
_, isFormatterErr := cause.(fmt.Formatter)
if !isStreamErr && isFormatterErr {
	fmt.Printf("fatal DI error: %+v\n", err)
	os.Exit(3)
}
fmt.Printf("error initializing server: %+v\n", cause)
os.Exit(2)
```

As you can see, I can check whether an error is coming from a known type, but not the opposite which makes this workaround very fragile

sywhang:
cc @tchung1118

AnatolyRugalev:
This issue has been resolved by #360

PR #360, opened 2022-10-25
Update Dig error types & reporting, more easily distinguish Dig errors vs. non-Dig errors

JacobOaks:
This update:
- Introduces the `dig.Error` interface, which is implemented by all errors returned by Dig, allowing users to use `errors.Is` and `errors.As` with Dig errors.
- Replaces the concept of errors causing other errors with Go's error wrapping conventions, so errors caused by other errors can now be unwrapped to their causing error with `errors.Unwrap`.
- Updates the behavior of `dig.RootCause` to return the first error non-`dig.Error` in the chain of errors passed to it, if there is one. Otherwise, it will return the first error that does not wrap another error (the bottom of the chain). In response to [this issue](https://github.com/uber-go/dig/issues/356), users can now do something like this:
```go
rootCause := dig.RootCause(err)
var de dig.Error
if errors.As(rootCause, &de) {
  // is a Dig error
} else {
  // is an error thrown by one of my provided or invoked functions
}
```
- Adds tests testing the new behavior of `dig.RootCause` in `error_test.go`

### DIFF
--- a/error.go
+++ b/error.go
@@ -32,55 +32,19 @@ import (
 )
 
 // Error is an interface implemented by all Dig errors.
+// dig.Errors require additional functionality for recursively
+// formatting string representations of wrapped errors.
 type Error interface {
 	error
-	dummy()
-}
-
-// errSpecification is returned whenever the user provides bad input when
-// interacting with the container. May optionally have a more detailed
-// error wrapped underneath.
-type errSpecification struct {
-	Message string
-	Cause   error
-}
-
-// newErrSpecification creates a new errSpecification. If there is no other
-// error to wrap, it is safe to pass nil.
-func newErrSpecification(msg string, cause error) errSpecification {
-	return errSpecification{msg, cause}
-}
-
-var _ Error = errSpecification{}
-
-func (e errSpecification) dummy() {}
-
-func (e errSpecification) Unwrap() error {
-	return e.Cause
-}
-
-func (e errSpecification) writeMessage(w io.Writer, _ string) {
-	fmt.Fprintf(w, e.Message)
-}
-
-func (e errSpecification) Error() string { return fmt.Sprint(e) }
-func (e errSpecification) Format(w fmt.State, c rune) {
-	formatError(e, w, c)
-}
-
-// wrappedError is a dig.Error with additional functionality
-// that allows them to be formatted via formatError()
-type wrappedError interface {
-	Error
 	fmt.Formatter
 	Unwrap() error
 	writeMessage(w io.Writer, v string)
 }
 
-// formatError will call a wrappedError's writeMessage() method to print the error message
+// formatError will call a dig.Error's writeMessage() method to print the error message
 // and then will automatically attempt to print errors wrapped underneath (which can create
-// a recursive effect if the wrapped error's Format() method points back to this function).
-func formatError(e wrappedError, w fmt.State, v rune) {
+// a recursive effect if the wrapped error's Format() method then points back to this function).
+func formatError(e Error, w fmt.State, v rune) {
 	multiline := w.Flag('+') && v == 'v'
 	verb := "%v"
 	if multiline {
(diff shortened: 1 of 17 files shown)


result.go
```diff
@@ -70,35 +69,33 @@ type resultOptions struct {
...
+			return nil, errValueGroup{
+				fmt.Sprintf("attempted to use soft with result value group: %q", g.Name)}
 		}
 		if g.Flatten {
 			if t.Kind() != reflect.Slice {
-				return nil, errf(
-					"flatten can be applied to slices only",
-					"%v is not a slice", t)
+				return nil, newErrSpecification(
+					fmt.Sprintf("attempted to use flatten on non-slice: %v", t), nil)
```
sywhang, line 98: let's try to minimize updating error strings unless there's a good reason to.

result.go
```diff
@@ -340,13 +339,13 @@ func (ro resultObject) DotResult() []*dot.Result {
...
-			"cannot specify a name for result objects", "%v embeds dig.Out", t)
+		return ro, newErrSpecification(
+			fmt.Sprintf("attempted to specify a name for result object which embeds dig.Out: %v", t), nil)
 	}
 
 	if len(opts.Group) > 0 {
-		return ro, errf(
-			"cannot specify a group for result objects", "%v embeds dig.Out", t)
+		return ro, newErrSpecification(
+			fmt.Sprintf("attempted to specify a group for result object which embeds dig.Out: %v", t), nil)
```
sywhang, line 348: same thing here.

visualize.go
```diff
@@ -62,27 +63,28 @@ func (o visualizeErrorOption) applyVisualizeOption(opt *visualizeOptions) {
...
-			errors = append(errors, ev)
+			errs = append(errs, ev)
 		}
-		e, ok := err.(causer)
-		if !ok {
+		e := errors.Unwrap(err)
+		if e == nil {
 			break
+		} else {
+			err = e
```
sywhang, line 76: with `break`, you don't need this else block.

https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

visualize.go
```diff
@@ -156,11 +158,12 @@ func CanVisualizeError(err error) bool {
 		if _, ok := err.(errVisualizer); ok {
 			return true
 		}
-		e, ok := err.(causer)
-		if !ok {
+		e := errors.Unwrap(err)
+		if e == nil {
 			break
+		} else {
+			err = e
```
sywhang, line 165: same issue here (nesting).

error.go
```diff
@@ -27,187 +27,99 @@ import (
...
-					len(args))
-			}
+// wrappedError is a dig.Error with additional functionality
+// that allows them to be formatted via formatError()
+type wrappedError interface {
+	Error
+	fmt.Formatter
+	Unwrap() error
+	writeMessage(w io.Writer, v string)
+}
```
sywhang, line 78: I haven't gone through and verified every single dig.Error implementations to see if they implement wrappedError, but it seems like they do. In that case, can you just merge the definitions into one, and remove the dummy() method on dig.Error interface?

sywhang (commented):
took a quick pass, will come back to it later tonight.

JacobOaks:
I have pushed a new commit:
- combine `dig.Error` and `wrappedError` interfaces, remove `dummy()` methods
- rename `errSpecification` to `errInvalidInput`. I wanted to be careful to disambiguate the idea of user input error from errors specifically related to building arguments for functions in the Dig container (`errArgumentsFailed`), hence my avoiding using the term "argument".
- clean up code, bring up test coverage, address @sywhang's other points

error.go
```diff
@@ -27,187 +27,99 @@ import (
...
+	Message string
+	Cause   error
+}
 
-	// Writes the message or context for this error in the chain.
-	//
-	// verb is either %v or %+v.
-	writeMessage(w io.Writer, verb string)
+// newErrSpecification creates a new errSpecification. If there is no other
+// error to wrap, it is safe to pass nil.
```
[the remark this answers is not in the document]
JacobOaks: I updated the comment which should explain what I meant by this

josephinedotlee (commented):
oh behalf of @abhinav for his last internal code review 😭

JacobOaks:
Update:
* RootCause() now more closely matches previous behavior: it will return first non-`dig.Error` or first error that does not wrap another error.
* There are now two interfaces for Dig errors. dig.Errors represents the public contract, and digError is an extension for private functionality (primarily error formatting). dig.Errors remains sealed.
* Add fields to structs in test tables

sywhang (approved):

PR #364, opened 2022-12-08
Opt-In Panic Recovery

JacobOaks:
Currently, when a panic occurs in provided/invoked/decorated functions, the message users see can be very unwieldy, especially with complicated dependency graphs.

This change allows us to construct our container with a new `Option` called `RecoverFromPanics()`:

```go
c := dig.New(dig.RecoverFromPanics())
```

When this is done, any panics that occur in provided/invoked/decorated functions will be recovered from, and wrapped in a new error Type, `PanicError`, containing the relevant function and the panic value. This `PanicError` will then be dealt with as a normal error when propagated back us, meaning panics would be just as readable as normal errors when dealing with long complicated call chains.

When opted in, we can distinguish panics from dig errors and non-dig errors when calling `c.Invoke` like so:

```go
if err := c.Invoke(myFunc); err != nil {
    rootCause := dig.RootCause(err)
    var pe dig.PanicError
    var de dig.Error
    if errors.As(rootCause, &pe) {
        // This was originally a panic, look at pe.Panic
    } else if errors.As(rootCause, &de) {
        // This is a Dig error
    } else {
        // This is just an error from my functions
    }
}
```

Or, if we just want to distinguish panics from errors we can simply check `errors.As(err, &pe)`.

### DIFF
--- a/constructor.go
+++ b/constructor.go
@@ -130,7 +130,7 @@ func (n *constructorNode) String() string {
 
 // Call calls this constructor if it hasn't already been called and
 // injects any values produced by it into the provided container.
-func (n *constructorNode) Call(c containerStore) error {
+func (n *constructorNode) Call(c containerStore) (err error) {
 	if n.called {
 		return nil
 	}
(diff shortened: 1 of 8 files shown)


sywhang (commented):
posting on behalf of @r-hang: 

Perhaps add some more details on how to use this in the PR description.

r-hang (approved):

PR #371, opened 2023-01-09
Stop Decorator Running When Dry Run Set

JacobOaks:
Currently, decorators still run even when dry run is set in the container (see Fx issue: https://github.com/uber-go/fx/issues/1018)

This changes decorators to use the scope's invoker when decorating, causing it to fill-in zero-valued results instead of actually running the function when the dry run option is set.


sywhang (approved):
👍

PR #376, opened 2023-02-17
CI: test with Go 1.19 and Go 1.20

sywhang:
Only Go 1.19 and Go 1.20 are supported after Go 1.20 has been released.

This makes CI only run against Go 1.19 and Go 1.20.


abhinav:
Need to update staticcheck

sywhang:
huh, weird. Looks like staticcheck has a bug - it reports `scope` as not being used when it clearly is being used. 
```
internal/digtest/container.go:49:6: type scope is unused (U1000)
```

abhinav:
Possibly "embed an alias" is enough of a corner case. 
Could probably narrow it down.

abhinav:
This can be closed. It was superseded by #394

PR #377, opened 2023-02-17
Introduce Callback Functions for Constructor & Decorator Calls

JacobOaks:
Recently folks have been asking for more visiblity into what constructors and decorators actually get called, especially Fx users (https://github.com/uber-go/fx/issues/1039).

This PR allows users to provide callback functions for Dig to call when it invokes a constructor or decorator. Dig will pass along some basic information to the callback function: the name of the function, and the error it returned if any.

Users can provide a callback by using the two new APIs, `dig.WithProviderCallback` and `dig.WithDecoratorCallback`, which return a type implementing both `dig.ProvideOption` and `dig.DecorateOption`. For example, this code prints a simple message when `myFunc` and `myDecorator` are called (presumably defined elsewhere):

```go
c := dig.New()
myCallback := func(ci CallbackInfo) {
	var errorAdd string
	if ci.Error != nil {
		errorAdd = fmt.Sprintf("with error: %v", ci.Error)
	}
	fmt.Printf("%q finished%v", ci.Name, errorAdd)
}
c.Provide(myFunc, WithProviderCallback(myCallback))
c.Decorate(myDecorator, WithDecoratorCallback(myCallback))
```

We can discuss Fx-side consumption of this functionality to aid observability in this way and help identify dead code, but one idea is to register callback functions that will generate new Fx events when constructors and decorators get called. I wrote a [prototype implementation](https://github.com/JacobOaks/fx/commit/1532d1bd3c7d3926ee3e98d8d15e47206e7192ef#diff-a04d0b8fa4df9f46eb6a4ee7c75554d41ab5c2343bea18ba09fc62538953d9a7) of this idea for Fx, and then wrote a [demo](https://github.com/JacobOaks/fx_dead_code_demo) showing how this can help identify dead code.

Internal Ref: GO-1873

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -20,8 +20,8 @@
 
 package dig
 
-// CallbackInfo contains information about a function called by Dig
-// and is passed to a Callback function registered with [WithCallback].
+// CallbackInfo contains information about a provided function or decorator
+// called by Dig, and is passed to a Callback registered with [WithCallback].
 type CallbackInfo struct {
 
 	// Name is the name of the function in the format:
@@ -33,12 +33,30 @@ type CallbackInfo struct {
 	Error error
 }
 
+// Callback is a function that can be registered with a provided function
+// or decorator with [WithCallback] to cause it to be called after the 
+// provided function or decorator is run.
 type Callback func(CallbackInfo)
 
-// WithCallback allows registering a callback function with Dig
-// to be called whenever a provided function or decorator of a container
-// is called successfully. [Callback] is a simple interface containing
-// the function to be called as a callback.
+// WithCallback returns an option that can be used with [(*Container).Provide]
+// or [(*Container).Decorate] to have Dig call the passed in [Callback]
+// after the corresponding constructor or decorator finishes running.
+//
+// For example, the following prints a simple message after "myFunc" and
+// "myDecorator" finish running:
+// 
+//	c := dig.New()
+//	myCallback := func(ci CallbackInfo) {
+//		var errorAdd string
+//		if ci.Error != nil {
+//			errorAdd = fmt.Sprintf("with error: %v", ci.Error)
+//		}
+//		fmt.Printf("%q finished%v", ci.Name, errorAdd)
+//	}
+//	c.Provide(myFunc, WithCallback(myCallback)),
+//	c.Decorate(myDecorator, WithCallback(myCallback)),
+//
+// See [CallbackInfo] for more info on the information passed to the [Callback].
 func WithCallback(callback Callback) withCallbackOption {
 	return withCallbackOption{
 		callback: callback,
--- a/callback.go
+++ b/callback.go
@@ -1,0 +1,86 @@
+// Copyright (c) 2023 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package dig
+
+// CallbackFuncKind represents the kind of function called in a CallbackInfo.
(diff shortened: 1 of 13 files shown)


callback.go
```diff
@@ -0,0 +1,80 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package dig
+
+// CallbackInfo contains information about a provided function or decorator
+// called by Dig, and is passed to a [Callback] registered with [WithCallback].
+type CallbackInfo struct {
```
manjari25, line 25: I am not sure if this can be done in this but, can this info also encapsulate the type of the associated function name (Decorate vs Provide).

dig_test.go
```diff
@@ -1638,6 +1638,121 @@ func TestRecoverFromPanic(t *testing.T) {
...
+
+		c := digtest.New(t)
+		c.RequireProvide(
+			func() (int, error) {
+				return 0, errors.New("terrible callback sadness")
+			},
+			dig.WithCallback(func(ci dig.CallbackInfo) {
+				assert.Equal(t, "go.uber.org/dig_test.TestCallback.func2.1", ci.Name)
+				require.Error(t, ci.Error)
+				assert.ErrorContains(t, ci.Error, "terrible callback sadness")
```
manjari25, line 1686: Consider using [assert.ErrorContains(...)](https://pkg.go.dev/github.com/stretchr/testify/assert#ErrorContains)

dig_test.go
```diff
@@ -1638,6 +1638,121 @@ func TestRecoverFromPanic(t *testing.T) {
...
+		c := digtest.New(t)
+		c.RequireProvide(giveInt)
+		c.RequireDecorate(
+			func(a int) (int, error) {
+				return 0, errors.New("terrible callback sadness")
+			},
+			dig.WithCallback(func(ci dig.CallbackInfo) {
+				assert.Equal(t, "go.uber.org/dig_test.TestCallback.func3.1", ci.Name)
+				require.Error(t, ci.Error)
+				assert.ErrorContains(t, ci.Error, "terrible callback sadness")
```
manjari25, line 1707: Consider using [assert.ErrorContains(...)](https://pkg.go.dev/github.com/stretchr/testify/assert#ErrorContains)

callback.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Name is the name of the function in the format:
+	// <package_name>.<function_name>
+	Name string
+
+	// Error contains the error returned by the [Callback]'s associated
+	// function, if there was one.
+	Error error
```
manjari25, line 33: It would be nice if we can allow differentiation between regular errors vs panic errors.

dig_test.go
```diff
@@ -1638,6 +1638,121 @@ func TestRecoverFromPanic(t *testing.T) {
...
+		var called bool
+
+		c := digtest.New(t, dig.RecoverFromPanics())
+		c.RequireProvide(giveInt)
+		c.RequireDecorate(
+			func(int) int { panic("unreal misfortune") },
+			dig.WithCallback(func(ci dig.CallbackInfo) {
+				assert.Equal(t, "go.uber.org/dig_test.TestCallback.func5.1", ci.Name)
+				require.Error(t, ci.Error)
+				assert.ErrorContains(t, ci.Error, "panic: \"unreal misfortune\"")
```
manjari25, line 1745: Consider using assert.ErrorContains(...)

callback.go
JacobOaks: Users should be able to do this using `errors.As`
https://pkg.go.dev/go.uber.org/dig#PanicError

callback.go
JacobOaks: We spoke offline and agreed we can err on the side of omitting this info for now, though it is very easy to add later if a need arises.

callback.go
```diff
@@ -0,0 +1,80 @@
...
+//		if ci.Error != nil {
+//			errorAdd = fmt.Sprintf("with error: %v", ci.Error)
+//		}
+//		fmt.Printf("%q finished%v", ci.Name, errorAdd)
+//	}
+//	c.Provide(myFunc, WithCallback(myCallback)),
+//	c.Decorate(myDecorator, WithCallback(myCallback)),
+//
+// See [CallbackInfo] for more info on the information passed to the [Callback].
+func WithCallback(callback Callback) withCallbackOption {
```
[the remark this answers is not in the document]
JacobOaks: I went ahead and made this change to resolve the lint error, but I'm also not sure if it's an anti-pattern. The one thing I can think of is that it's still not satisfying from an API documentation point of view. For example, I won't be able to go here: https://pkg.go.dev/go.uber.org/dig#ProvideOption and immediately see every possible `ProvideOption`, because this one won't be listed. But returning an unexported type doesn't solve that either.

I was trying to be conservative with API exposures, but perhaps two separate functions `func WithConstructorCallback(callback Callback) ProvideOption` and `func WithDecorateCallback(callback Callback) DecorateOption` really is the better option?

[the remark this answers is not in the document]
tchung1118: I see. Yeah, having separate functions might be a better option. I'm interested in what @sywhang thinks about this.

manjari25 (approved):
LGTM

callback.go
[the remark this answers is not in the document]
sywhang: Let's not do this and prefer to declare separate Options. 
```
WithProviderCallback(CallbackInfo) ProvideOption
WithDecoratorCallback(CallbackInfo) DecorateOption
```

It's possible we may be able to clean this up using generics after we clean up the definition of `ProvideOption` and `DecorateOption` by making some changes to its unexported method, but as of now I don't think that's within the scope of this PR. 

In the future we could consider doing this using generics, and by then we still have the name `WithCallback` reserved to do so.

tchung1118 (approved):

sywhang (approved):

PR #383, opened 2023-04-14
Add support for getting info from Invoke statements

manjari25:
This commit exposes `InvokeInfo` and `FillInvokeInfo` to help users extract the types requested by an Invoke statement.

Example usage -

```
var info InvokeInfo
c.Invoke(func(typ1 type1, typ2 type2){}, FillInvokeInfo(&info))

// Will print the types requested by the Invoke statement
for _, typ := range info.ReqTypes {
    fmt.Println(typ) 
}

``` 

Refs: GO-669

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -3757,7 +3757,7 @@ func TestInvokeInfoOption(t *testing.T) {
 			require.NotNil(t, m, "invoke got zero value map")
 		}, dig.FillInvokeInfo(&info))
 
-		assert.Equal(t, 1, len(info.ReqTypes))
+		assert.Len(t, info.ReqTypes, 1)
 		assert.Equal(t, "map[string]string", info.ReqTypes[0].String())
 	})
 
@@ -3776,7 +3776,7 @@ func TestInvokeInfoOption(t *testing.T) {
 			require.NotNil(t, typ1, "invoke got nil struct")
 		}, dig.FillInvokeInfo(&info))
 
-		assert.Equal(t, 2, len(info.ReqTypes))
+		assert.Len(t, info.ReqTypes, 2)
 		assert.Equal(t, "map[string]string", info.ReqTypes[0].String())
 		assert.Equal(t, "*dig_test.type1", info.ReqTypes[1].String())
 	})
@@ -3795,14 +3795,14 @@ func TestInvokeInfoOption(t *testing.T) {
 		c.RequireInvoke(func(typ1 *type1) {
 			require.NotNil(t, typ1, "invoke got nil struct")
 		}, dig.FillInvokeInfo(&info1))
-		assert.Equal(t, 1, len(info1.ReqTypes))
+		assert.Len(t, info1.ReqTypes, 1)
 		assert.Equal(t, "*dig_test.type1", info1.ReqTypes[0].String())
 
 		var info2 dig.InvokeInfo
 		c.RequireInvoke(func(typ2 *type2) {
 			require.NotNil(t, typ2, "invoke got nil struct")
 		}, dig.FillInvokeInfo(&info2))
-		assert.Equal(t, 1, len(info2.ReqTypes))
+		assert.Len(t, info2.ReqTypes, 1)
 		assert.Equal(t, "*dig_test.type2", info2.ReqTypes[0].String())
 	})
 
@@ -3825,7 +3825,7 @@ func TestInvokeInfoOption(t *testing.T) {
 		c.RequireInvoke(func(p params) {
 			require.Equal(t, params{Field1: "hello", Field2: 2023}, p)
 		}, dig.FillInvokeInfo(&info))
-		assert.Equal(t, 2, len(info.ReqTypes))
+		assert.Len(t, info.ReqTypes, 2)
 		assert.Equal(t, "string", info.ReqTypes[0].String())
 		assert.Equal(t, "int", info.ReqTypes[1].String())
 	})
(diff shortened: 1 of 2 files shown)


JacobOaks (approved):
LGTM

invoke.go
```diff
@@ -22,16 +22,41 @@ package dig
...
+	applyInvokeOption(*invokeOptions)
+}
+
+type invokeOptions struct {
+	Info *InvokeInfo
+}
+
+// InvokeInfo provides information about an Invoke.
+type InvokeInfo struct {
+	ReqTypes []*Output
```
sywhang, line 41: Should we also provide `Input`? (see ProvideInfo)

abhinav: Two things:

- Instead of ReqTypes, is there a fuller name? Inputs or Outputs?
- Are the items inputs or outputs? IIRC, ProvideInfo is Inputs for arguments.

So maybe this should be `Inputs []*Input`?

invoke.go
```diff
@@ -22,16 +22,41 @@ package dig
...
+}
+
+// FillInvokeInfo is an InvokeOption that writes info on the types requested
+// by the Invoke function into the specified InvokeInfo.
+func FillInvokeInfo(info *InvokeInfo) InvokeOption {
```
abhinav, line 46: It would help to include a sample usage here or an example test elsewhere:

```go
// For example:
//
//    var info dig.InvokeInfo
//    err := c.Invoke(..., dig.FillInvokeInfo(&info))
```

invoke.go
sywhang: yup, +1 on the naming of the field: let's change this to `Inputs` and `Outputs`.

invoke.go
manjari25: Done

dig_test.go
```diff
@@ -3741,6 +3741,119 @@ func TestProvideInfoOption(t *testing.T) {
 	})
 }
 
+func TestInvokeInfoOption(t *testing.T) {
```
sywhang, line 3744: These tests can easily be, and should be, consolidated to a table test. 

Also, some of these tests don't seem like they're necessary. i.e. why do you need the "two invokes" test? that's not testing a different code path that doesn't get exercised by a single invoke, afaict.

invoke.go
```diff
@@ -22,16 +22,45 @@ package dig
...
+type InvokeInfo struct {
+	Inputs []*Input
+}
+
+// FillInvokeInfo is an InvokeOption that writes info on the types requested
+// by the Invoke function into the specified InvokeInfo.
+// For example:
+//
+//	var info dig.InvokeInfo
+//	err := c.Invoke(..., dig.FillInvokeInfo(&info))
```
sywhang, line 49: Maybe provide a _real_ example of a function, and what the corresponding provided info looks like?

invoke.go
```diff
@@ -22,16 +22,45 @@ package dig
...
+func FillInvokeInfo(info *InvokeInfo) InvokeOption {
+	return fillInvokeInfoOption{info: info}
+}
+
+type fillInvokeInfoOption struct {
+	info *InvokeInfo
+}
+
+func (o fillInvokeInfoOption) String() string {
+	return fmt.Sprintf("FillInvokeInfo(%p)", o.info)
```
sywhang, line 59: you can add a test for this. See "TestFillProvideInfoString" in provide_test.go.

invoke.go
```diff
@@ -22,16 +22,45 @@ package dig
...
+type invokeOptions struct {
+	Info *InvokeInfo
+}
+
+// InvokeInfo provides information about an Invoke.
+type InvokeInfo struct {
+	Inputs []*Input
+}
+
+// FillInvokeInfo is an InvokeOption that writes info on the types requested
```
abhinav, line 44: nit: s/writes info/writes information/

invoke.go
```diff
@@ -22,16 +22,45 @@ package dig
...
+// by the Invoke function into the specified InvokeInfo.
```
abhinav, line 45: nit: "accepted by the Invoke function"

dig_test.go
abhinav: One easy way  to do this is:
declare a container and add a bunch of provides,
then build a table of `struct{ desc string; invoke any; want *InvokeInfo }`
where `invoke any` holds the function to invoke.

To keep things simple and more readable, you don't need to bother with the `require.NotNil` inside the function bodies. We've already exhaustively tested Dig, so this test can be limited to just testing the InvokeInfo functionality.

```go
c := digtest.New(t)
c.Provide(...)
// .. a few more depending on what you need.
// the provides themselves are not super important
// from the point of view of this test.
// as a reader, all i care about is "given this function signature, this is the invoke info".

tests := []struct {
  desc string
  give any
  want *InvokeInfo
}{
  {
    desc: "one argument",
    give: func(map[string]string) {},
    want: *InvokeInfo{ ... }
  },
  // ...
}

for _, tt := range tests {
  // ...
}
```

abhinav (commented):
New API looks and the implementation looks fine.

sywhang (approved):
👍

dig_test.go
```diff
@@ -3741,6 +3741,91 @@ func TestProvideInfoOption(t *testing.T) {
...
+		var info dig.InvokeInfo
+		c.RequireInvoke(tt.invokeFn, dig.FillInvokeInfo(&info))
+		require.NotNil(t, info)
+		require.Len(t, info.Inputs, tt.numTyps)
+		for i := 0; i < tt.numTyps; i++ {
+			assert.Equal(t, tt.wantStrs[i], info.Inputs[i].String())
+		}
+	}
+}
+
```
tchung1118, line 3810: could we add a test case where `info` that's passed to dig.FillInvokeInfo is nil?

tchung1118 (commented):
From group code review

alexisvisco:
@JacobOaks just to say that I have added outputs in the invoke information #395

Issue #389, opened 2023-06-08
Not providing any value group does not return an error during invocation


paullen:
**Describe the bug**
While requesting a value group in the input, in the `Invoke` method, dig does not check for the existence of any providers of the value group.

**To Reproduce**

Run the following code. The function is invoked without letting me know that there are no value group providers.
```
package main

import (
	"fmt"
	"log"

	"go.uber.org/dig"
)

type Logger struct{}
type LogType string

const (
	LogType_ERROR   LogType = "ERROR"
	LogType_INFO    LogType = "INFO"
	LogType_WARNING LogType = "WARNING"
)

func (l Logger) Log(logType LogType, message string) {
	log.Println(logType, message)
}

type HttpClient struct {
	logger *Logger
}

func (client *HttpClient) Get(url string) (string, error) {
	client.logger.Log(LogType_INFO, "getting "+url)

	return "You got " + url, nil
}

type HttpClientInput struct {
	dig.In
	Loggers []*Logger `group:"loggers"`
}

func NewHttpClient(input HttpClientInput) *HttpClient {
	if len(input.Loggers) > 0 {
		return &HttpClient{logger: input.Loggers[0]}
	} else {
		panic("should not execute")
	}
}

func BuildContainer() *dig.Container {
	container := dig.New()

	if err := container.Provide(NewHttpClient); err != nil {
		log.Println(err)
	}

	return container
}

func ExecuteFunction(client *HttpClient) {
	res, _ := client.Get("foo.bar")

	fmt.Println(res)
}

func main() {
	container := BuildContainer()

	if err := container.Invoke(ExecuteFunction); err != nil {
		log.Println(err)
	}

}

```

**Expected behavior**
The expected behaviour would be to return an error in the `Invoke` function after checking for providers.

**Additional context**
If this is by design, please help me understand the significance of this.

JacobOaks:
Hey @paullen - as I mentioned in #184, this shouldn't be changed blindly as it would break a lot of users, but I think it's a good idea to add an option that can be passed to Dig to check for this.

Issue #184, opened 2018-03-08
Feature request: hints for pointer vs struct in value groups


kriskowal:
We do a great job hinting when you provide a struct instead of a pointer to a struct, or the other way around. However, this magic does not appear to extend all the way to value groups.

I was stumped by a test:

```
	var p struct {
		fx.In

		ProcedureLists [][]*transport.Procedure `group:"yarpcfx"`
	}
```

This did not work because the value group contains slices of transport.Procedure structs, not pointers thereto.

Would be great if the Fx output hinted that it would populate the ProcedureLists were it without the pointer indirection.

breerly:
Ah, good catch.

abhinav:
If anyone wants to get their hands dirty with dig, this should be fun to implement. It's all contained in [newErrMissingType](https://github.com/uber-go/dig/blob/15c67344e48b1b3317399a647b0ec0069c34321d/error.go#L173).

paullen:
Hi. Been trying to implement this is in [newErrMissingType](https://github.com/uber-go/dig/blob/15c67344e48b1b3317399a647b0ec0069c34321d/error.go#L173) as suggested by @abhinav but the problem, as I've described in #389, is that this method is never called when you don't provide any constructors for a value group. Should we add a check for no constructors for value groups? Otherwise we'll need to go further up the call stack to check for this.

JacobOaks:
I do think that this would be a useful check to add, but my concerns are:

* I don't think it is necessarily invalid to have an empty value group, though you could make the argument that an empty value group is _very likely_ to be an issue when the opposite kind (pointer vs. value) was detected as being provided.
* It would break current users who DO have empty value groups

I think if we do add this, we should probably gate it behind an option, `dig.FailEmptyValueGroups`, or something like that, but curious to hear others' thoughts.

PR #390, opened 2023-06-08
Add suggestions for slices and array element missing types

paullen:
Changes:

Added suggestions for requesting slice/array of pointers instead of the elements and vice versa

### DIFF
--- a/dig_test.go
+++ b/dig_test.go
@@ -3255,6 +3255,11 @@ func testInvokeFailures(t *testing.T, dryRun bool) {
 
 			A `name:"hello"`
 		}
+		type outPointerA struct {
+			dig.Out
+
+			A *A `name:"hello"`
+		}
 
 		cases := []struct {
 			name        string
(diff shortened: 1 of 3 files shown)


paullen:
Hey. Just wanted to follow up on this PR. This seems like a good suggestion to have, it is a part of the solution for Issue #184 . Then there is the issue #389 that addresses the other half of the #184 issue.

JacobOaks:
Hey @paullen, thanks for this PR!

I think this is a good addition, can we just add a test case similar to [this one](https://github.com/uber-go/dig/blob/a30081d72915995cb8f934ba9393256c74df9143/dig_test.go#L3251) that tests that the suggestions get properly populated from a call to `fx.Invoke`?

paullen:
@JacobOaks Added test cases, the test coverage is now 99%, up from 98.3%. Please review. Thanks!

JacobOaks (approved):
Hey, sorry for the delay. LGTM!

sywhang (approved):
LGTM.

PR #395, opened 2023-09-01
Feature: add outputs infos into InvokeInfo struct

alexisvisco:
As we are building a home made framework we need to get informations of the output of a invoke function. 

This can be used for instance with http handler, where we use invoke to provide dependency injection and need information for output, for instance getting Raw[T] or Json[T] or Render[T] and do some stuff based on that. 

Hope it will be useful for others.

### DIFF
--- a/invoke.go
+++ b/invoke.go
@@ -38,7 +38,8 @@ type invokeOptions struct {
 
 // InvokeInfo provides information about an Invoke.
 type InvokeInfo struct {
-	Inputs []*Input
+	Inputs  []*Input
+	Outputs []reflect.Value
 }
 
 // FillInvokeInfo is an InvokeOption that writes information on the types
@@ -161,6 +162,12 @@ func (s *Scope) Invoke(function interface{}, opts ...InvokeOption) (err error) {
 	if len(returned) == 0 {
 		return nil
 	}
+
+	if options.Info != nil {
+		options.Info.Outputs = make([]reflect.Value, len(returned))
+		copy(options.Info.Outputs, returned)
+	}
+
 	if last := returned[len(returned)-1]; isError(last.Type()) {
 		if err, _ := last.Interface().(error); err != nil {
 			return err


alexisvisco:
HI, thanks for the reply. 

The reason why I am needing to add this option to the invoke and not to a provide is that provide does not invoke the function instantaneously. In my use case imagine you have this handler : 

```go
func CustomHandler(ctx context.Contet, u UserReq) (UserRes, error) 
```

If I pass this function to the `Provide` function of dig I need to then construct an invoker based on the value returned by the handler (?? seems hard) and then retrieve them with some custom logic. 

Providing a mechanism to retrieve value directly from an `Invoke` seems incomparably simpler and cheaper (at implementing in dig vs doing some custom code in my codebase). 

---- 

And without my case in mind I have other Idea why this feature could be cool, imagine a job function returning some result about what it does and delegate that output to another code that print it. 

---- 

If you have any other questions let me know :)

tchung1118:
I agree with @JacobOaks here. For your use case specifically, providing the handler to the container/scope and then invoking a function that depends on `UserRes` shouldn't be hard. I'm assuming that you're building and using request-specific scopes/containers for each request here:
```
// let's say you're creating a new request-specific scope
reqScope := container.Scope()
... // provide everything you need to call `CustomHandler`
reqScope.Provide(CustomHandler) // provide CustomHandler
err := reqScope.Invoke(func(res UserRes) {
    // do stuff with res
}) // this will call CustomHandler
```
Besides, even if you are somehow re-using your container or scope across requests in your setup and absolutely need to invoke your handler, there's nothing stopping you from storing `UserRes` to a variable that is scoped outside of the function being invoked:
```
var res UserRes
err := reqScope.Invoke(func(ctx context.Context, req UserReq) error {
    res, err := CustomHandler(ctx, req)
    return err
})
// do stuff with res
```
And I'm not convinced that this is any worse than having dig populate the outputs as `reflect.Value` and then trying to use that reflection.

Issue #397, opened 2023-10-16
Provide Order Matters for Scopes -- Detects Nonexistent Cyclical Dependencies


bryanroute:
**Describe the bug**
The order in which values are provided to the container appears to matter but only when using scopes, otherwise it does not.

**To Reproduce**
```
import (
	"go.uber.org/dig"
	"testing"
)

type sampleStruct struct{}

type otherStruct struct{}

func Test_Order(t *testing.T) {
	c := dig.New()

	// This depends on the next provide but normally order doesn't matter
	err := c.Provide(func(val string) sampleStruct { return sampleStruct{} })
	if err != nil {
		t.Error(err)
	}

	// If you put this Provide first the test will pass
	err = c.Provide(func() string { return "Sample" })
	if err != nil {
		t.Error(err)
	}

	scope := c.Scope("test")
	
	// This will error with an error about a cyclical dependency but there is no cycle.
	// If you call Provide directly on the container without a scope no issue
	err = scope.Provide(func() otherStruct { return otherStruct{} })
	if err != nil {
		t.Error(err)
	}
}
```

**Expected behavior**
I would expect this test to pass but it does not unless you move the order of the Provide calls on the container around. I hit this really weird scenario in a much more complicated example and it took me hours to track it down and create this simple repro. I don't understand why normally the order of Provide calls doesn't matter but once a Scope is introduced it does? This seems incorrect especially since the error comes when providing something completely unrelated. Perhaps more importantly the error is about a cyclical dependency which isn't actually a cyclical dependency. 

**Additional context**
If the order of the Provide calls is meant to matter that should be clearly documented and I'd expect the same behavior with the container and scopes.

bryanroute:
Awesome thank you! Any idea when you'll tag another release?

JacobOaks:
@bryanroute Now :)

https://github.com/uber-go/dig/releases/tag/v1.17.1

bryanroute:
Perfect, pulled and confirmed in my tests the problem is resolved. Thank you so much for the quick response!

PR #398, opened 2023-10-18
Fix false positives in cycle detection with child scopes.

JacobOaks:
Currently, cycle detection works by assigning each constructor node an "order"/index for the scope it was given in. If a path exists between an index and itself, a cycle is detected.

When we create a subscope, we copy all of the parent's nodes to the child scope, but we do not copy their orders. This causes all root scope constructors to effectively have order 0 in the child scope (zero value for a mapping to int). This causes false positives when doing cycle detection as an edge from order 0 to order 0 gets detected.

This PR fixes this bug by copying the order for constructor nodes from parent to child scope when a child scope is created.

The added test case fails before this PR, passes with it.

Ref: https://github.com/uber-go/dig/issues/397

### DIFF
--- a/constructor.go
+++ b/constructor.go
@@ -129,6 +129,11 @@ func (n *constructorNode) CType() reflect.Type        { return n.ctype }
 func (n *constructorNode) Order(s *Scope) int         { return n.orders[s] }
 func (n *constructorNode) OrigScope() *Scope          { return n.origS }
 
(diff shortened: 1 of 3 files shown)


tchung1118 (approved):
LGTM

sywhang (approved):
nice.

PR #401, opened 2023-11-01
Returns an existing Scope with the same name

Tiscs:
Returns an existing `Scope` with the same name when calling `Scope(name string, opts ...ScopeOption)` rather than creating a new one.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -107,12 +107,25 @@ func newScope() *Scope {
 	return s
 }
 
-// Scope creates a new Scope with the given name and options from current Scope.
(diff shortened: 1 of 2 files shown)


JacobOaks:
Hi @Tiscs, in its current state this is a breaking change because it was cause all current calls to `Scope()` with the same name to be merged into one scope which could break some users.

Instead, we could add a new API to scope that returns a child subscope with a given name, or perhaps an [Option](https://pkg.go.dev/go.uber.org/dig#Option) that you pass when creating the container that enables this behavior. Let me know your thoughts!

Issue #402, opened 2024-01-14
Child scope of container is not visualized on dot graph


greeflas:
**Describe the bug**
Child scope of container is not visualized on dot graph.

**To Reproduce**
1. Create container: 
`c := dig.New()`
2. Provide the container with constructor: 
`c.Provide(NewFirstComponent)`
3. Create child scope: 
`child := c.Scope("test_scope")`
4. Provide the child scope with constructor: 
`child.Provide(NewSecondComponent)`
5. Visualize container: 
`dig.Visualize(c, os.Stdout)`

**Expected behavior**
I expect to see child scope along with `NewSecondComponent` but I only see `NewFirstComponent`.

**Additional context**
Dig version: `v1.17.1`
Go Playground: https://go.dev/play/p/lhYAIUlfsPr
DOT:
```
digraph {
	rankdir=RL;
	graph [compound=true];
	
		subgraph cluster_0 {
			label = "main";
			constructor_0 [shape=plaintext label="NewFirstComponent"];
			
			"*main.FirstComponent" [label=<*main.FirstComponent>];
			
		}
		
		
	
}
```
Graph image:
![graphviz (1)](https://github.com/uber-go/dig/assets/17636915/07f2d20d-4cd8-46de-8b3f-d29852fc4551)

sywhang:
This is a valid feature request. I believe the DOT representation wasn't rewritten when we rewrote Dig's internal graph representation when we worked on the Scope feature.

greeflas:
@sywhang should we close this issue or you want to add here info about future improvement that was discussed in pull request?

sywhang:
I'll leave this open for now since we need some more visual improvements. Thanks

PR #403, opened 2024-01-16
Visualize child scopes

greeflas:
Hello!

This is fix for #402.

I'll appreciate your feedback 🙌

### DIFF
--- a/visualize.go
+++ b/visualize.go
@@ -174,11 +174,19 @@ func (c *Container) createGraph() *dot.Graph {
 func (s *Scope) createGraph() *dot.Graph {
 	dg := dot.NewGraph()
 
+	s.addNodes(dg)
+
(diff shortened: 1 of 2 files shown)


sywhang:
Thanks @greeflas! Ideally we would add some type of visual representation to highlight the different scopes, but i think this is better than what we have for now. I'll take a look at the implementation later tonight

JacobOaks (approved):
I agree with @sywhang that we will probably eventually want more clear demarcation of scopes, but for now this LGTM to at the least show module's constructors at all.

r-hang (approved):

sywhang (approved):
LGTM. Thank you for the contribution!

greeflas:
Thank you, guys, for the feedback!

PR #404, opened 2024-01-22
Improve visualize tests

greeflas:
Hello!

While working on #403 I've noted that unit tests for `Visualize()` worked not correctly in some cases.
If `CreateGraph()` method returns less `Ctor` elements than expected (and vice versa) - tests will pass.

So I fixed assert function and found new bug shown by `create graph with one constructor and as interface option` scenario.

### DIFF
--- a/visualize_test.go
+++ b/visualize_test.go
@@ -449,6 +449,8 @@ func assertCtorEqual(t *testing.T, expected *dot.Ctor, ctor *dot.Ctor) {
 }
 
 func assertCtorsEqual(t *testing.T, expected []*dot.Ctor, ctors []*dot.Ctor) {
+	assert.Len(t, ctors, len(expected))
+
 	for i, c := range ctors {
 		assertCtorEqual(t, expected[i], c)
 	}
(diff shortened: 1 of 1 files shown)


MuhammadKelanyy (approved):

visualize_test.go
```diff
@@ -449,6 +449,8 @@ func assertCtorEqual(t *testing.T, expected *dot.Ctor, ctor *dot.Ctor) {
 }
 
 func assertCtorsEqual(t *testing.T, expected []*dot.Ctor, ctors []*dot.Ctor) {
+	assert.Len(t, ctors, len(expected))
```
JacobOaks, line 452: Thanks for catching this!

In fact, one step better would probably be making this a `require.Len` so we stop the test immediately and not panic in the following loop if len(ctors) > len(expected)

JacobOaks (approved):

visualize_test.go
greeflas: Done.

JacobOaks:
Hey @greeflas, CI uncovered an error running this PR. This is actually due to some incorrect testing logic written by us. Since I don't have permission to update this PR, I have created a separate PR that supersedes this one (it includes your change and fixes the broken test): https://github.com/uber-go/dig/pull/405

greeflas:
Ok, welcome!
I'll close this PR.

PR #406, opened 2024-02-11
ci: Use golangci-lint

abhinav:
This PR switches to using golangci-lint to run linters.
staticcheck is included and enabled by default.
We add revive and a few others.
All issues reported by the linters have been fixed.

This also allows the linter to run in parallel with the tests.

License header check has also been moved to golangci-lint,
similarly to uber-go/fx#1157.

Refs uber-go/fx#1150
Refs uber-go/zap#1323


JacobOaks (approved):
Thanks Abhinav!

Issue #408, opened 2024-03-04
Intercept provided values


asmeikal:
**Describe the solution you'd like**
We need to intercept the provided values after they are constructed, to check for implemented interfaces and possibly call some initialization methods.

We have an implementation ready that adds a `WithProvidedCallback` option to the container that serves this purpose. The callback will receive all values after they are constructed, along with name/group information and type information.

**Describe alternatives you've considered**
This could be implemented as a `ProvideOption`, but would need a list of values as input of the callback, since constructors can return multiple values, and would be duplicated each time `Provide` is called on the container.

The callback could be implemented also as a `ScopeOption` to override behaviour in nested scopes, but since it would be the first actual option of this kind, we skipped implementation for now.

**Is this a breaking change?**
No, even if the callback is implemented as a `ScopeOption`.

**Additional context**
We are open sourcing an extremly opinionated dependency injection framework we use to develop microservices in Go. This (small) framework handles automatically health check registration, calling initialization and destruction methods at service startup and destruction, registration of prometheus metrics. Instead of relying on lifecycle hooks like Fx does, our framework relies on interfaces implemented by the provided values, e.g. by looking for the `Start` method on provided values and invoking it at application startup. This reduces code duplication, as most of this shared logic can be implemented with single methods on each component.

JacobOaks:
Hi @asmeikal, thanks for the issue & PR.

Dig already supports a callback system where you can register [`dig.Callback`](https://pkg.go.dev/go.uber.org/dig#Callback) using [`dig.WithProviderCallback`](https://pkg.go.dev/go.uber.org/dig#WithProviderCallback) and [`dig.WithDecoratorCallback`](https://pkg.go.dev/go.uber.org/dig#WithDecoratorCallback).

What are your thoughts on extending this system to support your needs? It seems like, specifically, you'd like:
* To be able to specify a default callback once for an entire Fx application
* Some additional information to get passed to callbacks.

I think both of these things can be accomplished in a backwards-compatible way without creating a second callback system. What do you think?

asmeikal:
Hi @JacobOaks, as mentioned in the initial message I wouldn't rely on the existing options, since they are registered at the provider level and would result in a strange interface.

If you are suggesting instead to rely on the `dig.Callback` for this new option, I'm not sure how to add the required information in the `dig.CallbackInfo` struct, since it contains information about the provider or decorator invocation, and the feature we are looking for is to get information on the constructed values instead.

I'm having these doubts, specifically:
[...]

PR #409, opened 2024-03-04
Intercept provided values

asmeikal:
This is our current implementation for #408. It adds a container Option to intercept values as they are constructed, to apply additional startup logic.

The naming could centrainly be improved!

### DIFF
--- a/constructor.go
+++ b/constructor.go
@@ -192,6 +192,27 @@ func (n *constructorNode) Call(c containerStore) (err error) {
 	// the rest of the graph to instantiate the dependencies of this
 	// container.
 	receiver.Commit(n.s)
+
+	if n.s.callback != nil {
+		for k, v := range receiver.values {
+			n.s.callback(ProvidedInfo{
+				Name:  k.name,
+				Type:  k.t,
+				Value: v,
+			})
+		}
+
+		for k, vs := range receiver.groups {
+			for _, v := range vs {
+				n.s.callback(ProvidedInfo{
+					Name:  k.group,
+					Type:  k.t,
+					Value: v,
+				})
+			}
+		}
+	}
+
 	n.called = true
 	return nil
 }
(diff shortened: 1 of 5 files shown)


JacobOaks:
Hey @asmeikal - thanks for working with us to get this to this state! One thing I think warrants discussion:

Giving access to the actual `reflect.Value`s returned by constructor/decorators in callback opens up opportunity for some funky stuff. For example, I think callback functions will be able to use reflect to modify the actual values stored inside containers (for addressable values at least), which could definitely cause unexpected/unsound behavior. One alternative I can think of would be to make sure that the values passed to callbacks are deep copies of those stored in the container, but I don't think this would be worth it.

So, my initial thoughts are that this is _probably_ fine, as long as we aggressively plaster clear documentation that modifying these values results in undefined behavior, but I'd like to hear your thoughts & thoughts of other maintainers @tchung1118 @sywhang.

tchung1118:
I agree with @JacobOaks that this is probably fine, but we should clearly document that those values in `CallbackInfo` shouldn't be modified.

callback.go
```diff
@@ -32,6 +34,10 @@ type CallbackInfo struct {
 	// function, if any. When used in conjunction with [RecoverFromPanics],
 	// this will be set to a [PanicError] when the function panics.
 	Error error
+
+	// Values contains all values constructed by the [Callback]'s
+	// associated function.
```
JacobOaks, line 39: Given the discussion, let's add some explicit documentation here that mentions the undefined results of modifying these values.

asmeikal: I updated the comment!

JacobOaks:
Hey there, the maintainers of Dig discussed this at some length and decided to open up a discussion to describe our concerns and solicit more thoughts: https://github.com/uber-go/dig/discussions/410

Issue #424, opened 2025-02-17
Allow linker to perform deadcode elimination for programs using fx


pgimalac:
**Describe the bug**
When `reflect`'s  `MethodByName` is used with a non-constant argument, the linker has to keep every public method of reachable types because it can't statically determine whether that method will be called through `reflect`. This makes binaries significantly bigger (in my experience, around 30%).
https://github.com/golang/go/blob/go1.23.6/src/cmd/link/internal/ld/deadcode.go#L420-L433
`text/template` uses `MethodByName` to call arbitrary methods when executing a template, so it disables that dead code elimination too.

`dig.Visualize` uses a template to generate the graph with dot format, and it is called by https://github.com/uber-go/fx when creating a new `App`, so basically using `fx` disables that dead code elimination.

**To Reproduce**
Consider this simple example:
```go
package main

import (
	_ "github.com/aws/aws-sdk-go-v2/service/ec2"
	"go.uber.org/fx"
)

func main() {
	_ = fx.New()
}
```

When building with upstream `dig`, the binary is 8.2MB.
When replacing `dig` with the version from [my branch](https://github.com/pgimalac/dig/tree/pgimalac/avoid-using-text-template), the binary is 5.6MB.

This is an extreme example (I import aws-sdk just to get a lot of types...) but the binary is 47% bigger when DCE is disabled.

**Fixing**
The template is static and not too complex, so replacing it with direct string manipulations is pretty easy, see https://github.com/uber-go/dig/pull/425 for example.

**Additional notes**
This is very similar to https://github.com/spf13/cobra/issues/2015 for `cobra`.

PR #425, opened 2025-02-17
Replace use of `text/template` with string manipulations

pgimalac:
Fixes https://github.com/uber-go/dig/issues/424.

This PR replaces using a template to generate the graph in dot format with directly writing strings in the writer.
In particular it avoids disabling dead code elimination.

I tried to keep the same whitespaces (tab, newlines) as in the template to avoid modifying the test cases, and increase confidence that the change is equivalent.
I'd be very happy to remove the extra tab and newlines as they just make the code less readable.

The proposed implementation uses `fmt.Fprintf` to simplify, but it can be optimized by just using `w.Write` directly instead, avoiding some parsing and allocations. I considered that it was not a hot path so making the code simpler was worth it.

### DIFF
--- a/visualize.go
+++ b/visualize.go
@@ -25,7 +25,6 @@ import (
 	"fmt"
 	"io"
 	"strconv"
-	"text/template"
 
 	"go.uber.org/dig/internal/dot"
 )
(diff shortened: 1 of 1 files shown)


JacobOaks (approved):
LGTM. For the purposes of this change, keeping the tests the same seems like a good idea to keep confidence. We can clean up the spacing later. Thanks!

PR #428, opened 2025-03-03
Prepare release v1.18.1

sywhang:
This preps release of Dig v1.18.1.

### DIFF
--- a/version.go
+++ b/version.go
@@ -21,4 +21,4 @@
 package dig
 
 // Version of the library.
-const Version = "1.19.0-dev"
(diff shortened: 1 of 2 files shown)


tchung1118 (approved):

version.go
```diff
@@ -21,4 +21,4 @@
 package dig
 
 // Version of the library.
-const Version = "1.19.0-dev"
+const Version = "1.18.1-dev"
```
hydra13, line 24: Is it correct to leave this suffix `-dev` here?

PR #440, opened 2025-10-20
Add option for naming decorator outputs

### LINKED ISSUE #439: Add `dig.Name()` decorator option
**Is your feature request related to a problem? Please describe.**
I have multiple named providers of the same type, and I'd like to decorate one of them
```go
c.Provide(NewReadOnlyConnection, dig.Name("ro"))
c.Provide(NewReadWriteConnection, dig.Name("rw"))
```
However the decorate function doesn't have an option to target these https://github.com/uber-go/dig/blob/c8eb7b9fb8478fe1f667298289bcea50a5de5faf/decorate.go#L159-L162

I use this mainly for tests- replacing certain dependencies with test versions.

Is there a non-obvious way to do this? I couldn't find an example in the unit tests.

**Is this a breaking change?**
No, we could create a new `DecorateOption`

Ran into this when looking into #230

andremarianiello:
I added a new decorate option `DecorateName`. The goal was to provide the same behavior as the `Name` provide option, but for decoration instead of provision. Fixes #439


andremarianiello:
@JacobOaks is there someone I should ask for a review from?

